PR #6, opened 2022-11-24
Alb002 alignment layout

Ghesselink:
First attempt for sandbox


aothms:
Sandbox ok 131f20951291e4aa5da7d8c30bad6e01a2ef7bc7

features/ALB002_Alignment-layout.feature
```diff
@@ -0,0 +1,35 @@
...
+      Then Each IfcAlignmentHorizontal must be nested only by 1 IfcAlignment
+       And  Each IfcAlignmentVertical must be nested only by 1 IfcAlignment
+       And  Each IfcAlignmentCant must be nested only by 1 IfcAlignment
+    
+  Scenario: Agreement on constraints of allowed attributes nesting nested attributes of IfcAlignment
+    # @note can probably be formulated better
+
+      Given a file with Schema Identifier "IFC4x3"
+
+      Then Each IfcAlignment must nest only the following entities: IfcAlignmentHorizontal, IfcAlignmentVertical, IfcAlignmentCant, IfcReferent, IfcAlignments
```
pjanck, line 27:
```suggestion
      Then Each IfcAlignment may nest only the following entities: IfcAlignmentHorizontal, IfcAlignmentVertical, IfcAlignmentCant, IfcReferent, IfcAlignment
```

- I think the meaning of `must` is too enforcing, i.e. that the entities listed `must` also be present. 
- Typo at the end, there is no `IfcAlignments` in the schema.

pjanck (changes requested):
I'm still trying to wrap my head around the whole architecture, therefore, first a small suggestion. I still need to go through the code and understand it.

features/ALB002_Alignment-layout.feature
aothms: I agree, this appears implemented as the entities types in the model should be a subset of the entities listed here https://github.com/buildingSMART/ifc-gherkin-rules/pull/6/files#diff-cf027576a6812e3403bd3d9579d83abfaa74a41607a4f3c986398ef8492a5d3dR248 In that case I find `may` also more suitable.

Ghesselink: Yes, the term 'only' seems the to be the most important/restrictive factor in this rule scenario; if this term was instead 'all', the implementation would be completely different (but changing the term 'must' to 'may' or 'should' does not change the implementation). Indeed, there should be only elements in that subset (in this case IfcAlignmentHorizontal .. AfcAlignment') in the nesting entity types (of IfcAlignment).

I've changed it to 'may' in the feature file, and thank you for noting the typo.
 I'm currently testing the scenarios with some new test files (this pull request was also partly to see the sandbox url functionality) and adding some changes. I'll also comment here when that's finished (and commit/pull request those changes)

features/steps/steps.py
```diff
@@ -200,6 +209,61 @@ def step_impl(context, constraint, num, entity):
...
+
+@then('Each {entity} must nest only the following entities: {other_entities}')
+def step_impl(context, entity, other_entities):
+
+    errors = []
+
+    if getattr(context, 'applicable', True):
+        for inst in context.model.by_type(entity):
+            nested_entities = [i for rel in inst.IsNestedBy for i in rel.RelatedObjects]
+            if not set([i.is_a() for i in nested_entities]).issubset(other_entities.split(',')):
```
aothms, line 248: As a team we need to decide on using

- `instance.is_a(str)`, or
- `instance.is_a() == str`, or
- `instance.is_a().lower() == str.lower()`

The first takes into account (a) inheritance and (b) difference in case and (c) will throw an error when `str` is not a valid entity name given the file schema of `instance`. The second takes into account nothing. The third takes into account differences in case.

aothms: A comment on `issubset()` (that I had to look up) it actually also returns true when the sets are equal. For that reason I prefer writing `<=` (subset or equal) because it's a better differentiated from `<` (strict subset).

features/steps/steps.py
```diff
@@ -200,6 +209,61 @@ def step_impl(context, constraint, num, entity):
...
+
+@then('Each {entity} nests a list of {segment_annotation}, each of which has DesignParameters typed as {parameter_annotation}')
+def step_impl(context, entity, segment_annotation, parameter_annotation):
+
+    errors = []
+    
+    if getattr(context, 'applicable', True):
+        for inst in context.model.by_type(entity):
+            segments = [inst for rel in inst.IsNestedBy for inst in rel.RelatedObjects]
+            if not all(operator.eq(segment.is_a(),segment_annotation) for segment in segments):
```
aothms, line 260: Let's await the discussion on the gherkin part of this. But `operator.eq(segment.is_a(),segment_annotation)` is a bit very lisp-like. `segment.is_a() == segment_annotation` or `segment.is_a(segment_annotation)` (see other discussion).

features/steps/steps.py
```diff
@@ -200,6 +209,61 @@ def step_impl(context, constraint, num, entity):
...
+                errors.append(instance_nesting_error(inst, segments))
+            parameters = [segment.DesignParameters for segment in segments]
+            if not all(operator.eq(parameters.is_a(),parameter_annotation) for parameters in parameters):
+                errors.append(instance_nesting_error(inst, parameters))
+            #@todo perhaps a different data class with more precise information, how many? which other attributes were found?
```
aothms, line 265: Indeed, that's exactly the reason why I find my approach to the gherkin clearer. The error category would be much more natural.

aothms (changes requested):

aothms:
Awesome work. It's a long list of comments, but as @evandroAlfieri mentioned that's simply just because we're at the start of all of this and as a team we have to set a consistent approach.

features/steps/steps.py
```diff
@@ -200,6 +209,61 @@ def step_impl(context, constraint, num, entity):
...
+
+@then('Each {entity} must be nested only by {other_entity}')
+def step_impl(context, entity, other_entity):
+    
+    errors = []
+
+    if getattr(context, 'applicable', True):
+        for inst in context.model.by_type(entity):
+            relating_objects_lst = [i.RelatingObject.is_a() for i in inst.Nests]
+            relating_object = relating_objects_lst[0]
```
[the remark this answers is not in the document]
Ghesselink: True, when I wrote it (before the meeting on Friday) I did not took this into consideration

features/ALB002_Alignment-layout.feature
```diff
@@ -0,0 +1,35 @@
...
+
+  Scenario: Agreement on structure of alignment segments
+
+      Given a file with Schema Identifier "IFC4x3"
+
+       Then Each IfcAlignmentHorizontal nests a list of IfcAlignmentSegment, each of which has DesignParameters typed as IfcAlignmentHorizontalSegment
```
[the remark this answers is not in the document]
Ghesselink: I was working on some 'fail files' and also encountered this. For now I just created a file when the nesting list contains a complete different element. 

There's also a possibility (I think) to pass more identifiers, as briefly discussed last Friday. For example: Given a file with (one of the) Schema Identifier (s) "IFC4X3", "IFC4x3_ADD1". Then you can be very precise and add versions later. I'm still trying to work out how this works exactly (An alternative is Jakub's solution, but both solutions give errors at the moment for me; https://github.com/buildingSMART/ifc-gherkin-rules/commit/45bcfa051139339d1d6fb315526d90ee591f57d0). 
A downside I can think of is that you'll have to manually add each new schema to every scenario (and possibly in every feature files). That will create lot of opportunities for typo's and missed scenario/feature files .. Maybe there is a possibility of creating a list with an overview of schema-rule relationships?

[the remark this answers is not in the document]
Ghesselink: A last comment (but maybe mostly to myself, hehe): we'll have to be very precise with comma's and whitespaces as well when we give multiple elements in a single statement. I've tried to split 'Then Each IfcAlignment may nest only the following entities: IfcAlignmentHorizontal, IfcAlignmentVertical (etc..)' and kept running into errors. Turned out, after a minor headache, I forgot the whitespace after the comma of each element and was splitting as str.split(',') instead of str.split(', ').  Can you maybe add this to your 'styling document' @Evandro ?

features/steps/steps.py
```diff
@@ -78,6 +78,16 @@ class instance_structure_error:
     def __str__(self):
         return f"The instance {fmt(self.related)} is assigned to {fmt(self.relating)}"
 
+@dataclass
+class instance_nesting_error:
```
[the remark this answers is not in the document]
Ghesselink: Thanks, I already found myself returning to the data classes that you wrote but was not sure how to structure them precisely/move data over. These comments help.

features/steps/steps.py
Ghesselink: You're right, it should return True if the two sets are equal. A better name (logically) would then be .isorequaltosubset but I can see why they did not pick such a name and went for .is instead .. <= is short and clear

instance.is_a(str) seems to be the most comprehensive you how describe it. Maybe I am understanding it wrong, but in case of (c) it just returns False (e.g.  ifc_file.by_type("IfcAlignment")[0].is_a('IfcNotExistingBlob') returns False) but not an error?

features/ALB002_Alignment-layout.feature
Ghesselink: The changes are added in https://github.com/buildingSMART/ifc-gherkin-rules/commit/a33767f28032eb37600de5d4d0bb6834773bd60e

aothms:
Sandbox ok a33767f28032eb37600de5d4d0bb6834773bd60e

aothms:
Sandbox ok e31cf2000e3b4f1fed553d4728b9217b142c6b7a

pjanck (changes requested):
Staying within the Gherkin feature review, I wonder if this is really the correct place?

features/ALB002_Alignment-layout.feature
```diff
@@ -0,0 +1,37 @@
...
+       And  Each IfcAlignmentVertical must be nested only by 1 IfcAlignment
+       And  Each IfcAlignmentCant must be nested only by 1 IfcAlignment
+  
+    
+  Scenario: Agreement on constraints of allowed attributes nesting nested attributes of IfcAlignment
+    # @note can probably be formulated better
+
+      Given a file with Schema Identifier "IFC4X3"
+
+      Then Each IfcAlignment may nest only the following entities: IfcAlignmentHorizontal, IfcAlignmentVertical, IfcAlignmentCant, IfcReferent, IfcAlignment
```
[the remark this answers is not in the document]
evandroAlfieri: Why not? If the grammar is successfully implemented, we could move this sentence in a dedicated rule (.feature file), with an explicit name (e.g., ALB003_Allowed entities nested in Alignment)

features/steps/steps.py
```diff
@@ -200,6 +234,65 @@ def step_impl(context, constraint, num, entity):
 
     handle_errors(context, errors)
 
+@then('Each {entity} must nest {constraint} {num:d} instance(s) of {other_entity}')
```
aothms, line 237: I know this is confusing and I also don't necessarily agree on the wording, but at the moment, we have the terminology in gherkin and IFC reversed.

Gherkin|IFC
---|---
Each {entity} must nest|inst.IsNestedBy 
Each {entity} must be nested only by |inst.Nests

features/steps/steps.py
```diff
@@ -200,6 +234,65 @@ def step_impl(context, constraint, num, entity):
...
+    handle_errors(context, errors)
+
+@then('Each {entity} must be nested only by {num:d} {other_entity}')
+def step_impl(context, entity, other_entity, num):
+    errors = []
+
+    if getattr(context, 'applicable', True):
+        for inst in context.model.by_type(entity):
+            relating = [i.RelatingObject for i in inst.Nests]
+            if not all([len(relating) <= num, other_entity == relating[0].is_a()]):
```
aothms, line 262: This is a bit peculiar.

`inst.Nests` is length 0 or 1 in the schema already. So I would propose:

- the Gherkin rule should be rewritten without the `num:d` (it becomes a duplicate then of `.. nests a list of only ..`?
- if we have a list of length 0..1 I'd wouldn't use the `all()` predicate, especially since you already only do `relating[0]`.

features/steps/steps.py
```diff
@@ -217,6 +310,17 @@ def step_impl(context, related, relating, other_entity, condition):
...
 
     handle_errors(context, errors)
+
+@then ('The value of attribute {attribute} should be of type {expected_entity_type}')
+def stemp_impl(context, attribute, expected_entity_type):
+
+    def _():
+        for inst in context.instances:
+            related_entity = getattr(inst, attribute, [])
+            if not related_entity.is_a(expected_entity_type):
```
aothms, line 323: As mentioned earlier, we need to get `is_a(s)` and `is_a() == s` consistent,

aothms (changes requested):

features/steps/steps.py
Ghesselink: It's changed in https://github.com/buildingSMART/ifc-gherkin-rules/commit/df79f1f62d7459a92e7a59cbaa0ab27eb89a5508
The wording is perhaps a bit confusing now, but Gherkin and IFC are consistent in this commit.

features/steps/steps.py
Ghesselink: Implemented in https://github.com/buildingSMART/ifc-gherkin-rules/commit/0c8fdcad7ba2cb8627485193daa2643b93e770a3
The rule is rewritten without the num:d and is indeed a duplicate. The implementation is also more general, so it will check for both more than one relationship as well as more relating objects within a relationship. If the gherkin sentence is specific however, as 'nests a list ..', the code will check for just a list.

features/steps/steps.py
Ghesselink: All set to is_a(s) in https://github.com/buildingSMART/ifc-gherkin-rules/commit/0c8fdcad7ba2cb8627485193daa2643b93e770a3

features/steps/steps.py
Ghesselink: In https://github.com/buildingSMART/ifc-gherkin-rules/commit/0c8fdcad7ba2cb8627485193daa2643b93e770a3

test/files/alb002/README.md
```diff
@@ -0,0 +1,4 @@
+| File name     | Expected result | Error log / further info                                   |
+|---------------|-----------------|------------------------------------------------------------|
+| Pass-Alb002   | success         | n.a.                                                       |
```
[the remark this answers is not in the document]
Ghesselink: In https://github.com/buildingSMART/ifc-gherkin-rules/commit/0c8fdcad7ba2cb8627485193daa2643b93e770a3

features/steps/steps.py
```diff
@@ -200,6 +209,61 @@ def step_impl(context, constraint, num, entity):
...
+    stmt_to_op = {'exactly': operator.eq, "at most": operator.le}
+    assert constraint in stmt_to_op
+    op = stmt_to_op[constraint]
+
+    errors = []
+
+    if getattr(context, 'applicable', True):
+        for inst in context.model.by_type(entity):
+            nested_entities = [entity for rel in inst.IsNestedBy for entity in rel.RelatedObjects]
+            if not op(sum(1 for i in nested_entities if i.is_a() == other_entity), num):
```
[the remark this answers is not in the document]
Ghesselink: In https://github.com/buildingSMART/ifc-gherkin-rules/commit/0c8fdcad7ba2cb8627485193daa2643b93e770a3

features/steps/steps.py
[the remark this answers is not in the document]
Ghesselink: Dataclasses are merged in https://github.com/buildingSMART/ifc-gherkin-rules/commit/0c8fdcad7ba2cb8627485193daa2643b93e770a3

features/ALB002_Alignment-layout.feature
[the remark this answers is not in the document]
Ghesselink: Discussion moved to https://github.com/buildingSMART/ifc-gherkin-rules/issues/18

features/ALB002_Alignment-layout.feature
```diff
@@ -0,0 +1,50 @@
...
+       And Each IfcAlignmentVertical must nest: only 1 instance(s) of IfcAlignment
+       And Each IfcAlignmentCant must nest: only 1 instance(s) of IfcAlignment
+  
+  Scenario: Agreement of structure of alignments segments
+
+      Given a file with Schema Identifier "IFC4X3"
+
+      Then Each IfcAlignmentHorizontal is nested by: a list of only instance(s) of IfcAlignmentSegment
+      Then Each IfcAlignmentVertical is nested by: a list of only instance(s) of IfcAlignmentSegment
+      Then Each IfcAlignmentCant is nested by: a list of only instance(s) of IfcAlignmentSegment
```
pjanck, line 28: Syntax question: when to use `:` in sentences, when not to?

features/ALB002_Alignment-layout.feature
```diff
@@ -0,0 +1,50 @@
...
+
+
+  Scenario: Agreement of the segments of the horizontal alignment
+      
+      Given an IfcAlignmentSegment
+        And The element nests an IfcAlignmentHorizontal
+      
+       Then The value of attribute DesignParameters should be of type IfcAlignmentHorizontalSegment
```
pjanck, line 36: I find the syntax confusing:

> The _value_ of attribute ... should be of _type_ ...

This is mixing two concepts `value` and `type` within the same sentence.

Proposal:

> The _type_ of attribute ... should be ...

pjanck (changes requested):
Remaining within Gherkin still ...

features/ALB002_Alignment-layout.feature
Ghesselink: Ideally never, it is a temporary solution for a problem regarding placeholders in python. For example, the implementation for 'must nest only 1 instance' might in the future be re-used by 'must not be nested by at most 2', so {must: must not}, {nest: nested by}, {only: by only}.

Ghesselink: Ideally never, it is a temporary solution for a problem regarding placeholders for strings in python. There are a couple of solutions, but we still need to discuss a couple of theoretical issues regarding alignment before putting a lot of effort into this.

features/ALB002_Alignment-layout.feature
Ghesselink: That is simple, makes a lot of sense, and is more suitable for reuse as well. I'll implement it, thanks.

features/ALB002_Alignment-layout.feature
Ghesselink: Late reply, I solved it but the checks did not pass. That turned out to be due to an error with imports in the github environment.
The code now uses custom parsing of the gherkin sentence https://github.com/buildingSMART/ifc-gherkin-rules/pull/6/commits/4baf964890d619e7aaa8dcee74394152d1d7ac0c . That way, it works. 

The problem was in the sentence:
`Each Alignment is {relationship_type} {condition} of AlignmentVertical`
if you type 
`Each IfcAlignment is nested by a list of only instance(s) of IfcAlignmentVertical`
the behave module by default does not know we are searching for **'is nested by'** and **'a list of only'** and will return **'is**' and **'nested by a list of only',** which is not very helpful. 

With custom parsing (more precisely, with the pyparsing module) we can be very precise which terms to look for, which is good I think.

aothms (approved):

aothms:
It was discussed that in the future we might revisit this when we wrap parameters in brackets so that a dedicated grammar in this case might be eliminated (but remains useful in the future).

PR #13, opened 2022-11-28
WiP: alb001 alignment in spatial structure

JTurbakiewicz:
WiP -> to discuss


features/steps/steps.py
```diff
@@ -79,6 +79,19 @@ class instance_structure_error:
     def __str__(self):
         return f"The instance {fmt(self.related)} is assigned to {fmt(self.relating)}"
 
+@dataclass
+class instance_contained_error:
```
aothms, line 83: We should probably merge this with @Ghesselink's `instance_structure_error` https://github.com/buildingSMART/ifc-gherkin-rules/pull/6/files#diff-cf027576a6812e3403bd3d9579d83abfaa74a41607a4f3c986398ef8492a5d3dR74

features/steps/steps.py
```diff
@@ -221,3 +234,46 @@ def step_impl(context, related, relating, other_entity, condition):
...
 
     handle_errors(context, errors)
+
+
+@then('Each {entity} {condition} be {directness} contained in {other_entity}')
+def step_impl(context, entity, condition, directness, other_entity):
+    stmt_to_op = ['must', 'must not']
+    assert condition in stmt_to_op
+
+    stmt_about_directness = ['directly', 'indirectly', 'directly or indirectly', 'indirectly or directly']
```
aothms, line 244: I am not sure if there ever will be a use cases for `indirectly` only, but we can leave it in.

features/steps/steps.py
```diff
@@ -221,3 +234,46 @@ def step_impl(context, related, relating, other_entity, condition):
                         errors.append(instance_structure_error(inst, rel.RelatingObject))
 
     handle_errors(context, errors)
+
+
+@then('Each {entity} {condition} be {directness} contained in {other_entity}')
```
aothms, line 239: This is not incorporating the `... or subtypes` postfix we discussed in a call. It's also not immediately relevant of course, but since the gherkin is a bit speculative anyway, maybe the purpose of this PR can be to investigate how general we can get.

features/steps/steps.py
```diff
@@ -221,3 +234,46 @@ def step_impl(context, related, relating, other_entity, condition):
...
+    assert directness in stmt_about_directness
+
+    errors = []
+
+    if context.instances and getattr(context, 'applicable', True):
+        for ent in context.model.by_type(entity):
+            if len(ent.ContainedInStructure) > 0:
+                containing_relation = ent.ContainedInStructure[0]
+                relating_spatial_element = containing_relation.RelatingStructure
+                is_directly_contained = getattr(relating_spatial_element, 'is_a')(other_entity)
```
aothms, line 254: `getattr(relating_spatial_element, 'is_a')`

is equivalent to

`relating_spatial_element.is_a`

I'd say only use `getattr(object, name[, default])` if:
- name is a variable
- default is used so that you don't get a KeyError on non-existant attr

aothms, line 254: We need to decide on `is_a(str)` and `is_a() == str` wrt to subtypes.

features/steps/steps.py
```diff
@@ -221,3 +234,46 @@ def step_impl(context, related, relating, other_entity, condition):
...
+                while len(relating_spatial_element.Decomposes) > 0:
+                    decomposed_element = relating_spatial_element.Decomposes[0]
+                    relating_spatial_element = decomposed_element.RelatingObject
+                    is_indirectly_contained = getattr(relating_spatial_element, 'is_a')(other_entity)
+                    if is_indirectly_contained:
+                        break
+            elif len(ent.ContainedInStructure) == 0:
```
aothms, line 261: `elif ...` is equivalent to `else` here (as len never goes negative). I would prefer `else`.

features/steps/steps.py
```diff
@@ -221,3 +234,46 @@ def step_impl(context, related, relating, other_entity, condition):
...
+                is_directly_contained = False
+                is_indirectly_contained = False
+            if condition == 'must':
```
aothms, line 264: I think these 6 cases can be generalized when thinking about them as set intersection.

You have:
- required_directness 
- observed_directness

if both of these are {'directly'}, {'indirectly'} or {'directly', 'indirectly'}. Then you can simply use the set intersection on both.

The set intersection should be empty in case of `must_not`, non empty in case of `must`.

`if bool(required_directness & observed_directness) == (condition == must):` <- untested but sth like this. It wouldn't win the price for readability so a comment is nice, or breaking this up into some variables can also help for readability. But it would greatly reduce the risk of errors when all cases are reduced to a single case.

aothms:
I really like how far you went here with creating a general solution. I have some additional remarks to that end. We'll wait for @evandroAlfieri's verdict on whether we merge this:

- without the gherkin feature
- with the gherkin feature commented out
- with the gherkin feature with an additional tag so that it is ignored

aothms (changes requested):

features/steps/steps.py
Ghesselink: I'd propose to use is_a(str), is that okay for you? It seems more readable, more build-in checks and little (obvious at least) downsides, other than some restructuring here and there. I also didn't run into any problems while testing.

features/ALB001_Alignment-in-spatial-structure.feature
```diff
@@ -0,0 +1,9 @@
+@implementer-agreement
+@ALB
+Feature: ALB001_Alignment in spatial structure
+
+  Scenario: Agreement on each IfcAlignment being contained in an IfcSite
+
+      Given A file with Schema Identifier "IFC4X3"
+      And An IfcAlignment
+      Then Each IfcAlignment must be directly contained in IfcSite
```
pjanck, line 9: What does it look like if something is `indirectly` contained?

pjanck (commented):
I have a difficulty understanding a rule.

features/steps/steps.py
JTurbakiewicz: Very cool idea, added!

features/steps/steps.py
JTurbakiewicz: I am not 100% sure what you mean here. 

Is it the following? 

`@then('Each {entity} {condition} be {directness} contained in {other_entity} or its subtypes')`
Then in our particular case, it would be:
`@then'Each IfcAlignment must be directly contained in IfcSite or its subtypes'
So the check will pass if the IfcAlignment is contained either in the IfcSite or in any subtype (via inheritance) of IfcSite?

Good idea, I'll try to add it.

features/ALB001_Alignment-in-spatial-structure.feature
JTurbakiewicz: The intention was to have `Indirectly` take into account spatial relationships. Let's provide some examples.

Directly:
IfcAlignment > IfcSite 
IfcAlignment connected to IfcSite via the relationship IFCRELCONTAINEDINSPATIALSTRUCTURE

Indirectly: 
IfcAlignment > SpacialElement1 > SpacialElement2 > IfcSite
IfcAlignment connected to SpacialElement1 via the relationship IFCRELCONTAINEDINSPATIALSTRUCTURE
SpacialElement1 to SpacialElement2  via the relationship IFCRELAGGREGATES
SpacialElement2  to IfcSite via the relationship IFCRELAGGREGATES

Maybe it is worthwhile to try different wording. Perhaps the following?
`Each IfcAlignment must be directly or through spatial relationships contained in IfcSite`

@pjanck @aothms @evandroAlfieri

aothms: Maybe it helps to think of this as a regular expression or sparql property path. E.g what's the pattern of relationships that can be followed in order to be valid.

```
elem -> ContainedIn -> Site
elem -> Decomposing -> Assembly -> ContainedIn -> Site
elem -> Decomposing -> Assembly -> ContainedIn -> Something -> Decomposing -> Site
```

We could write that as something like:

    elem -> Decomposing * -> ContainedIn -> Decomposing * -> Site
    
Where * means zero or more repetitions of the preceeding relationship.

I don't think multiple ContainedIn relationships would be allowed actually in the IFC spec, so "directly or through spatial relationships" maybe better be "directly or through decomposition relationships"

The question is also do we want to make this more explicit in Gherkin (`following a pattern of one or more RelX exactly one RelY zero ore more RelZ`) or do we want to keep it something descriptive (e.g `indirectly`) matched in code? The ever remaining tension between explicit IFC constructs in rules (nice for people familiar with them, less so for the people that predominantly care about business logic).

If we want to make this more explicit in Gherkin we can rely on Behave to do the course matching for us but in the implementation use a custom grammar to more precisely match against the semantics of the predicate/sentence fragment. Something like this:

```
>>> from pyparsing import *
>>> zom = CaselessKeyword("zero or more")
>>> eo = CaselessKeyword("exactly one")
>>> entity = Word(alphas)('entity')
>>> grammar = OneOrMore((zom | eo) + entity)
>>> grammar.parseString("zero or more IfcRelX exactly one IfcRelY zero or more IfcRelZ")
ParseResults(['zero or more', 'IfcRelX', 'exactly one', 'IfcRelY', 'zero or more', 'IfcRelZ'], {'entity': 'IfcRelZ'})
```

features/steps/steps.py
JTurbakiewicz: I didn't specifically add `... or subtypes`, as it's already checking the inheritance via is_a method.

JTurbakiewicz:
Changed the feature file to add the comment: @disabled, but the rule is still performed. Will look into the code if it can be fixed.

aothms:
@JTurbakiewicz we have merge conflicts now, would you be so kind... Better if you have a quick look to prevent unwanted side effects.

JTurbakiewicz:
Resolved merge conflicts. I'm afraid they will reappear when other branches are merged, probably we must do it one by one. 

But I like resolving the conflicts, so don't mind :)

Ghesselink:
> Resolved merge conflicts. I'm afraid they will reappear when other branches are merged, probably we must do it one by one.
> 
> But I like resolving the conflicts, so don't mind :)

Yes, I got the same thing; had to merge the open PR when another PR was merged into main. 
If I understood correctly, tomorrow we're going to merge all the open PR? Maybe an idea for us to be also present. Not at the meeting of Thomas and Evandro per se, but to solve potential merge conflicts when they occur hopefelly we can merge more in one go. 
I'll be online in the (EU) afternoon myself.

aothms (approved):

aothms:
You were very right, sorry I stole your merging fun.

PR #15, opened 2022-11-29
allow multiple schemas for given statements

Ghesselink:
As discussed, it might be useful to be able to allow for multiple schemas. In case a rule is applicable to more schemas or when new schemas are added in the future. It will still work when only one schema is allowed. 

As an example, it can be implemented in a gherkin feature given statement like this:
Given A file with Schema Identifier "IFC4X3", "IFC4X3_ADD1"

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -25,8 +25,8 @@ def get_mvd(ifc_file):
         detected_mvd = ifc_file.header.file_description.description[0].split(" ", 1)[1]
         detected_mvd = detected_mvd[1:-1]
     except:
-        detected_mvd = None
-    return detected_mvd
+        detected_mvd = ''
+    return detected_mvd.lower()
 
 def get_inst_attributes(dc):
     if hasattr(dc, 'inst'):
@@ -167,14 +167,15 @@ def step_impl(context, attribute, value):
 
 @given('A file with {field} "{values}"')
 def step_impl(context, field, values):
-    values = list(values.replace('"', "").split(', '))
+    values = list(map(str.lower, map(lambda s: s.strip('"'), values.split(' or '))))
     if field == "Model View Definition":
-        applicable = any([get_mvd(context.model) == value for value in values])
+        applicable = get_mvd(context.model) in values
     elif field == "Schema Identifier":
-        applicable = any([context.model.schema.lower() == value.lower() for value in values])
+        applicable = context.model.schema.lower() in values
     else:
         raise NotImplementedError(f'A file with "{field}" is not implemented')
 
+    context.applicable = getattr(context, 'applicable', True) and applicable
 
 @then('There shall be {constraint} {num:d} instance(s) of {entity}')
 def step_impl(context, constraint, num, entity):
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -25,8 +25,8 @@ def get_mvd(ifc_file):
         detected_mvd = ifc_file.header.file_description.description[0].split(" ", 1)[1]
         detected_mvd = detected_mvd[1:-1]
     except:
-        detected_mvd = ''
-    return detected_mvd.lower()
+        detected_mvd = None
+    return detected_mvd
 
 def get_inst_attributes(dc):
     if hasattr(dc, 'inst'):
@@ -164,12 +164,12 @@ def step_impl(context, attribute, value):
         filter(lambda inst: getattr(inst, attribute) == value, context.instances)
     )
 
-
 @given('A file with {field} "{values}"')
 def step_impl(context, field, values):
     values = list(map(str.lower, map(lambda s: s.strip('"'), values.split(' or '))))
     if field == "Model View Definition":
-        applicable = get_mvd(context.model) in values
+        conditional_lowercase = lambda s: s.lower() if s else None
+        applicable = conditional_lowercase(get_mvd(context.model)) in values
     elif field == "Schema Identifier":
         applicable = context.model.schema.lower() in values
     else:
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -165,16 +165,15 @@ def step_impl(context, attribute, value):
     )
 
 
-@given('A file with {field} "{value}"')
-def step_impl(context, field, value):
+@given('A file with {field} "{values}"')
+def step_impl(context, field, values):
+    values = list(values.replace('"', "").split(', '))
     if field == "Model View Definition":
-        applicable = get_mvd(context.model) == value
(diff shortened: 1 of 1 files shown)


features/steps/steps.py
```diff
@@ -165,17 +165,16 @@ def step_impl(context, attribute, value):
     )
 
 
-@given('A file with {field} "{value}"')
-def step_impl(context, field, value):
+@given('A file with {field} "{values}"')
+def step_impl(context, field, values):
+    values = list(values.replace('"', "").split(', '))
```
aothms, line 170: In @evandroAlfieri's guide we have "or". So instead of `split` on "," why don't we split on " or "? Instead of just blindly replacing `"` with ``. I'd say.

values = map(str.lower, map(lambda s: s.strip('"'), values.split(' or ')))

features/steps/steps.py
```diff
@@ -165,17 +165,16 @@ def step_impl(context, attribute, value):
...
     if field == "Model View Definition":
-        applicable = get_mvd(context.model) == value
+        applicable = any([get_mvd(context.model) == value for value in values])
     elif field == "Schema Identifier":
-        applicable = context.model.schema.lower() == value.lower()
+        applicable = any([context.model.schema.lower() == value.lower() for value in values])
```
aothms, line 174: With values already containing all lowercase (due to comment above) you could then simply do

features/steps/steps.py
```diff
@@ -165,17 +165,16 @@ def step_impl(context, attribute, value):
...
     else:
         raise NotImplementedError(f'A file with "{field}" is not implemented')
 
-    context.applicable = getattr(context, 'applicable', True) and applicable
```
aothms, line 177: Why was this removed. This is necessary for "chaining" / "combining" multiple givens from this pattern.

aothms (changes requested):

features/steps/steps.py
Ghesselink: That is better, yes. The rationale was that a good sentence would be "Given a file with Schema Identifier "IFC4X3", "IFC4X3_TC1" or "IFC4X3_ADD1". But maybe it is better to split on 'or', indeed. The sentence is would then be not completely perfect English ('a' or 'b' or 'c') but neither would the code I wrote  ('a','b','c'). The first one makes logical sense and is the closest to Evandro's guide. 

I only had to add the map object to a list to work with the rest of the function:
```suggestion
    values = list(map(str.lower, map(lambda s: s.strip('"'), values.split(' or '))))
```

features/steps/steps.py
Ghesselink: The works for Schema Identifier, but not Model View Definition. The line you wrote converts a string to lower cases, but then the value for model view definition will not match 'get_mvd(context.model)' in some IFC2X3 schemas in other rules. An option would be to convert this to lower cases as well, but this throws an error since the get_mvd function has an error handler to return None (TypeError). I've changed the except condition in get_mvd() function to return an empty string instead, and for the function to return lower cases. 

The changes are in, to make it clearer https://github.com/buildingSMART/ifc-gherkin-rules/pull/15/commits/4b6c110e75f4be3bda8d04836ac1ae696336cd41 
Maybe you can take a quick look at it @aothms ? Another option would be to have a workaround in the function linked to the given statement (making sure get_mvd does not return None), which one has your preference?
Pytest runs well in both options

features/steps/steps.py
Ghesselink: That was unintentional. I've commented some things out (including the mentioned line), worked on it on another branch to test it with the rules I wrote and copy-pasted it back. I see now how that can lead to errors, it would then be a better/more careful practice to actually develop on the branch that will be pushed ..

features/steps/steps.py
aothms: @Ghesselink https://peps.python.org/pep-0505/ would have made that easier "none-aware operators", but it looks like it's not going to make it.

I prefer to have get_mvd() unaltered to keep this kind of detail as local as possible, we might later need to print it to the user for example.

Most concise would be:

`(get_mvd() or '').lower() in ...`

It would also not propagate the None. So maybe best is something like:

```
conditional_lowercase = lambda s: s.lower() if s else None
conditional_lowercase(get_mvd()) in ...
```

features/steps/steps.py
aothms: If we want to support "a, b or c". Then we have to use regexes.

```
>>> import re
>>> re.split(r' *(\bor\b|,) *', 'a, b, or c, d, or e')
['a', ',', 'b', ',', '', 'or', 'c', ',', 'd', ',', '', 'or', 'e']
>>> ps = _
>>> ps[1::2]
[',', ',', 'or', ',', ',', 'or']
>>> re.match(r'\,*or', "".join(ps[1::2]))
>>> ps = re.split(r' *(\bor\b|,) *', 'a, b, c, d, or e')
>>> re.match(r'\,*or', "".join(ps[1::2]))
<re.Match object; span=(0, 6), match=',,,,or'>
```

* zero or more spaces
* followed by
  - either `or` (at word boundaries)
  - or `,`
* followed by zero or more spaces

The second regex is to test we use zero or more `,` followed by one `or` to join the parts.

I'm not against it, it's fun, but not sure if this is worth the effort.

features/steps/steps.py
Ghesselink: Doesn't look like it, created in 2015 and still deferred .. In the discussion programming gets almost a spiritual/philosophical component, 'is there something when there's nothing'? ;p Maybe not a coincidence that quite a bit of philosophy students turned to programming. 

Anyway, what you suggested is tested and works fine in https://github.com/buildingSMART/ifc-gherkin-rules/pull/15/commits/b9c8689101207aa3f68d66c8b46e11274e9f0b66

features/steps/steps.py
Ghesselink: That's also what I saw when I looked it up. I've used regexes before, but it tends to cost a lot of time (for me at least) to get it exactly right. 

Would it be okay for you to merge as in https://github.com/buildingSMART/ifc-gherkin-rules/pull/15/commits/b9c8689101207aa3f68d66c8b46e11274e9f0b66 ? 
I have time tomorrow, in the weekend here and there, and think it would be better spent on things more directly related to gherkin (fourth scenario needs to be adjusted, some documentation, test files, etc.) before Monday. I've discussed it briefly with Evandro and he shares this point of view.

aothms: Ok from me :)

aothms (approved):

aothms:
Great work :)

PR #19, opened 2022-12-11
move files folder, delete two pass ifc files

Ghesselink:
If I'm correct, the gem002 test files have to be inside '/files' in order for pytest to work (we did the same with alb001 and alb002).
The pass file ('no space') was the same as the same fail 'no-representation' and was also failing in pytest due to the file not having a building (SPS001), thereby triggering an AssertionError 
`elif base.startswith("pass-"):
        assert len(results) == 0`

Like this it should work. For the rest, nice code, it helps me as well ..  Did not know you can do a filter(None ... :)


johltn:
> The pass file ('no space') was the same as the same fail 'no-representation' and was also failing in pytest due to the file not having a building
The unit test files status depends on the rule it has been created for. The goal is to test it for a particular rule. So in this case it indeed fails for sps001 but doesn't for gem002, as expected.
In addition, you need to be careful, `pass-gem002-no-space.ifc` and `fail-gem002-no-representation` do not represent the same case.

Ghesselink:
Ah, that makes it more clear. There was some discussion if new files that are created needed to pass other rules as well or the goal was to test for a particular rule, as you mention. It makes more sense to test only for a particular rule, its quite some additional effort to make sure every new file passes the other rules. 
A concern is that ‘pytest -sv’ gives back AssertionErrors errors on other rules. Am I right it then also won’t pass the checks on a new commit? 

Von meinem iPhone gesendet

PR #21, opened 2022-12-13
GEM003 Unique Representation Identifier

Ghesselink:


test/files/gem003/generate.py
```diff
@@ -0,0 +1,46 @@
...
+            RelatingObject=site,
+            RelatedObjects = [building]
+        )
+
+    create_ifcpolyline = lambda point_list : file.createIfcPolyLine([file.createIfcCartesianPoint(point) for point in point_list])
+    polyline = create_ifcpolyline([(0.0, 0.0, 0.0), (5.0, 0.0, 0.0)])
+
+    identifiers = ['Body' for i in range(num_body_identifier)] + ['Axis' for i in range(num_axis_identifier)]
+
+    representations = [file.createIfcShapeRepresentation(context, i, "Curve2D", [polyline]
```
aothms, line 37: This creates a `Body` representation with `Curve2D` and IfcPolyline. I really prefer we have an IfcExtrudedAreaSolid here, so that the file makes a bit more sense.

test/files/gem003/generate.py
```diff
@@ -0,0 +1,46 @@
...
+                      ) for i in identifiers]
+
+    shape = file.createIfcProductDefinitionShape(None, None, representations)
+    from collections import Counter
+
+    duplicates = [value for (value,count) in Counter(identifiers).items() if count > 1]
```
aothms, line 43: This works, but another way is to check `duplicates = len(identifiers)  != len(set(identifiers))` because a `set` will never have duplicates.

features/GEM003_Unique-representation-identifier.feature
```diff
@@ -0,0 +1,11 @@
+@informal-proposition
+@GEM
+Feature: GEM003 - Unique Representation Identifer
+
+  Scenario: No Shape Representation identifier shall be used twice within the product representation of an element 
+
+      Given An IfcProductDefinitionShape
+        And The element has more than 1 instance(s) of IfcShapeRepresentation
+
+        Then Each instance of IfcShapeRepresentation has a unique value for the attribute RepresentationIdentifier
```
aothms, line 10: I would question here whether it's not better to use the attribute name in gherkin as opposed to the type of attribute value. It save a lookup table in python but also works better if there are multiple attributes of the same type.

aothms: What is lacking here is really composability. Point for discussion. How would you feel about this:

```
Given  An IfcProductDefinitionShape
And    Its values for attribute Representations
And    Its values for attribute RepresentationIdentifier

Then   The values should be unique
```

I'm not convinced whether this is a good idea. But you can see the kind of composability and reuse it gives. But it's confusing that the Given statement becomes so order-dependent and so dramatically alters the set of instances/values that later statements operate on.

features/GEM003_Unique-representation-identifier.feature
```diff
@@ -0,0 +1,11 @@
+@informal-proposition
+@GEM
+Feature: GEM003 - Unique Representation Identifer
+
+  Scenario: No Shape Representation identifier shall be used twice within the product representation of an element 
+
+      Given An IfcProductDefinitionShape
```
aothms, line 7: It might make sense to start at the IfcProduct, which has attribute Representations of type IfcProductDefinitionShape. The Product (wall, column, ...) is much more recognizable to the end user.

features/GEM003_Unique-representation-identifier.feature
```diff
@@ -0,0 +1,11 @@
...
+        And The element has more than 1 instance(s) of IfcShapeRepresentation
```
aothms, line 8: Technically speaking this is not necessary. The code and logic works also for lists of length 1.

features/steps/steps.py
```diff
@@ -58,6 +58,15 @@ class edge_use_error:
     def __str__(self):
         return f"On instance {fmt(self.inst)} the edge {fmt(self.edge)} was referenced {fmt(self.count)} times"
 
+@dataclass
+class representation_value_error:
+    inst: ifcopenshell.entity_instance
+    duplicate_values: str
```
aothms, line 64: I think it's clearer to make duplicate_value singular.

So that you get:
- Inst has multiple Representations for Identifier Body: #123=asdasd #1233123=asdasd
- Inst has multiple Representations for Identifier Axis: #456=asdasd #456456456=asdasd
- ....

features/steps/steps.py
```diff
@@ -202,6 +229,28 @@ def step_impl(context, constraint, num, entity):
...
 
 
+@then('Each instance of {entity} has a unique value for the attribute {attribute}')
+def step_impl(context, entity, attribute):
+    ent_attr = {'IfcShapeRepresentation':'Representations'}
+    assert entity in ent_attr
+    attr_get = ent_attr[entity]
+
+    errors = []
+    if context.instances:
```
aothms, line 239: The if statement is not necessary. You can iterate over an empty list. As long as `context.instances` is not None which we can make sure it's not.

features/steps/steps.py
```diff
@@ -202,6 +229,28 @@ def step_impl(context, constraint, num, entity):
...
+        for inst in context.instances:
+            attribute_intances = getattr(inst, attr_get, [])
+            duplicate_values = [
```
aothms, line 242: I would say: compute `attribute_values = list(map(lambda i: getattr(i, attribute, []), attribute_intances))` only once.

aothms (changes requested):

features/GEM003_Unique-representation-identifier.feature
```diff
@@ -0,0 +1,18 @@
...
+
+      #   Then Each instance of IfcShapeRepresentation has a unique value for the attribute RepresentationIdentifier
+
+  Scenario: No Shape Representation identifier shall be used twice within the product representation of an element 
+
+      Given An IfcProduct
+        And Its values for attribute Representation
+        And Its values for attribute Representations
+
+        Then The element must have unique values for the attribute RepresentationIdentifier
```
pjanck, line 18: I'm confused: 

1. `IfcProduct` does not have `Representations` attribute. What exactly is being checked here?
2. _Given an `IfcProduct`_, but _The **element** must have ..._ - provided, that IFC schema does have an `IfcElement`, should we perhaps choose a different word here to diminish the possibility of misunderstanding? Especially given the scenario title: _**product** representation of an **element**_.
3. `IfcProduct` does not have `RepresentationIdentifier` attribute. Similar question to 1..

features/steps/steps.py
```diff
@@ -201,6 +253,54 @@ def step_impl(context, constraint, num, entity):
...
+    errors = []
+    for inst in context.instances:
+        attribute_intances = getattr(inst, attr_get, [])
+        instance_value_pair = list(map(lambda i: (i, getattr(i, attribute, random_string())), attribute_intances))  #[instance, value),(i,v),..]
+        duplicate_values = [value for (value,count) in Counter([v[1] for v in instance_value_pair]).items() if count > 1]
+        if len(duplicate_values):
+            false_instances = [i[0] for i in instance_value_pair if i[1] in duplicate_values]
+            errors.append(representation_value_error(inst, duplicate_values, false_instances))
+
+    handle_errors(context, errors)
```
pjanck, line 302: This _then_ is commented out - should the function remain in code? Is it called anywhere?

features/steps/steps.py
```diff
@@ -173,6 +189,42 @@ def step_impl(context, attribute, value):
...
+
+@given('The element has {constraint} {num:d} instance(s) of {entity}')
+def step_impl(context, constraint, num, entity):
+    ent_attr = {'IfcShapeRepresentation':'Representations'}
+    assert entity in ent_attr
+    attr = ent_attr[entity]
+
+    stmt_to_op = {"at least": operator.ge, "more than": operator.gt}
+    assert constraint in stmt_to_op
+    op = stmt_to_op[constraint]
```
pjanck, line 219: Can we have a specialized function that resolves these? I feel like I've seen a construct like this too many times already ...

pjanck (changes requested):
Trying to understand python, please bear with me :)

features/GEM003_Unique-representation-identifier.feature
Ghesselink: The following IA is checked: https://standards.buildingsmart.org/documents/Implementation/IFC_Implementation_Agreements/CV-2x3-171.html
 `Any product representation of an element may have multiple shape representations. However there shall be no two or more IfcShapeRepresentations having the same RepresentationIdentifier within one IfcProductDefinitionShape.`

There was a discussion about this (in https://github.com/buildingSMART/ifc-gherkin-rules/issues/22). In short, with making gherkin statements there is often a tradeoff between readability and reuse. A more understandable statement would be, for example
```
Given an IfcProduct
 Then Its representations should have unique identifiers
```
In Python code/IFC we would have to navigate from the IfcProduct (a wall, a column, etc) to the shape representations and then check if the IA is met. This works, but we will write it once and never use it again. Modular statements as the ones above lead to the same result (an IfcWall for example has a representation IfcProductDefinitionShape, these have representations as IfcShapeRepresentation and these will have values RepresentationIdentifier), but will be way more suitable for reuse as rules will probably often be some attributes being checked for a certain condition. This will then save development time and lead to cleaner code. 
The only thing is that we will have to add some additional comments/description for when the statements are not completely self-explanatory, which is the case for this rule as we see now ..

features/GEM003_Unique-representation-identifier.feature
```diff
@@ -2,17 +2,11 @@
...
 
-  # Scenario: No Shape Representation identifier shall be used twice within the product representation of an element 
-
-      # Given An IfcProductDefinitionShape
-      #   And The element has more than 1 instance(s) of IfcShapeRepresentation
-
-      #   Then Each instance of IfcShapeRepresentation has a unique value for the attribute RepresentationIdentifier
-
-  Scenario: No Shape Representation identifier shall be used twice within the product representation of an element 
+  Scenario: A Shape Representation identifier must not be used twie within the product representation of an IfcProduct element
```
pjanck, line 5:
```suggestion
  Scenario: A Shape Representation identifier must not be used twice within the product representation of an IfcProduct element
```

features/GEM003_Unique-representation-identifier.feature
```diff
@@ -2,17 +2,11 @@
...
 
       Given An IfcProduct
         And Its values for attribute Representation
         And Its values for attribute Representations
+        And Its values for attribute RepresentationIdentifier
```
pjanck, line 10: I still cannot wrap my head around:

> *its* values for attribute *Representations* or *RepresentationIdentifier*

given that an IfcProduct has neither of them. 

If it is meant as: go down the attributes tree and find the entity with the fitting attribute, it doesn't seem to be very clear what is being checked if the attribute name is non-unique.

---
Scenario: Imagine ensuring 3D geometry on an IfcProduct:

    ....
    And  Its values for attribute Dim 
    Then The values must be 3

The file would have a product like:
IfcProduct
- IfcProduct.Representation = IfcProductDefinitionShape
- IfcProduct.Representation.Representations[0] = IfcShapeRepresentation
- IfcProduct.Representation.Representations[0].Items[0] = IfcExtrudedAreaSolid with **SELF/IfcSolidModel.Dim = 3**
- IfcProduct.Representation.Representations[0].Items[0].SweptArea = IfcArbitraryClosedProfileDef
- IfcProduct.Representation.Representations[0].Items[0].SweptArea.OuterCurve = IfcCurve with **IfcCurve.Dim = 2** (see IfcArbitraryClosedProfileDef.WR1)
- IfcProduct.Representation.Representations[0].Items[0].ExtrudedDirection = IfcDirection with **IfcDirection.Dim = HIINDEX(DirectionRatios), which may be 2 or 3**

What is the expected behavior? Should the product fail or pass the rule?

pjanck (changes requested):

features/GEM003_Unique-representation-identifier.feature
aothms: I understand your confusion. `And` is not meant to be read here as a boolean and, but more a sequential chaining, think of it as tree/graph traversal. So:

```
Given An IfcProduct                                                # IfcProduct[]
        And Its values for attribute Representation                # IfcProduct[].Representation
        And Its values for attribute Representations               # IfcProduct[].Representation.Representations[]
        And Its values for attribute RepresentationIdentifier      # IfcProduct[].Representation.Representations[].RepresentationIdentifier
```

Do you have better ideas? We're also thinking about referencing concept templates from gherkin.

pjanck: I see. *its* applies to the object named in the expression above it. Thank you for the explanation.

A question, though: how to write the rule if one wishes to check the values of two attributes of the same object? For example, 

- `ExtrudedDirection` and `SweptArea` from my example above?
- `RepresentationIdentifier` and `RepresentationType` from your example above?

Would one need to write two separate scenarios?

aothms: > how to write the rule if one wishes to check the values of two attributes 

It would be possible to recover multiple "bindings" from the stack frames. See #22 for an exploration. We'd have to connect the stack frames with the step definitions to build a mapping of attribute name -> value. Or store it as a dictionary in the context to begin with. So we might have to modify the approach a bit, but we're not too far away.

We can also think of a way to allow for explicit naming:

```
Given   An IfcProduct
  And   Its values for attribute Representation
  And   Its values for attribute Representations
  And   Its values for attribute RepresentationIdentifier to be referenced as RepId
  And   Its values for attribute RepresentationType to be referenced as RepType
  
 Then   RepId should be equal to RepType
```

(silly example obviously).

aothms (approved):

features/steps/steps.py
```diff
@@ -34,6 +36,11 @@ def get_inst_attributes(dc):
         yield 'inst_type', dc.inst.is_a()
         yield 'inst_id', dc.inst.id()
 
+def random_string():
```
aothms, line 39: What's the purpose of this?

aothms (dismissed):

features/steps/steps.py
Ghesselink: I think it's there by accident with going back and forth to ALB002 for testing with pyparsing. Deleted it

features/steps/steps.py
```diff
@@ -173,6 +212,36 @@ def step_impl(context, attribute, value):
...
+
+@given('The element has {constraint} {num:d} instance(s) of {entity}')
+def step_impl(context, constraint, num, entity):
+    ent_attr = {'IfcShapeRepresentation':'Representations'}
+    assert entity in ent_attr
+    attr = ent_attr[entity]
+
+    stmt_to_op = {"at least": operator.ge, "more than": operator.gt}
+    assert constraint in stmt_to_op
+    op = stmt_to_op[constraint]
```
pjanck, line 236: Please merge the changes from `main` and correct accordingly.

pjanck (changes requested):

features/steps/steps.py
Ghesselink: I was wrong, it was added in this commit https://github.com/buildingSMART/ifc-gherkin-rules/pull/21/commits/cca7b3ae03ac411c267c96efaa1b49e189685876  but unused due to some changes later on.

features/steps/steps.py
```diff
@@ -285,6 +382,71 @@ def step_impl(context, constraint, num, entity):
...
+
+    if getattr(context, 'applicable', True):
+        stack_tree = list(filter(None, list(map(lambda layer: layer.get('instances'), context._stack))))
+        instances = [context.instances] if within_model else context.instances
+
+        for i, values in enumerate(instances):
+            if not values:
+                continue
+
+            msg = value_error_msg(identical_or_unique=identical_or_unique, attribute=getattr(context, 'attribute', 'None'))
```
aothms, line 437: I'm a bit confused by all of this. You need to help me a bit:

- You overwrite `msg.values`.
- I would prefer to have the `value_error_msg` class to be initialized completely. Now you have everything with default values as "None". But they are not really optional you just supply them later. This is confusing to the reader. Rather use local variables here, and initialize the error class entirely in a single call.
- I'd always expect a function named `convert_values()` to be "pure", i.e no side effects, but this function sets include_relating_entities on the context which is confusing. It's also unclear what this triggers because there are no comments in this section.
- convert_values could also be named a bit more descriptive, I'd imagine in other steps we'll be doing also a lot of 
converting.
- the error class is also very complex, maybe we've reached a point that it needs to be broken down into several classes. I think the formatting within them should be purely obvious.

aothms (changes requested):

features/steps/steps.py
```diff
@@ -250,6 +317,36 @@ def step_impl(context, attribute, value):
...
+        return fn(values)
+
+@given('Its attribute {attribute}')
+def step_impl(context, attribute):
+    context._push()
+    context.instances = map_state(context.instances, lambda i: getattr(i, attribute, None))
+    setattr(context, 'instances', context.instances)
+    setattr(context, 'attribute', attribute)
+
+@given('The element has {constraint} {num:d} instance(s) of {entity}')
```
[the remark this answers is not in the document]
Ghesselink: Does not seem so, I've made an issue for this: https://github.com/buildingSMART/ifc-gherkin-rules/issues/34

features/steps/steps.py
Ghesselink: It seems initializing the dataclass first and supplying data later leads to more confusion than clarity. It's changed in https://github.com/buildingSMART/ifc-gherkin-rules/commit/a3c7871507f07fbcb83cc95d839259780d7f4b7e as well as https://github.com/buildingSMART/ifc-gherkin-rules/commit/6ba2c0a43788bdefacf2905347e7853849d02814

The dataclass is also simplified, and the logic is moved to the 'Then' statements that use this dataclass. This, I think, makes it more obvious to read.

features/steps/steps.py
```diff
@@ -250,6 +317,36 @@ def step_impl(context, attribute, value):
...
+def map_state(values, fn):
+    if isinstance(values, (tuple, list)):
+        return type(values)(map_state(v, fn) for v in values)
+    else:
+        return fn(values)
+
+@given('Its attribute {attribute}')
+def step_impl(context, attribute):
+    context._push()
+    context.instances = map_state(context.instances, lambda i: getattr(i, attribute, None))
```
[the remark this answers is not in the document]
Ghesselink: Answered in GRF001

features/steps/steps.py
```diff
@@ -250,6 +317,36 @@ def step_impl(context, attribute, value):
...
+    setattr(context, 'instances', context.instances)
```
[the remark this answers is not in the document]
Ghesselink: Implemented

features/steps/steps.py
```diff
@@ -331,16 +407,25 @@ def step_impl(context, relationship_type, entity):
...
         applicable = conditional_lowercase(get_mvd(context.model)) in values
     elif field == "Schema Identifier":
         applicable = context.model.schema.lower() in values
     else:
         raise NotImplementedError(f'A file with "{field}" is not implemented')
-
+ 
     context.applicable = getattr(context, 'applicable', True) and applicable
+    
+@given('Its values')
```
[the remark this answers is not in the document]
Ghesselink: Done, in https://github.com/buildingSMART/ifc-gherkin-rules/pull/21/commits/46604b8ec559170eedef226f9ec276564824f6de

aothms (approved):

PR #23, opened 2022-12-20
stmt_to_op in function

Ghesselink:
See discussion on GEM003; https://github.com/buildingSMART/ifc-gherkin-rules/pull/21#pullrequestreview-1222883274

This functionality is often repeated, so wrote a small function for it. 
If approved, I'll also change it in non-merged branches (alb001/2/gem003) to avoid merge conflicts.

### DIFF
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -2,6 +2,7 @@ import ast
 import json
 import typing
 import operator
+import re
 
 from collections import Counter
 from dataclasses import dataclass
@@ -35,7 +36,9 @@ def get_inst_attributes(dc):
         yield 'inst_id', dc.inst.id()
 
 def stmt_to_op(statement):
+    statement = statement.replace('is', '').strip()
     stmt_to_op = {
+        '': operator.eq, # a == b
         "equal to": operator.eq, # a == b
         "exactly": operator.eq, # a == b
         "not": operator.ne, # a != b
@@ -254,9 +257,7 @@ def step_impl(context, constraint, num, entity):
 
 @then('The {related} shall be assigned to the {relating} if {other_entity} {condition} present')
 def step_impl(context, related, relating, other_entity, condition):
-    stmt_to_op = {"is": operator.eq, "is not": operator.ne}
-    assert condition in stmt_to_op
-    pred = stmt_to_op[condition]
+    pred = stmt_to_op(condition)
 
     op = lambda n: not pred(n, 0)
 
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -36,12 +36,13 @@ def get_inst_attributes(dc):
 
 def stmt_to_op(statement):
     stmt_to_op = {
-        "is": operator.eq, # a == b
+        "equal to": operator.eq, # a == b
         "exactly": operator.eq, # a == b
-        "is not": operator.ne, # a != b
+        "not": operator.ne, # a != b
         "at least": operator.ge, # a >= b
         "more than": operator.gt, # a > b
-        "at most": operator.le # a <= b
+        "at most": operator.le, # a <= b
+        "less than": operator.lt # a < b
     }
     assert statement in stmt_to_op
     return stmt_to_op[statement]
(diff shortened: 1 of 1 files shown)


features/steps/steps.py
```diff
@@ -34,6 +34,18 @@ def get_inst_attributes(dc):
         yield 'inst_type', dc.inst.is_a()
         yield 'inst_id', dc.inst.id()
 
+def stmt_to_op(statement):
+    stmt_to_op = {
+        "is": operator.eq, # a == b
+        "exactly": operator.eq, # == b
```
pjanck, line 40:
```suggestion
        "exactly": operator.eq, # a == b
```

pjanck (changes requested):
`less than` is missing.

aothms:
I would also standardize on whether the include the is-verb (leave it out probably)

- is -> equal to
- is not -> not

features/steps/steps.py
```diff
@@ -246,6 +257,7 @@ def step_impl(context, related, relating, other_entity, condition):
     stmt_to_op = {"is": operator.eq, "is not": operator.ne}
     assert condition in stmt_to_op
     pred = stmt_to_op[condition]
+
```
pjanck, line 260: How come this no longer calls the function?

pjanck (changes requested):
While I understand the idea of removing `is` from the statements, I don't think the chosen solution is the best. Especially given the example that had to be reverted.

Proposal: remove `is` within the `stmt_to_op` function as the first action.

features/steps/steps.py
Ghesselink: That will cause an AssertionError I think, as the 'is' comes from the sentence ".. shall be assigned to {relating} if {other_entity} {condition"} present'
How do you feel about putting the stmt_to_op dictionary into the function call? This way, the sentences are not dependent on the stmt_to_op function, it's called with only one line of code and it's easy to see for someone looking into the code what is happening.  
```
>>> import operator
>>> def get_op(statement, stmt_to_op = {}):
...     assert statement in stmt_to_op
...     return stmt_to_op[statement]
... 
>>> pred = get_op("is", {"is":operator.eq, "is not": operator.ne})
>>> pred
<built-in function eq>
```

pjanck: > How do you feel about putting the stmt_to_op dictionary into the function call?

Strongly against.

**Proposal 1**

Consider calling

> `statement = statement.replace("is", "")`

or sth similar as the first call within `stmt_to_op` function (probably need to account for spaces as well).

Add `"" : operator.eq, # a == b` to the `stmt_to_op` dictionary. In this way, a lot more of cases should be supported.

**Proposal 2**

Add `"is"` and all other permutations to the dictionary `stmt_to_op` as well. Similar result as Proposal 1.

pjanck: > `statement = statement.replace('is','').strip()`

I like this very much. This also extends the function to cover cases like "is equal to" to map to "equal to" and thus make a very versatile function.

> Then we get to the same situation where we started, or do I miss something?

Proposal 2 was just an alternative, I didn't like it either. But would still prefer it over a dictionary parameter to the function - in that case there is no reuse of the logic: "some string" maps to `operator` and we truly are where we started :)

Ghesselink: No, there would be a need to provide a dictionary parameter every time indeed. 
Proposal 1 will be easy to work with I think, it's implemented in https://github.com/buildingSMART/ifc-gherkin-rules/pull/23/commits/5c15d0e8d0a6e1278ef5b515d5f6cc052b8d511d

pjanck (commented):
This now looks a lot cleaner and reusable. <3

features/steps/steps.py
```diff
@@ -34,6 +35,21 @@ def get_inst_attributes(dc):
         yield 'inst_type', dc.inst.is_a()
         yield 'inst_id', dc.inst.id()
 
+def stmt_to_op(statement):
+    statement = statement.replace('is', '').strip()
```
pjanck, line 39: Should we also account for `are` here?

pjanck (commented):
An idea:

features/steps/steps.py
Ghesselink: We could, In the case  below(for which is it used now) it is quite simple. Either the entity is or is not present.
`The X shall be assigned to the Y if Z is present`

But in the case of 'are', I can imagine some complications
`X shall be assigned to Y if all of the following entities are present: A, B, C`
or 
`X shall be assigned to Y if one of the following entities are present:A, B, C `
The second example is similar to the current python implementation of the stmt_to_op function, the first one would be different however.

aothms (approved):

Issue #18, opened 2022-12-06
A question on entities order


ciro-vendrame:
https://github.com/buildingSMART/ifc-gherkin-rules/blob/81b07c03b8c9726b4c0b580eb0efd63c1899ab2e/features/ALB002_Alignment-layout.feature#L27

Is there a specific order the SW vendors expect these entities?
I also suppose there is already an agreement in place for which IfcReferent shall be nested in IfcAlignment and not in IfcAlignment Horizontal, am I right?

aothms:
Currently this statement is only about the allowed types irrespective about order. It is implemented as a (unordered) set comparison https://github.com/buildingSMART/ifc-gherkin-rules/pull/6/files#diff-cf027576a6812e3403bd3d9579d83abfaa74a41607a4f3c986398ef8492a5d3dR270

I'll assign this to @evandroAlfieri if the unordered nature of this constraint needs to be made explicit in gherkin and about the question on referent.

evandroAlfieri:
Good point Ciro. For this and other reasons, I'd follow the suggestion of @pjanck here: https://github.com/buildingSMART/ifc-gherkin-rules/pull/6#discussion_r1036142227
**Let's move that sentence in a dedicated ALB rule (e.g., ALB003_Allowed entities nested in Alignment).**

I'll discuss this specific topic with vendors in the IF. I guess a comprehensive rule should say:
- what are the allowed entities that can be nested in alignment (the full list is the one in the current sentence)
- what order shall they have, when present

If I remember correctly, it has been said once that:
- alignment layouts must be in one RelNests relationship, with order H-V-C
- referents shall be in another RelNests than the one of the alignment layouts
- other alignment shall be in another RelNests than the one of alignment layouts, and referents

I'll confirm with SV and come back to you. Meanwhile, @Ghesselink can we take this step out of the ALB002 rule?

Ghesselink:
The scenario is removed in https://github.com/buildingSMART/ifc-gherkin-rules/pull/6/commits/e5efda347d083e88c593b99a4324a8c4a28735ec 
I've saved the python code, gherkin feature and (code to generate-) testfiles locally. The assigned person(s) for this new created rule (if not myself) can contact me if they'd like to have it.

Ghesselink:
Moved to https://github.com/buildingSMART/ifc-gherkin-rules/pull/26

PR #26, opened 2022-12-21
Alb003 - Alignment Nesting

Ghesselink:
Continuation of Issue https://github.com/buildingSMART/ifc-gherkin-rules/issues/18


features/ALB003_Alignment-directions.feature
```diff
@@ -0,0 +1,11 @@
+@implementer-agreement
+@ALB
+Feature: ALB002 - Alignment Directions
```
pjanck, line 3: The Feature disagrees with filename.

pjanck (changes requested):

features/steps/steps.py
```diff
@@ -75,9 +75,20 @@ def __str__(self):
 class instance_structure_error:
     related: ifcopenshell.entity_instance
     relating: ifcopenshell.entity_instance
+    relationship_type: str
+    optional_values: field(default_factory=dict)
 
     def __str__(self):
-        return f"The instance {fmt(self.related)} is assigned to {fmt(self.relating)}"
+        self.pos_neg = 'is not' if do_try(lambda: self.optional_values['condition'], '') == 'must' else 'is'
```
aothms, line 82: This is a bit "muggenziften", but try not to modify `self` in `__str__`-like functions. It's not really a concern here, but you wouldn't expect. Below is a silly example, but hopefully it demonstrates how surprising it is when functions like `__str__()` affect the state (they are called side effects). In other languages you would expect these functions to be marked `const` so they can't affect state. Python doesn't have that. In short: just define local variables `pos_neg` `directness` don't modify self.

```
>>> class A:
...   value = 1
...   def __str__(self):
...     v = str(self.value)
...     self.value *= 2
...     return v
...
>>>
>>>
>>> a = A()
>>>
>>> a.value
1
>>> print(a)
1
>>> a.value
2
>>> print(a)
2
>>> print(a)
4
```

features/steps/steps.py
```diff
@@ -257,7 +268,7 @@ def step_impl(context, related, relating, other_entity, condition):
             for inst in context.model.by_type(related):
                 for rel in getattr(inst, 'Decomposes', []):
                     if not rel.RelatingObject.is_a(relating):
-                        errors.append(instance_structure_error(inst, rel.RelatingObject))
+                        errors.append(instance_structure_error(inst, [rel.RelatingObject], 'assigned to',{}))
```
aothms, line 271: You already have `default_factory=dict`. I would leave it out here. Looks prettier.

features/steps/steps.py
```diff
@@ -283,4 +294,19 @@ def step_impl(context, representation_id):
...
+    allowed_entity_types = other_entities.split(', ')
+
+    errors = []
+    if getattr(context, 'applicable', True):
+        for inst in context.model.by_type(entity):
+            nested_entities = [i for rel in inst.IsNestedBy for i in rel.RelatedObjects]
+            nested_entity_types = [i.is_a() for i in nested_entities]
+            if not set(nested_entity_types) <= set((allowed_entity_types)):
+                differences = list(set(nested_entity_types) - set(allowed_entity_types))
+                errors.append(instance_structure_error(inst, [i for i in nested_entities if i.is_a() in differences], 'nested by',{}))
```
aothms, line 310: Same. You already have default_factory=dict. I would leave it out here. Looks prettier.

But what is the use of `optional_values` then actually?

features/steps/steps.py
```diff
@@ -283,4 +294,19 @@ def step_impl(context, representation_id):
                     errors.append(representation_shape_error(inst, representation_id))
     
     handle_errors(context, errors)
+
+@then('Each {entity} may be nested by only the following entities: {other_entities}')
+def step_impl(context, entity, other_entities):
+
+    allowed_entity_types = other_entities.split(', ')
```
aothms, line 301: I would prefer:

    set(map(str.strip(), other_entities.split(','))

- Remove white space after splitting to be a bit more whitespace agnostic.
- Convert to set immediately.

features/steps/steps.py
```diff
@@ -283,4 +294,19 @@ def step_impl(context, representation_id):
...
+
+    errors = []
+    if getattr(context, 'applicable', True):
+        for inst in context.model.by_type(entity):
+            nested_entities = [i for rel in inst.IsNestedBy for i in rel.RelatedObjects]
+            nested_entity_types = [i.is_a() for i in nested_entities]
```
aothms, line 307: Same here. Convert to set immediately.

    set(i.is_a() for i in nested_entities)

features/steps/steps.py
```diff
@@ -283,4 +294,19 @@ def step_impl(context, representation_id):
...
+            if not set(nested_entity_types) <= set((allowed_entity_types)):
+                differences = list(set(nested_entity_types) - set(allowed_entity_types))
```
aothms, line 309: No need to convert to list. With the comments above taken into account (sets already) then `nested_entity_types - allowed_entity_types` suffices.

aothms (changes requested):
Great. Only minor comments I think. I like how you've structured it using set logic.

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
...
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python: module",
+            "type": "python",
+            "request": "launch",
+            "module": "behave",
+            "justMyCode": true,
+            "args": ["-D","input='/home/geert/Documents/validation_service/buildingSMART/gherkin/alb003/ifc-gherkin-rules/test/files/alb003/fail-alb003-wrong_direction.ifc'"]
```
pjanck, line 13: I'd prefer not to have absolute paths in public projects.

features/ALB003_Alignment-directions.feature
```diff
@@ -0,0 +1,11 @@
+@implementer-agreement
+@ALB
+Feature: ALB003 - Alignment Directions
+
+  Scenario: Agreement on nested directions of IfcAlignment
```
pjanck, line 5: I probably don't understand, which meaning of *direction* is chosen here. (In my head, an alignment could have a direction forward/backwards; or perhaps going to the left/right/up/down ...) I'd propose to reword to *Alignment Nesting* or similar.

pjanck (changes requested):
I need to go through the python code with a fresher head.

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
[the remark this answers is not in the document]
Ghesselink: Yes, I use it for debugging and it was .gitignore in the other branches, but not in this one apparently. The folder is deleted (and mentioned in .gitignore) now.

features/steps/steps.py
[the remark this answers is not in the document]
Ghesselink: That is indeed a way more simple solution

Ghesselink: That is good to know, I would not expect that indeed. It is modified for now and you probably will save a future-me some frustrations when this occurs on another occasion.

features/steps/steps.py
```diff
@@ -75,9 +75,20 @@ def __str__(self):
...
+    relationship_type: str
+    optional_values: field(default_factory=dict)
 
     def __str__(self):
-        return f"The instance {fmt(self.related)} is assigned to {fmt(self.relating)}"
+        self.pos_neg = 'is not' if do_try(lambda: self.optional_values['condition'], '') == 'must' else 'is'
+        self.directness = do_try(lambda: self.optional_values['directness'],'')
+        
+        if len(self.relating):
+            if len(self.relating) > 1:
```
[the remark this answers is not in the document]
Ghesselink: It was because the wording is slightly different when it is a single case as opposed to more entities in the error log. The message in this error log is changed now, it's less confusing and will also take into consideration (future) cases in which 'pos_neg' and 'directness' (now in ALB001) are linked to more entities.

features/steps/steps.py
Ghesselink: I have tried that, but ran into a 'class expected x arguments but got x-1 arguments' type of error. I don't really understand why and am looking into the dataclasses documentation/stackoverflow, but so far it is not clearer (maybe by modifying the __init__ function ourselves?)

features/steps/steps.py
Ghesselink: The use of optional_values is to avoid merge conflicts between ALB001 and ALB002/ALB003 https://github.com/buildingSMART/ifc-gherkin-rules/commits/alb001_alignment-in-spatial-structure
We used different dataclasses and it was better to merge them. @JTurbakiewicz used sentences as
`A is/is not directly/indirectly contained in B` 
'is/isnot' and 'directly/indirectly' play a role in ALB001, but not in the current PR.

features/steps/steps.py
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/commit/433fe4ca8d9318574873a26542a6f321856aacd1

features/steps/steps.py
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/commit/433fe4ca8d9318574873a26542a6f321856aacd1

features/steps/steps.py
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/commit/433fe4ca8d9318574873a26542a6f321856aacd1

.vscode/launch.json
Ghesselink: I used this launch.json file for debugging in VScode on my local machine. In other branches, I made sure for git to ignore it by putting it into .gitignore. It's deleted and .gitignore modified now

features/ALB003_Alignment-directions.feature
Ghesselink: I would think that just 'alignment nesting' will lead to confusions with the scenarios in ALB002 (they are all about alignment nesting). This is very specific for IfcAlignmentHorizontal, IfcAlignmentVertical and IfcAlignmentCant. There is not an overlapping term for these three in IFC, and the best wording I could come up with is 'direction'. Do you maybe know a good overlapping word/term (or a couple) for these three cases?

pjanck: > This is very specific for IfcAlignmentHorizontal, IfcAlignmentVertical and IfcAlignmentCant.

The rule includes `IfcReferent` and `IfcAlignment` as well. Please double check.

> Do you maybe know a good overlapping word/term (or a couple) for these three cases?

Layout. Horizontal / vertical / cant are the three layouts of an alignment.

features/steps/steps.py
Ghesselink: Fixed https://github.com/buildingSMART/ifc-gherkin-rules/commit/5267f10a101c67fa8d4ecdf01b761d9ce4ba48c0
Had to change 
`optional_values: field(default_factory=dict)`
to
`optional_values: dict = field(default_factory=dict)`

aothms: Oh, I missed that, correct.

features/ALB003_Alignment-directions.feature
Ghesselink: > Layout. Horizontal / vertical / cant are the three layouts of an alignment.
That makes sense, changed the names in https://github.com/buildingSMART/ifc-gherkin-rules/pull/26/commits/1a08989b12b035aa11668d0f57e94ff821f2f078

> The rule includes IfcReferent and IfcAlignment as well. Please double check.
I've removed IfcAlignment from it from now. At the moment, it is not very clear to me what to do with this rule. There was a discussion at https://github.com/buildingSMART/ifc-gherkin-rules/issues/18 with this idea:

```
    alignment layouts must be in one RelNests relationship, with order H-V-C
    referents shall be in another RelNests than the one of the alignment layouts
    other alignment shall be in another RelNests than the one of alignment layouts, and referents
```
I've asked for a confirmation of this. Do you have any thoughs/ideas?

features/ALB003_Alignment-directions.feature
```diff
@@ -1,11 +1,11 @@
...
-Feature: ALB003 - Alignment Directions
+Feature: ALB003 - Alignment Layout
 
-  Scenario: Agreement on nested directions of IfcAlignment
+  Scenario: Agreement on nested Layout of IfcAlignment
 
       Given a file with Schema Identifier "IFC4X3"
 
-       Then Each IfcAlignment may be nested by only the following entities: IfcAlignmentHorizontal, IfcAlignmentVertical, IfcAlignmentCant, IfcReferent, IfcAlignment
+       Then Each IfcAlignment may be nested by only the following entities: IfcAlignmentHorizontal, IfcAlignmentVertical, IfcAlignmentCant, IfcReferent
```
pjanck, line 9: A Referent is not a layout of an alignment. I believe this is a mistake.

pjanck (changes requested):

features/ALB003_Alignment-directions.feature
pjanck: > I've asked for a confirmation of this. Do you have any thoughs/ideas?

Good question! I don't have a definite answer and am in no position to give one. This is something that should be derived from a requirement analysis, IMHO. I like the proposed idea, however can't confirm its validity.

aothms (approved):

Issue #22, opened 2022-12-16
Comprehensive vs modular statements and error reporting


aothms:
## Background:

There's the constant trade-off between small modular statements and big comprehensive statement:

### Small and modular

```
Given An IfcProduct
  And It's attribute Representation
  And It's attribute Representations
  And It's attribute RepresentationIdentifier
 Then The values should be unique
```

Pros:
 - Reuse of statements obvious
 - High flexibility in writing rules from existing steps
 
Cons:
 - Requires greater knowledge of IFC schema relationships
 - Reads a bit funky

### Comprehesive

```
Given an IfcProduct
 Then Its representations should have unique identifiers
```

Pros:
 - Readability for non-ifc experts
 - Error reporting can be tailored to specific context (because it's implemented as a specific step)

Cons:
 - No reuse
 - No flexibility

## Potential middle ground

Behave (the gherkin python impl) has this concept of `stack frames`. This allows us to go back in time on the context variable.

#### features\f.feature

```
Feature: Test

  Scenario: Test

      Given A = 1
        And B = 1
        And A = 2

       Then A == 3
```

#### features\steps\steps.py

```
@given('{var} = {val}')
def step_impl(context, var, val):
    # push to a new stack frame
    context._push()
    setattr(context, var, val)
    
@then('{var} == {val}')
def step_impl(context, var, val):
    import pprint
    pprint.pprint(context._stack)
    assert getattr(context, var) == val
```

#### Output:

```
[{'@cleanups': [], 'A': '2', 'table': None, 'text': None},
 {'@cleanups': [], 'B': '1', 'table': None, 'text': None},
 {'@cleanups': [], 'A': '1', 'table': None, 'text': None},
 {'@cleanups': [],
  '@layer': 'scenario',
  'log_capture': <LoggingCapture (INFO)>,
  'scenario': <Scenario "Test">,
  'stderr_capture': <_io.StringIO object at 0x00000173CB1FF430>,
  'table': None,
  'tags': {'informal-proposition', 'GEM'},
  'text': None},
...
```

#### Summary

You can see how the overwritten value of A is still recoverable in a previous stack frame. Can we use this for reporting?

Ghesselink:
I have tried a first attempt to implement some gherkin statements with these stack frames with the following statements https://github.com/buildingSMART/ifc-gherkin-rules/pull/21/commits/d1c1dfbc2fb76947bffece70c28507a52ff5f79d . 
```
[...]

PR #27, opened 2022-12-28
OJP001 - Relative placement for elements aggregated to another element

JTurbakiewicz:


features/steps/steps.py
```diff
@@ -186,6 +202,18 @@ def step_impl(context, field, values):
 
     context.applicable = getattr(context, 'applicable', True) and applicable
 
+
+@given('The {entity} is a part of another {other_entity} through the relationship {relationship}')
```
aothms, line 206: I think this is a good attempt at formalizing the statement we were discussing earlier on separating the direction (is part of / has parts) and the type of the relationship. Should we make the direction a variable too?

features/steps/steps.py
```diff
@@ -186,6 +202,18 @@ def step_impl(context, field, values):
...
+def step_impl(context, entity, other_entity, relationship):
+    relationships = context.model.by_type(relationship)
+    instances = []
+    for rel in relationships:
+        if rel.RelatingObject.is_a(other_entity):
```
aothms, line 211: This end of the relationship is not always called `RelatingObject` see table below (only for IFC4).

A quick hack would be:

```
>>> import ifcopenshell
>>> f = ifcopenshell.file()
>>> rel = f.createIfcRelAggregates()
>>> [d for d in dir(rel) if d.startswith('Relating')][0]
'RelatingObject'
```

But we can also build a lookup table.

entity | attribute | cardinality | ordered | type
-- | -- | -- | -- | --
IfcRelAssigns | RelatedObjects | 1..* | n | IfcObjectDefinition
IfcRelAssignsToActor | RelatingActor | 1..1 |   | IfcActor
IfcRelAssignsToControl | RelatingControl | 1..1 |   | IfcControl
IfcRelAssignsToGroup | RelatingGroup | 1..1 |   | IfcGroup
IfcRelAssignsToProcess | RelatingProcess | 1..1 |   | IfcProcess,IfcTypeProcess
IfcRelAssignsToProduct | RelatingProduct | 1..1 |   | IfcProduct,IfcTypeProduct
IfcRelAssignsToResource | RelatingResource | 1..1 |   | IfcResource,IfcTypeResource
IfcRelAssociates | RelatedObjects | 1..* | n | IfcObjectDefinition,IfcPropertyDefinition
IfcRelAssociatesApproval | RelatingApproval | 1..1 |   | IfcApproval
IfcRelAssociatesClassification | RelatingClassification | 1..1 |   | IfcClassification,IfcClassificationReference
IfcRelAssociatesConstraint | RelatingConstraint | 1..1 |   | IfcConstraint
IfcRelAssociatesDocument | RelatingDocument | 1..1 |   | IfcDocumentInformation,IfcDocumentReference
IfcRelAssociatesLibrary | RelatingLibrary | 1..1 |   | IfcLibraryInformation,IfcLibraryReference
IfcRelAssociatesMaterial | RelatingMaterial | 1..1 |   | IfcMaterialDefinition,IfcMaterialList,IfcMaterialUsageDefinition
IfcRelConnectsElements | RelatingElement | 1..1 |   | IfcElement
IfcRelConnectsElements | RelatedElement | 1..1 |   | IfcElement
IfcRelConnectsPortToElement | RelatingPort | 1..1 |   | IfcPort
IfcRelConnectsPortToElement | RelatedElement | 1..1 |   | IfcDistributionElement
IfcRelConnectsPorts | RelatingPort | 1..1 |   | IfcPort
IfcRelConnectsPorts | RelatedPort | 1..1 |   | IfcPort
IfcRelConnectsStructuralActivity | RelatingElement | 1..1 |   | IfcElement,IfcStructuralItem
IfcRelConnectsStructuralActivity | RelatedStructuralActivity | 1..1 |   | IfcStructuralActivity
IfcRelConnectsStructuralMember | RelatingStructuralMember | 1..1 |   | IfcStructuralMember
IfcRelConnectsStructuralMember | RelatedStructuralConnection | 1..1 |   | IfcStructuralConnection
IfcRelContainedInSpatialStructure | RelatedElements | 1..* | n | IfcProduct
IfcRelContainedInSpatialStructure | RelatingStructure | 1..1 |   | IfcSpatialElement
IfcRelCoversBldgElements | RelatingBuildingElement | 1..1 |   | IfcElement
IfcRelCoversBldgElements | RelatedCoverings | 1..* | n | IfcCovering
IfcRelCoversSpaces | RelatingSpace | 1..1 |   | IfcSpace
IfcRelCoversSpaces | RelatedCoverings | 1..* | n | IfcCovering
IfcRelFillsElement | RelatingOpeningElement | 1..1 |   | IfcOpeningElement
IfcRelFillsElement | RelatedBuildingElement | 1..1 |   | IfcElement
IfcRelFlowControlElements | RelatedControlElements | 1..* | n | IfcDistributionControlElement
IfcRelFlowControlElements | RelatingFlowElement | 1..1 |   | IfcDistributionFlowElement
IfcRelInterferesElements | RelatingElement | 1..1 |   | IfcElement
IfcRelInterferesElements | RelatedElement | 1..1 |   | IfcElement
IfcRelReferencedInSpatialStructure | RelatedElements | 1..* | n | IfcProduct
IfcRelReferencedInSpatialStructure | RelatingStructure | 1..1 |   | IfcSpatialElement
IfcRelSequence | RelatingProcess | 1..1 |   | IfcProcess
IfcRelSequence | RelatedProcess | 1..1 |   | IfcProcess
IfcRelServicesBuildings | RelatingSystem | 1..1 |   | IfcSystem
IfcRelServicesBuildings | RelatedBuildings | 1..* | n | IfcSpatialElement
IfcRelSpaceBoundary | RelatingSpace | 1..1 |   | IfcExternalSpatialElement,IfcSpace
IfcRelSpaceBoundary | RelatedBuildingElement | 1..1 |   | IfcElement
IfcRelDeclares | RelatingContext | 1..1 |   | IfcContext
IfcRelDeclares | RelatedDefinitions | 1..* | n | IfcObjectDefinition,IfcPropertyDefinition
IfcRelAggregates | RelatingObject | 1..1 |   | IfcObjectDefinition
IfcRelAggregates | RelatedObjects | 1..* | n | IfcObjectDefinition
IfcRelNests | RelatingObject | 1..1 |   | IfcObjectDefinition
IfcRelNests | RelatedObjects | 1..* | y | IfcObjectDefinition
IfcRelProjectsElement | RelatingElement | 1..1 |   | IfcElement
IfcRelProjectsElement | RelatedFeatureElement | 1..1 |   | IfcFeatureElementAddition
IfcRelVoidsElement | RelatingBuildingElement | 1..1 |   | IfcElement
IfcRelVoidsElement | RelatedOpeningElement | 1..1 |   | IfcFeatureElementSubtraction
IfcRelDefinesByObject | RelatedObjects | 1..* | n | IfcObject
IfcRelDefinesByObject | RelatingObject | 1..1 |   | IfcObject
IfcRelDefinesByProperties | RelatedObjects | 1..* | n | IfcObjectDefinition
IfcRelDefinesByProperties | RelatingPropertyDefinition | 1..1 |   | IfcPropertySetDefinition,IfcPropertySetDefinitionSet
IfcRelDefinesByTemplate | RelatedPropertySets | 1..* | n | IfcPropertySetDefinition
IfcRelDefinesByTemplate | RelatingTemplate | 1..1 |   | IfcPropertySetTemplate
IfcRelDefinesByType | RelatedObjects | 1..* | n | IfcObject
IfcRelDefinesByType | RelatingType | 1..1 |   | IfcTypeObject

features/steps/steps.py
```diff
@@ -186,6 +202,18 @@ def step_impl(context, field, values):
...
+            for obj in rel.RelatedObjects:
```
aothms, line 212: See table above:

- not always named like that
- also not always an aggregate some relationships are 1..1 (see cardinality).

features/steps/steps.py
```diff
@@ -221,3 +249,41 @@ def step_impl(context, related, relating, other_entity, condition):
                         errors.append(instance_structure_error(inst, rel.RelatingObject))
 
     handle_errors(context, errors)
+
+@then('The relative placement of that {entity} must be provided by an {other_entity} entity')
+def step_impl(context, entity, other_entity):
+    if getattr(context, 'applicable', True):
+        errors = []
+        for obj in context.instances:
+            if not obj.ObjectPlacement.is_a(other_entity):
```
aothms, line 258: `obj.ObjectPlacement` may be None

features/steps/steps.py
```diff
@@ -221,3 +249,41 @@ def step_impl(context, related, relating, other_entity, condition):
...
+def step_impl(context, entity, other_entity, relationship):
+    if getattr(context, 'applicable', True):
+        errors = []
+
+        stmt_to_op = {'IfcRelAggregates': 'Decomposes'}
+        assert relationship in stmt_to_op
+
+        for entity in context.instances:
+            entity_relation = getattr(entity, stmt_to_op[relationship])[0]
+            if relationship == 'IfcRelAggregates':
```
aothms, line 274: Why do we not follow the same approach as above:

```
    for rel in relationships:
        if rel.RelatingObject ...
```

features/steps/steps.py
```diff
@@ -221,3 +249,41 @@ def step_impl(context, related, relating, other_entity, condition):
...
+    if getattr(context, 'applicable', True):
+        errors = []
+        for obj in context.instances:
+            if not obj.ObjectPlacement.is_a(other_entity):
+                errors.append(instance_placement_error(obj, other_entity, "", "", "", ""))
+        handle_errors(context, errors)
+
+
+@then(
+    'The {entity} entity must point to the {other_entity} of the container element established with {relationship} relationship')
```
aothms, line 264: `other_entity` appears unused.

aothms: I'm also thinking of whether it's necessary to repeat the `relationship` (=IfcRelAggregates) here and see what it would mean in a general case. Do you think it would also be possible to recover the IfcRelationship using the stack based approach? https://github.com/buildingSMART/ifc-gherkin-rules/issues/22 That might also help when there are multiple relationships of that type pointing to entity?

aothms (changes requested):

features/OJP001_Relative-placement-for-elements-aggregated-to-another-element.feature
```diff
@@ -0,0 +1,11 @@
...
+@OJP
+Feature: OJP001 - Relative placement for elements aggregated to another element
+
+  Scenario: Agreement on the relative placement of IfcElements being a part of another IfcElement through the relationship IfcRelAggregates
+
+      Given A file with Schema Identifier "IFC2X3"
+      And An IfcElement
+      And The IfcElement is a part of another IfcElement through the relationship IfcRelAggregates
+      Then The relative placement of that IfcElement must be provided by an IfcLocalPlacement entity
+      And The PlacementRelTo entity must point to the IfcLocalPlacement of the container element established with IfcRelAggregates relationship
```
pjanck, line 11: What is a *PlacementRelTo **entity***? Did you mean **attribute**?

features/steps/steps.py
pjanck: I am unsure if there is any other relationship that would mean *being part of* / *has parts* in IFC. Is there another `{relationship}` that could even be a candidate for reuse of this rule?

features/steps/steps.py
```diff
@@ -221,3 +249,41 @@ def step_impl(context, related, relating, other_entity, condition):
...
     handle_errors(context, errors)
+
+@then('The relative placement of that {entity} must be provided by an {other_entity} entity')
+def step_impl(context, entity, other_entity):
+    if getattr(context, 'applicable', True):
+        errors = []
+        for obj in context.instances:
+            if not obj.ObjectPlacement.is_a(other_entity):
+                errors.append(instance_placement_error(obj, other_entity, "", "", "", ""))
+        handle_errors(context, errors)
```
pjanck, line 260: Formal parameter `{entity}` isn't used anywhere - consider removing it.

features/steps/steps.py
pjanck: Similar to my comment above, not every relationship means *containment*. I would even argue that `IfcRelAggregates` does not *contain*, but *aggregates*; there are other relationships for containment. Consider rewording.

pjanck (changes requested):
Initial comments and questions.

features/steps/steps.py
JTurbakiewicz: Fair point. Tried an approach with the lookup table, as it may prove to be the approach that can be reused the most.

I think it would be neat, to move the dicts I created to a separate python .py file, not to clutter steps.py. 
Would you be ok with it? Some time ago, I recall, there was an agreement to try to avoid it.

features/steps/steps.py
JTurbakiewicz: Done, similarly to the above, that is using the lookup table. 
As for the point `also not always an aggregate some relationships are 1..1 (see cardinality)`: I've added a try/except block that checks, if the returned value is an iterator. If not, it "packages it" into a tuple, so the for loop would still work.

features/OJP001_Relative-placement-for-elements-aggregated-to-another-element.feature
JTurbakiewicz: Right, thank you.

features/steps/steps.py
JTurbakiewicz: The intention was to have a little bit more flexibility/reusability in the .feature file, allowing the use of whatever word is required, without forcing the use of specifically `IfcElement`. 

I don't think it's a huge benefit, as the sentence is quite rigid, but perhaps it's worth to keep it?

features/steps/steps.py
aothms: Discussed during meeting:

We rephrase the statement to make relationship the subject of the sentence, that way we can use from and to as a generic way to establish the direction of the relationship.

```
> Given An IfcWall
>   And A relationship IfcRelAggregates (from IfcWall) to IfcBuildingElementPart
  
> Given An IfcBuildingElementPart
>   And A relationship IfcRelAggregates (to IfcBuildingElementPart) from IfcWall
     
     (from)               (to)
    Relating             Related
IfcWall <- IfcRelAggregates -> IfcBuildingElementPart
```

JTurbakiewicz: Implemented.

features/steps/steps.py
JTurbakiewicz: Agreed to use a .csv file. Implemented.

features/steps/steps.py
JTurbakiewicz: Added a check for it.

Ghesselink: In other branches Johan added a do_try function, maybe it'll be worth it to use in cases like these?
```
def do_try(fn, default=None):
    try: return fn()
    except: return default
```
In this particular case, it would be
`for obj in f.by_type("IfcElement"): do_try(lambda: obj.is_a('IfcStair'), False)`

But the check you have now seems to work fine as well :) But found it nice to share/perhaps in the future this can be reused

JTurbakiewicz: Just noticed that we use do_try in our function to get csv file paths, so why not reuse it here. 
Implemented, thank you for the good tip.

features/steps/steps.py
JTurbakiewicz: Changed. Additionally relationship attributes are derived from the csv file now, not hardcoded.

aothms (approved):

aothms:
Blergh more merge conflicts. But I think they were all just side by side inserts so I quickly resolved them. Maybe @JTurbakiewicz you can have a quick double check.

JTurbakiewicz:
Looks good I think. :)

PR #28, opened 2022-12-31
GRF001 identical coordinate operation for all representation contexts

Ghesselink:
Draft PR for GRF001

Still WIP, work to do:
 - Increase reuse of step implementation between GEM003-GRF001
 - Generate more error files
 - Test
 - Discuss theory and cases to include and exclude
 - README


features/GRF001_identical-coordinate-operations.feature
```diff
@@ -0,0 +1,11 @@
+@GRF
+Feature: GRF - Identical coordinate operations for all representation contexts
+
+  Scenario: IfcGeometricRepresentationContext
+
+      Given A file with Schema Identifier "IFC4X3" or "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4x1"
+        And An IfcGeometricRepresentationContext without subtypes
```
aothms, line 7: I see two discussion points regarding this:

- "An ... without subtypes" is a bit unclear. It looks like you're asking for instances that don't have subtypes (which is not really applicable). I guess the "Given An IfcX" as also already a bit unclear. Should we try something like:
  - Given the instances of type IfcGeometricRepresentationContext [without considering subtypes].
- There is also an ambiguity related to the then-clause. That we might want to resolve in the Given clause:

```
Given the instances of type IfcPolyline
  And the value for attribute Polygon
 Then all values must be unique
```

There are three possible interpretations here:
- All polylines considered there cannot be any shared points, ever
- Within a polyline duplicate points cannot exist
- No polyline should have the same sequence of points as another polyline.

So in python, considering we have `p1, p2, p3 = range(3); v = [(p1, p2), (p1, p3), (p1, p1)]`

- len(set(y for x in v for y in x)) == len([y for x in v for y in x])
- all(len(set(x)) == len(x) for x in v)
- len(set(v)) == len(v)

I think option 1 is not really interesting. So that leaves us to distinguish between 2 and 3. Which we both have:

- **Within a product** all shape reps should be unique
- **Within a model** all coordinate ops references by rep contexts must be identical.

So the question is do we make this distinction in the Given clause or in the Then clause. And maybe Given is easiest. In that case we can say sth like:

- Given **the** instances of type IfcGeometricRepresentationContext [without considering subtypes].
- Given **all** instances of type IfcGeometricRepresentationContext [without considering subtypes].

The only difference would be that in the first case you get

    values = [context_1, context_2]

In the second case you get

    values = [[context_1, context_2]]

The Then clause then remains unaltered. I didn't test this, hope it makes some sense. I didn't look at the python impl yet, but wanted to give this as input to the discussion on Monday.

Ghesselink: The sentence [without considering subtypes] is added https://github.com/buildingSMART/ifc-gherkin-rules/pull/28/commits

Initially, this was solved in the 'Then' clause. Solving it in the given clause seems to be clearer as well. 
> In the second case you get v = [[c1,c2]]

Not as easy as I thought because of the map_state function. It worked however when the situation (within model or within product) was saved in the context at the Given statement. In the Then clause, this can be converted to a list within a list in the 'within model' situation. The then clause remained almost the same then indeed. 

I've also altered the Then and the corresponding error message in such a way it can be reused with GEM003 (unique Product Shape RepresentationIdentifier). If we take the last attribute of the Given statement in GEM003 and the last rule, it seems possible to always generate a meaningful error message. It would also be possible to alter the error message in case the attributes are stacked more than once, so that it is clearer how the product and the last values relate (to avoid confusion like in https://github.com/buildingSMART/ifc-gherkin-rules/pull/21). 

Lastly, I compared entity type (IfcMapConversion in this case) with entity type, and not the internal coordinates. I'm not sure doing that would be the scope of this rule. I've read from Dion Moult: https://thinkmoult.com/ifc-coordinate-reference-systems-and-revit.html

> Although technically possible, it is bad practice and seasoned GIS professionals will tell you to never mix different CRSes on one map unless you want to be more artistic than useful

Issue #12, opened 2022-11-28
Rules in ALB001_Alignment-in-spacial-structure.feature might be premature


apinzenoehler:
Gherkin-feature file "ALB001_Alignment-in-spacial-structure.feature" contains the following rule:

      Scenario Outline: Agreement on each IfcAlignment being contained in an IfcSite

          Given A file with Schema Identifier "<schema_identifier>"
          And An IfcAlignment
          Then Each IfcAlignment must be contained in IfcSite

There is an ongoing discussion in bSI implementer forum which might come to a different conclusion.

BR Andreas

evandroAlfieri:
Good point, thanks Andreas. This rule was used as exercise to practice with Gherking and python implementation. We won't merge it before an agreement is reached. But I'll leave the branch open as this is still a good example of how to write rules for the Validation Service

PR #3, opened 2022-09-29
Feature: IfcSweptDiskSolidPolygonal Informal Proposition

aothms:
This PR adds support for the Informal Proposition that compares FilletRadius to the length of the segments of the directrix.

PR #29, opened 2023-01-01
GEM111- No duplicated points within a polyloop or polyline

JTurbakiewicz:


features/GEM111_No_duplicated_points_within_a_polyloop_or_polyline.feature
```diff
@@ -0,0 +1,23 @@
+@implementer-agreement
+@GEM
+Feature: GEM111 - No duplicated points within a polyloop or polyline
+
+  Scenario: Agreement on no duplicated points within a polyloop
+
+      Given A file with Schema Identifier "IFC2X3"
```
aothms, line 7: While this comes from the 2x3 impl agreements, I guess it's safe to assume it should apply files from all schema versions.

features/steps/steps.py
```diff
@@ -49,6 +49,22 @@ def fmt(x):
...
+    if entity.is_a('IfcPolyLoop'):
+        points = entity.Polygon
+    elif entity.is_a('IfcPolyline'):
+        points = entity.Points
+    return points
+
+
+def get_sending_system_precision(context):
+    try:
+        geo_representations = context.model.by_type('IFCGEOMETRICREPRESENTATIONCONTEXT')
```
aothms, line 62: There can be multiple rep contexts. I think we should take an effort in connecting the appropriate rep context to the rep item, which is unfortunately not trivial.

```
IfcRepresentation |-> IfcRepresentationContext
                  |-> IfcRepresentationItem (e.g IfcExtrudedAreaSolid) -> ... -> IfcPolyline
```

So you'd need to use ifcopenshell.file.get_inverse() (because there isn't a named inverse from polyline back to where it is referenced) to trace back the representation item chain to the representation and (there could be multiple) and then take the context with the largest precision.

features/GEM111_No_duplicated_points_within_a_polyloop_or_polyline.feature
```diff
@@ -0,0 +1,23 @@
...
+      And An IfcPolyLoop
+      Then The IfcPolyLoop must have no duplicate points
+
+  Scenario: Agreement on no duplicated points within a polyline
+
+      Given A file with Schema Identifier "IFC2X3"
+      And An IfcPolyLine
+      And IfcPolyLine first and last point are different
```
aothms, line 15: We have something similar here (still not merged) with "... forms an open/closed curve". https://github.com/buildingSMART/ifc-gherkin-rules/pull/3/files#diff-e883e365368e1e08c63684a4c4978f76f354dad388f6d71c41e45051063046efR8

I think we should try and unify this.

aothms, line 15: Also we should remove IfcPolyline from this step because later we'll probably make it so that it should also be applicable to IfcIndexedPolyCurve (an entity added in IFC4 which is similar in usage, just didn't exist yet at the time this impl agreement was written)

features/GEM111_No_duplicated_points_within_a_polyloop_or_polyline.feature
```diff
@@ -0,0 +1,23 @@
+@implementer-agreement
+@GEM
+Feature: GEM111 - No duplicated points within a polyloop or polyline
+
+  Scenario: Agreement on no duplicated points within a polyloop
+
+      Given A file with Schema Identifier "IFC2X3"
+      And An IfcPolyLoop
+      Then The IfcPolyLoop must have no duplicate points
```
aothms, line 9: I'm actually not so sure this is necessary, but we can leave it in. I would assume this is already covered by the Informal Propositions in the schema https://standards.buildingsmart.org/IFC/RELEASE/IFC2x3/TC1/HTML/ifctopologyresource/lexical/ifcclosedshell.htm (but I'm also not 100% sure).

features/steps/steps.py
```diff
@@ -283,4 +329,30 @@ def step_impl(context, representation_id):
...
+    handle_errors(context, errors)
+
+
+@then("The {entity} first and last point must be identical by reference")
+def step_impl(context, entity):
+    if getattr(context, 'applicable', True):
+        errors = []
+        for instance in context.instances:
+            points = get_entity_point_components(instance)
+            if str(points[0]) != str(points[-1]):
```
aothms, line 356: `points[0] == points[-1]` is ok for this. For instances of file it checks whether they are the same instance.

Same for is_a() with/without subtypes we should probably create section in the wiki on entity comparison.

- `points[0] == points[-1]` same instance
- `points[0].get_info(recursive=True, include_identifier=False) == points[-1].get_info(recursive=True, include_identifier=False)` same set of attribute values (recursively).

features/steps/steps.py
```diff
@@ -283,4 +329,30 @@ def step_impl(context, representation_id):
...
+def step_impl(context, entity):
+    if getattr(context, 'applicable', True):
+        errors = []
+        precision = get_sending_system_precision(context)
+        for instance in context.instances:
+            points = get_entity_point_components(instance)
+            points_coordinates = [point.Coordinates for point in points]
+            points_coordinates_rounded = []
+            for point_coordinates in points_coordinates:
+                point_coordinates_rounded = tuple(round(point / precision) * precision for point in point_coordinates)
```
aothms, line 343: This is clever because you can use sets, but it's not geometrically what is intended.

This is the unpleasant consequence of geometric tolerance. You loose the ability to certain things efficiently and loose desirable mathematical traits such as transitiveness (a = b, b = c ---> a = c).

I would propose to do a bruteforce approach using itertools.combinations(2) to take all 2-combinations from the list of points and then check whether the distance between those two points (2-norm, numpy.linalg.norm or just good old pythagoras) is within tolerance.

aothms (changes requested):

features/GEM111_No_duplicated_points_within_a_polyloop_or_polyline.feature
pjanck: This is already covered in the schema with `WR21`: https://standards.buildingsmart.org/IFC/RELEASE/IFC2x3/TC1/HTML/ifctopologyresource/lexical/ifcpolyloop.htm

aothms: That only seems to check whether there are no points with a different derived Dim attribute from the first point? 

Even if it would, in a formal express rule it would be really hard to factor in the tolerance.

pjanck: My bad, not the `WR21`, its the `UNIQUE` specifier :)

pjanck: I retract my comment, two distinct points could still have same coordinate values ...

aothms: Correct, I also had to double check to see what it means:

> each element in a list value of this data type shall be different from (i.e., not instance equal to)

instance equal later specified as

> the implementation dependent identifiers are the same

In **8.2.5 Value uniqueness on aggregates** they give two other approaches using the `value_unique()` function or using a custom comparison function.

features/GEM111_No_duplicated_points_within_a_polyloop_or_polyline.feature
JTurbakiewicz: Fair, removed this limitation.

features/GEM111_No_duplicated_points_within_a_polyloop_or_polyline.feature
JTurbakiewicz: Cool. I've made an attempt to unify this, took most of your code and tweaked it just a little bit, to make it fit my case.

JTurbakiewicz: If I remove `And An IfcPolyLine` I get some errors, as some unexpected instances are in the context. 

Would you be ok, if I simply added another scenario, like the one below? 

```
  Scenario: Agreement on no duplicated points within a polyline

      Given An IfcIndexedPolyCurve 
      Given IfcIndexedPolyCurve forms an open curve
      Then The IfcIndexedPolyCurve must have no duplicate points
```

If not, I will find a way around this.

JTurbakiewicz: Also, perhaps if we're including IfcIndexedPolyCurve (a polycurve?), would it be appropriate to change the lingo around this rule, e.g.  from `Feature: GEM111 - No duplicated points within a polyloop or polyline` to `Feature: GEM111 - No duplicated points within a polyloop, polyline or polycurve` ?

features/steps/steps.py
JTurbakiewicz: Done. :)

features/steps/steps.py
JTurbakiewicz: Done :)

features/GEM111_No_duplicated_points_within_a_polyloop_or_polyline.feature
JTurbakiewicz: To sum up, are we ok leaving it in?

features/GEM111_No_duplicated_points_within_a_polyloop_or_polyline.feature
JTurbakiewicz: Used a pronoun instead of a specified object, e.g.:
```

      Given An IfcPolyLine
      And It forms an open curve
      Then It must have no duplicate points
```

features/steps/steps.py
JTurbakiewicz: Added.

JTurbakiewicz: I've tested the new solution with a few files, generally works well, but I found some cases, in which an IfcRepresentationContext was not found. An example below:

`#352922=IfcPolyline((#352918,#352920))` -> 
`#352924=IfcGridAxis('11',#352922,.T.) `-> 
`#353156=IfcGrid('1H2s9IBD6pu931S6f7FDpD',#12,'Grid',$,$,#352917,#353153,(#352924,#352942,#352952,#352962,#352972,#353012,#353022,#353032,#353042,#353052,#353062,#353072,#353082,#353092,#353102,#353112,#353122,#353132),(#352982,#352992,#353002,#353142),$)` ->
`#7404=IfcRelContainedInSpatialStructure('04ldtj6cp2dME6CiP80Bzh',#12,$,$,(#7401,#7736`

So starting from `#352922=IfcPolyline((#352918,#352920))` it didn't quite reach an object with Context/Context reference. In the current implementation, a default precision is implemented in that case. Does that make sense or would it be reasonable to perhaps use some other method in addition to get_inverse()?

features/steps/steps.py
```diff
@@ -98,6 +149,19 @@ class representation_type_error:
...
+
+@dataclass
+class polyobject_structure_error:
+    inst: ifcopenshell.entity_instance
+    points: list
+    error_cause: str
+
+    def __str__(self):
+        if self.error_cause == 'duplicate_points':
+            return f"On instance {fmt(self.inst)} some of the points {self.points} are duplicate"
```
aothms, line 161: Here I think @Ghesselink put in more effort in similar error categories (duplicate representation ids I think) to really also identify to the user, which are the duplicates. In a long polyline with precision I think it can be helpful. It's not a very important point for me. Can also be done later if/when a user requests.

features/steps/steps.py
```diff
@@ -283,4 +358,27 @@ def step_impl(context, representation_id):
...
+@then("It must have no duplicate points")
+def step_impl(context):
+    if getattr(context, 'applicable', True):
+        errors = []
+        for instance in context.instances:
+            entity_contexts = get_entity_contexts(context, instance)
+            precision = get_precision_from_contexts(entity_contexts)
+            points_coordinates = get_points(instance)
+            for i in itertools.combinations(points_coordinates, 2):
+                if math.dist(i[0], i[1]) < precision:
```
aothms, line 371: "New in version 3.8." Thanks I didn't know this one :)

features/steps/steps.py
```diff
@@ -283,4 +358,27 @@ def step_impl(context, representation_id):
...
+                    errors.append(polyobject_structure_error(instance, points_coordinates, 'duplicate_points'))
```
aothms, line 372: If we don't specify which points are the duplicates I think we should `break` here to not report a ton of duplicates on the same pline.

features/steps/steps.py
```diff
@@ -98,6 +149,19 @@ class representation_type_error:
     def __str__(self):
         return f"On instance {fmt(self.inst)} the {self.representation_id} shape representation does not have {self.representation_type} as RepresentationType"
 
+
+@dataclass
+class polyobject_structure_error:
```
aothms, line 154: I'm a bit in doubt on whether we shouldn't simply use 2 different dataclasses. Similar comment I made to @Ghesselink earlier. I guess they should be as simple as possible and maybe later we assign an error code to them. But I don't have a very strong opinion on this.

aothms (changes requested):

aothms:
Great! It's a lot of minor comments, sorry..., but the one on the recursive nature of the inverses is especially important.

features/steps/steps.py
Ghesselink: FYI. It's in the GEM003/GRF001 subbranch in the 'Then' statement 'The values must be {identical_or_unique}'.

features/steps/steps.py
```diff
@@ -48,6 +50,55 @@ def fmt(x):
...
+    for rev_item in ifc_context.model.get_inverse(ent):
+        if hasattr(rev_item, 'ContextOfItems'):
+            set_instances.append(rev_item.ContextOfItems)
+    if len(set_instances) > 0:
+        return set_instances
+    else:
+        entity_context = get_entity_contexts(ifc_context, rev_item)
+        return entity_context
+
+def get_precision_from_contexts(entity_contexts, func_to_return='max', default_precision= 1e-05):
```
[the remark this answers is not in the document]
JTurbakiewicz: Of course, that's much better indeed. Done.

features/steps/steps.py
JTurbakiewicz: Done.

features/GEM111_No-duplicated-points-within-a-polyloop-or-polyline.feature
```diff
@@ -0,0 +1,22 @@
...
+
+  Scenario: Agreement on no duplicated points within a polyloop
+
+      Given An IfcPolyLoop
+      Then It must have no duplicate points
+
+  Scenario: Agreement on no duplicated points within a polyline
+
+      Given An IfcPolyLine
+      And It forms an open curve
```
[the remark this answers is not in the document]
JTurbakiewicz: Needed to build it up a little bit more, as it needed differentiation between open and closed curve in regard to what happens when first and last points are the same. Done.

features/steps/steps.py
```diff
@@ -225,6 +288,18 @@ def step_impl(context, field, values):
 
     context.applicable = getattr(context, 'applicable', True) and applicable
 
+@given('{attr} forms {closed_or_open} curve')
+def step_impl(context, attr, closed_or_open):
+    assert closed_or_open in ('a closed', 'an open')
+    should_be_closed = closed_or_open == 'a closed'
+    try:
```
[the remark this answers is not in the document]
JTurbakiewicz: Played with it for quite a while, doesn't quite work, although I loved the idea:

```
behave.step_registry.AmbiguousStep: @given('It forms {closed_or_open} curve') has already been defined in
existing step @given('{attr} forms {closed_or_open} curve') at features/steps/steps.py:289
```
I suppose we could perhaps use `behave.use_step_matcher("re")` and define `@given` using regex, which would allow the exclusion of `It ` in one of the steps. Something like below:

```
use_step_matcher("re")
@given(r"(It) forms (.+) curve") # looks for a string like: "It forms {an open} curve"
@given(r"(\b(?!It\b)\w+) forms (.+) curve") # looks for a string like: "Anything but 'It' forms {an open} curve"

```
I've tested it and it works, but in my opinion, it's not too readable. That's why I left the `@given` as was but exchanged the `try `for `if attr == 'It'`. What do you think?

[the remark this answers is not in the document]
JTurbakiewicz: Also reformated due to a different comment, I would say it's more readable now. 

Multiple decorators is an useful tool, could come in handy.

features/steps/steps.py
```diff
@@ -48,6 +50,55 @@ def fmt(x):
             return "...".join((v[:25], v[-7:]))
         return v
 
+def is_closed(seq_of_pt):
+    # @todo tolerance
```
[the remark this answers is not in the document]
JTurbakiewicz: Added.

features/steps/steps.py
JTurbakiewicz: Done.

features/steps/steps.py
```diff
@@ -48,6 +50,55 @@ def fmt(x):
...
+        raise NotImplementedError(f'get_points() not implemented on {inst.is_a}')
+
+def get_entity_contexts(ifc_context, ent):
+    if hasattr(ent, 'ContextOfItems'):
+        return ent.ContextOfItems
+    set_instances = []
+    if len(ifc_context.model.get_inverse(ent)) == 0:
+        return None # inverse unattainable, context not found
+    for rev_item in ifc_context.model.get_inverse(ent):
+        if hasattr(rev_item, 'ContextOfItems'):
```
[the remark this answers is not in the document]
JTurbakiewicz: Both implemented.

features/steps/steps.py
```diff
@@ -48,6 +50,55 @@ def fmt(x):
...
+    else:
+        entity_context = get_entity_contexts(ifc_context, rev_item)
+        return entity_context
+
+def get_precision_from_contexts(entity_contexts, func_to_return='max', default_precision= 1e-05):
+    precisions = []
+    if entity_contexts is None:
+        return default_precision
+    for entity_context in entity_contexts:
+        if hasattr(entity_context, 'Precision') and entity_context.Precision:
```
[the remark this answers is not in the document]
JTurbakiewicz: No problem, done.

features/steps/steps.py
JTurbakiewicz: I don't have a strong opinion either. Separated them here.

features/steps/steps.py
```diff
@@ -48,6 +50,55 @@ def fmt(x):
...
+    set_instances = []
+    if len(ifc_context.model.get_inverse(ent)) == 0:
+        return None # inverse unattainable, context not found
+    for rev_item in ifc_context.model.get_inverse(ent):
+        if hasattr(rev_item, 'ContextOfItems'):
+            set_instances.append(rev_item.ContextOfItems)
+    if len(set_instances) > 0:
+        return set_instances
+    else:
+        entity_context = get_entity_contexts(ifc_context, rev_item)
```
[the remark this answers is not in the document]
JTurbakiewicz: Changed. I will not close this and the previous conversation, in case you want to take a look at the changed solution. 

I will take a look tomorrow as well, just to be sure. :)

features/steps/steps.py
```diff
@@ -66,6 +68,51 @@ def fmt(x):
...
+            return inst.Points
+    elif inst.is_a('IfcPolyLoop'):
+        if return_type == 'coord':
+            return [p.Coordinates for p in inst.Polygon]
+        elif return_type == 'points':
+            return inst.Polygon
+    else:
+        raise NotImplementedError(f'get_points() not implemented on {inst.is_a}')
+
+def recurrently_get_entity_attr(ifc_context, ent, entity_to_look_for, attr_to_get, attr_found):
```
aothms, line 93: I think this looks good:

Just two small comments:

- I really prefer `inst` instead of `ent` as the variable name. To prevent confusion between entity names and instances.
- Good job on not doing `attr_found=set()` because default values are only initialized once so it would get all occurences to fill into the same set occurence. But I would prefer:

```python
def recurrently_get_entity_attr(..., attr_found=None):
  if attr_found is None:
    attr_found=set()
  ...
```

aothms (changes requested):

aothms:
Thanks, much appreciated ;)

aothms (approved):

PR #33, opened 2023-01-14
GEM004 - Constraints on RepresentationIdentifier

Ghesselink:
Creates an csv from the building smart documentation and checks if value for RepresentationIdentifier in the given IFC is valid (i.e. is mentioned in the documentation).


aothms (changes requested):
Sorry for my comments getting lost:

Summary:
- These are two tables for two attributes
- I'd really think it's better if we refer to a .csv filename in the gherkin statement so it's less magical
- I'd also keep everything out of the repo besides the .csv

Ghesselink:
Latest commit: 

- Only table for RepresentationIdentifer is used (not RepresentationType)
- .csv in gherkin statement
- update description
- Only keep .csv in repository

features/steps/steps.py
```diff
@@ -485,6 +511,44 @@ def step_impl(context, representation_id):
...
+
+        if constraint[-5:] == ".csv'":
+            csv_name = constraint.strip("'")
+            for i, values in enumerate(instances):
+                if not values:
+                    continue
+                attribute = getattr(context, 'attribute', None)
+
+                dirname = os.path.dirname(__file__)
+                filename = Path(dirname).parent / f"resources/{csv_name}"
```
aothms, line 542: This is a bit inconsistent. If you use Path() with the / operator then do it consistently:

```suggestion
                filename = Path(dirname).parent / "resources" / csv_name
```

aothms (changes requested):

features/steps/steps.py
Ghesselink: Ah, I did not see this was possible, thanks.

aothms (approved):

Issue #31, opened 2023-01-09
Wrap parameters in brackets?


aothms:
See #6.

Ghesselink:
Pyparsing has been applied and seems to work good https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/6a125d364887b9b539df9be48fd7d56175c54a5c 

Setting up proper pyparsing costs time (less so when applying it more frequent). However, agreeing and applying a convention for using brackets, double quotes, etc. in parameters also not effort-free. It is perhaps also clearer to leave the logic in the python implementation when possible.

Issue #25, opened 2022-12-21
Change "shall" to "must" as per #20


evandroAlfieri:
https://github.com/buildingSMART/ifc-gherkin-rules/blob/2c25656e9429e95ed2176305da722a003e482f0b/features/SPS001_Basic-spatial-structure-for-buildings.feature#L10-L19

evandroAlfieri:
Ref: #20

evandroAlfieri:
Avoiding opening a different issue, the same applies to
https://github.com/buildingSMART/ifc-gherkin-rules/blob/2c25656e9429e95ed2176305da722a003e482f0b/features/GEM001_closed-shell-edge-usage.feature#L9-L26

evandroAlfieri:
and this
https://github.com/buildingSMART/ifc-gherkin-rules/blob/2c25656e9429e95ed2176305da722a003e482f0b/features/GEM002_space-representation.feature#L7-L14

Issue #24, opened 2022-12-20
Resolving comments


pjanck:
@aothms I can't resolve my own comments on a PR - is there a setting I need to make?

evandroAlfieri:
For now, I would suggest the good practice where **the person opening the PR is responsible for resolving all conversations - before the PR is approved**.
cc: @aothms @Ghesselink @JTurbakiewicz @johltn 

If the suggestion in a comment is out of a pull request's scope, or if the reviewer is not happy with the resolution, or if the comments triggered a bigger discussion, in all these cases I suggest that anyone can open a new issue that tracks the feedback and links back to the original comment. See "[Opening an issue from a comment](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue#creating-an-issue-from-a-comment)"

pjanck:
Right, so be it.

Issue #11, opened 2022-11-28
Testfile "File001.ifc" in branch "alb001_alignment-in-spatial-structure" should pass but fails


apinzenoehler:
According to README.md in https://github.com/buildingSMART/ifc-gherkin-rules/tree/alb001_alignment-in-spatial-structure/test/ALB001 testfile should pass.

It fails in https://validate.buildingsmart.org with error report below:

    For instance:
        #26 = IFCALIGNMENT('0y2hQsQKfAqxEufY7Z_1jL', $, 'RS02 short VG', '12d Alignment String', 'natural clothoid', $, $, $, $);

    Wrong number of attributes on instance with id #26 expected 8 got 9
    For instance:
        #28 = IFCALIGNMENTHORIZONTAL('1FahSvf2z8YeREQCzieDuo', #2, 'Horizontal Alignment', '12d Horizontal Alignment', $, $, $, 0., (#29, #31, #33));

    Wrong number of attributes on instance with id #28 expected 7 got 9
    For instance:
        #35 = IFCALIGNMENTVERTICAL('3gA2mMqFX0xfZWiXfZxgI6', #2, 'Vertical Alignment', '12d Vertical Alignment', $, $, $, (#36, #37, #38, #39, #40));

    Wrong number of attributes on instance with id #35 expected 7 got 8

BR Andreas

aothms:
Every test file in the ifc-gherkin-rules assumes should have a valid syntax and schema. We have dedicated test files for the other validation steps. So the file needs to be corrected. So that also applies to #8, we'll likely still expand the readme with some additional content, but only IA IP (to be merged as simply "(conformance) rules") are relevant to this repo.

JTurbakiewicz:
Implemented in [03a855](https://github.com/buildingSMART/ifc-gherkin-rules/pull/13/commits/03a8551077efad773b8da42d51c5c9fedd0fdded).

The branch is still under development/review, but that error should be mitigated now.

PR #35, opened 2023-01-16
Cherry pickgrf001 identical coordinate operation for all representation contexts

Ghesselink:
Cleanup of GRF001


features/steps/steps.py
```diff
@@ -88,6 +88,17 @@ def __str__(self):
         else:
             return f"No instances of type {self.type_name} were encountered"
 
+@dataclass
+class representation_value_error:
+    inst: ifcopenshell.entity_instance
+    duplicate_value: str
+    duplicate_representations: ifcopenshell.entity_instance
```
aothms, line 95: type should be `list` or, if you want to annotate also the element, sth like `typing.Sequence[ifcopenshell.entity_instance]`

features/steps/steps.py
```diff
@@ -137,6 +148,23 @@ class representation_type_error:
     def __str__(self):
         return f"On instance {fmt(self.inst)} the {self.representation_id} shape representation does not have {self.representation_type} as RepresentationType"
 
+@dataclass 
+class value_identical_unique_error:
+    related: ifcopenshell.entity_instance
```
aothms, line 153: I don't understand here, which is `related` and which is `relating`?

features/steps/steps.py
```diff
@@ -211,14 +239,59 @@ def instance_getter(i,representation_id, representation_type, negative=False):
...
+
+def map_state(values, fn):
+    if isinstance(values, (tuple, list)):
+        return type(values)(map_state(v, fn) for v in values)
+    else:
+        return fn(values)
+
+@given("An {entity_opt_stmt}")
+def step_impl(context, entity_opt_stmt):
+    entity = entity_opt_stmt.split()[0]
```
aothms, line 266: I actually would really like the pyparsing approach here again. What do you think?

```
ifc_entity = Ifc\w+
grammar = ifc_entity + [ "with subtypes" | "without subtypes" ]
```

features/steps/steps.py
```diff
@@ -211,14 +239,59 @@ def instance_getter(i,representation_id, representation_type, negative=False):
...
+
+    try:
+        context.instances = context.model.by_type(entity, include_subtypes = include_subtypes(entity_opt_stmt))
+    except:
+        context.instances = []
+
+@given("{the_or_all} instances of {entity_opt_stmt}")
```
aothms, line 273: Wait, why do we have "Given An {entity_opt_stmt}" and "Given {the_or_all} instances of {entity_opt_stmt}". Couldn't these be using the same implementation?

features/steps/steps.py
```diff
@@ -211,14 +239,59 @@ def instance_getter(i,representation_id, representation_type, negative=False):
         if condition(i, representation_id, representation_type):
             return i
 
+def strip_split(stmt, strp = ' ', splt = ' '):
+    return list(
+        map(str.lower, map(lambda s: s.strip(strp), stmt.split(splt)))
```
aothms, line 244: It's easier to write `stmt.lower().split(splt)` then to use another map() for it after splitting

features/steps/steps.py
```diff
@@ -211,14 +239,59 @@ def instance_getter(i,representation_id, representation_type, negative=False):
         if condition(i, representation_id, representation_type):
             return i
 
+def strip_split(stmt, strp = ' ', splt = ' '):
```
aothms, line 242: The default values here don't make a lot of sense. If you split on `' '`, there wouldn't be any `' '` left to strip.

aothms: I think this approach is fine for now, but I think we later better revisit this using a grammar as well:

- It's more descriptive in terms of what values it can accept.
- It's stricter in making sure you don't supply wrong pattern
- You get much better error messages when it doesn't comply

features/steps/steps.py
aothms: ^ but save this for later :)

features/steps/steps.py
```diff
@@ -211,14 +239,59 @@ def instance_getter(i,representation_id, representation_type, negative=False):
...
+    return list(
+        map(str.lower, map(lambda s: s.strip(strp), stmt.split(splt)))
+    )
+
+def include_subtypes(stmt):
+    stmt = strip_split(stmt, strp = '[]')
+    if len(stmt) > 1 and 'subtypes' in stmt:
+        excluding_statements = ['without', 'not', 'excluding', 'no']
```
aothms, line 250: Too much permissiveness can also be confusing :) but again, let's revisit this when we turn it into a grammar.

features/steps/steps.py
```diff
@@ -211,14 +239,59 @@ def instance_getter(i,representation_id, representation_type, negative=False):
...
+        if len(set(stmt).intersection(set(excluding_statements))):
```
aothms, line 251: 1. len() is not necessary. empty set evaluates to false, same as 0 evaluates to false.
2. you can simply write: `return not set(stmt).intersection(set(excluding_statements))`. `if something: return False else return True` is considered a bit unecessary because you can in that case just return `not something`.
3. better write `excluding_statements = {'without', 'not', 'excluding', 'no'}` as a set literal so that you don't need to convert to set later on.

features/steps/steps.py
```diff
@@ -211,14 +239,59 @@ def instance_getter(i,representation_id, representation_type, negative=False):
...
+    entity = entity_opt_stmt.split()[0]
+
+    try:
+        context.instances = context.model.by_type(entity, include_subtypes = include_subtypes(entity_opt_stmt))
+    except:
+        context.instances = []
+
+@given("{the_or_all} instances of {entity_opt_stmt}")
+def step_impl(context, the_or_all, entity_opt_stmt):
+    the_or_all = the_or_all
```
aothms, line 275: This doesn't do much :)

I find `'all' in the_or_all.lower()` a bit too permissive again. If users by accident misspell "All" as "Ball" you don't want it to silently succeed.

features/steps/steps.py
```diff
@@ -211,14 +239,59 @@ def instance_getter(i,representation_id, representation_type, negative=False):
...
     except:
         context.instances = []
 
+    context.within_model = getattr(context, 'within_model', True) and within_model
+
+@given('Its attribute {attribute}')
+def step_impl(context, attribute):
+    context._push()
+    context.instances = map_state(context.instances, lambda i: getattr(i, attribute, None))
+    setattr(context, 'instances', context.instances)
```
aothms, line 291: As said earlier in another PR, don't use setattr in these cases where the attribute is a static string.

features/steps/steps.py
```diff
@@ -413,3 +485,64 @@ def step_impl(context, entity, other_entities):
...
+    if getattr(context, 'applicable', True):
+
+        stack_tree= list(filter(None, list(map(lambda layer: layer.get('instances'), context._stack))))
+
+        for i, values in enumerate([context.instances] if getattr(context, 'within_model', False) else context.instances):
+            if not values:
+                continue
+                        
+            values_str = list(map_many(values, empty_tuple_to_string, ifcopenshell_instance_type_to_string))
+            attribute = getattr(context, 'attribute', 'None')
```
aothms, line 526: I'd prefer `None` and not `"None"` so that we can meaningfully distinguish this case in the formatting.

features/steps/steps.py
```diff
@@ -413,3 +485,64 @@ def step_impl(context, entity, other_entities):
...
+def get_duplicate_values(values):
+    seen = set()
+    duplicates = [x for x in values if x in seen or seen.add(x)]
+    return duplicates
+
+def ifcopenshell_instance_type_to_string(v):
+    """
+        Converts ifcopenshell instance type to strings, if applicable
+    """
+    return do_try(lambda: v[0].is_a(), v)
```
aothms, line 498: For documentation purposes can you explain why we need to do `v[0]` here? I don't understand.

features/steps/steps.py
```diff
@@ -413,3 +485,64 @@ def step_impl(context, entity, other_entities):
...
+
+def empty_tuple_to_string(v):
+    """
+        Converts empty tuples type to strings, if applicable
+        To be used for meaningful error messages
+    """
+    return 'None' if isinstance(v, tuple) and not v else v
```
aothms, line 505: Isn't using `'None'` here risky since it's a value that could also exist in the model. In which case we would get false positives or negatives?

features/steps/steps.py
```diff
@@ -413,3 +485,64 @@ def step_impl(context, entity, other_entities):
...
+                        
+            values_str = list(map_many(values, empty_tuple_to_string, ifcopenshell_instance_type_to_string))
+            attribute = getattr(context, 'attribute', 'None')
+
+            duplicates = get_duplicate_values(values_str)
+
+            if (identical_or_unique == 'identical' and len(values_str) > 1 and not duplicates):
+                relating = context.instances
+                related = stack_tree[-1] 
+            elif (identical_or_unique == 'unique' and len(duplicates)):
```
aothms, line 533: `len(duplicates) -> duplicates` (like above)

features/steps/steps.py
```diff
@@ -413,3 +485,64 @@ def step_impl(context, entity, other_entities):
...
+        for i, values in enumerate([context.instances] if getattr(context, 'within_model', False) else context.instances):
+            if not values:
+                continue
+                        
+            values_str = list(map_many(values, empty_tuple_to_string, ifcopenshell_instance_type_to_string))
+            attribute = getattr(context, 'attribute', 'None')
+
+            duplicates = get_duplicate_values(values_str)
+
+            if (identical_or_unique == 'identical' and len(values_str) > 1 and not duplicates):
```
aothms, line 530: I don't think `len(values_str) > 1 and not duplicates` is correct. `['a', 'a', 'b']` implies `len(values_str) == 3`, but `duplicates == ['a']` so not empty. I think it can only be `len(set(values_str)) == 1`. Knowing only the duplicates and the number of original values is not sufficient.

features/steps/steps.py
aothms: I also don't really understand why we need to convert things to string.

aothms (changes requested):

features/steps/steps.py
Ghesselink: You're right. I've added an extra testfile with values ['IfcMapConversion', 'IfcRigidOperation', 'IfcMapConversion'] and this leads to a false positive. 
If I'm correct, len(set(values)) will fix this, i.e.
```
>>> aa = ['a', 'a']
>>> len(set(aa)) > 1
False # function will continue
>>> ab = ['a', 'b']
>>> len(set(ab)) > 1
True # generate error
>>> aab = ['a', 'a', 'b']
>>> len(set(aab)) > 1
True # generate error
>>> a = ['a']
>>> len(set(a)) > 1
False # continue function
```
Added in https://github.com/buildingSMART/ifc-gherkin-rules/commit/ab11bfef554cdb739d2beb505f2c0e99b80362bb

features/steps/steps.py
Ghesselink: Implemented.

features/steps/steps.py
Ghesselink: Deleted it in GEM003, but not here. Deleted in https://github.com/buildingSMART/ifc-gherkin-rules/commit/4dcfb4c509b06a23275b5c5af3e031dc0dd20677

What do you think is a good workflow for overlapping pull requests? Merging them every time after implementing fixes? It seems too easy to miss these things when doing it manually.

features/steps/steps.py
```diff
@@ -88,6 +88,17 @@ def __str__(self):
         else:
             return f"No instances of type {self.type_name} were encountered"
 
+@dataclass
+class representation_value_error:
```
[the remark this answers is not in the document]
Ghesselink: You're right, this one is from an old error class from GEM003 that is now merged with this branch. Deleted it https://github.com/buildingSMART/ifc-gherkin-rules/commit/3ec9423a83cd85c4696635da4d81306d53547600

features/steps/steps.py
Ghesselink: This class is deleted, but I think it is the same for the other class. Updated in https://github.com/buildingSMART/ifc-gherkin-rules/commit/44652325ccf0529efbee4f32a25e7b7f56650243
I took this approach from an already existing rule (SPS001): https://github.com/buildingSMART/ifc-gherkin-rules/blob/a714b1fe1d248d1e9088db3432d9d51487c01331/features/steps/steps.py#L80-L89 
To my understanding this should then also be changed, as well as type specifications dataclasses for other rules. But I think it would be better to do this in the non-rule sprint and not in this PR to avoid overlap.

features/steps/steps.py
```diff
@@ -137,6 +148,23 @@ class representation_type_error:
     def __str__(self):
         return f"On instance {fmt(self.inst)} the {self.representation_id} shape representation does not have {self.representation_type} as RepresentationType"
 
+@dataclass 
+class value_identical_unique_error:
+    related: ifcopenshell.entity_instance 
+    values: str
```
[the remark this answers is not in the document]
Ghesselink: See above. Changed in https://github.com/buildingSMART/ifc-gherkin-rules/commits/cherry-pickgrf001-identical-coordinate-operation-for-all-representation-contexts

features/steps/steps.py
Ghesselink: The idea was to always go from the top attribute to the bottom in the feature file, so to speak. So, in this case
`IfcGeometricRepresentationContext -> HasCoordinateOperation > IfcMapConversion`
Related in this case is IfcGeometricRepresentationContext (first), and relating IfcMapConversion.
It is consistent with ALB002, for example:
`IfcAlignment (top; related) > IfcAlignmentVertical (bottom; relating)`
 
With this rule, it is perhaps a bit confusing as 'relating' (e.g. #22IfcMapConversion...) is very closely linked to the value (IfcMapConversion). See also comment below

features/steps/steps.py
Ghesselink: Sounds good :) It is also nice work to do.

Linked it to the related issue so I can easily find it back later https://github.com/buildingSMART/ifc-gherkin-rules/issues/31

features/steps/steps.py
Ghesselink: Yes, I have tried that and I run into the same issue as before https://github.com/buildingSMART/ifc-gherkin-rules/issues/31 (also link to the issue so I can find it back)
It was unsure to whether to use pyparsing for this, use the brackets approach or accept a perhaps confusing distinction.

To be more precise, I thought it is easier to distinguish
`Given an IfcProduct
Given all instances of IfcGeometricRepresentations`
than
'Given the instances of IfcProduct
Given all instances of IfcGeometricRepresentations'

features/steps/steps.py
Ghesselink: Link to issue to find this back later #31

features/steps/steps.py
Ghesselink: That looks a lot cleaner. Implemented in https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/f002ab5cd3d2b969a63fcf05d16796dab2cee9fd

features/steps/steps.py
Ghesselink: Hehe, yes. 

See above, I was unsure what to do with it and left it as is for the time being. 
Link to issue #36

features/steps/steps.py
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/bedf3c266e7248420159dec3e15c4c2daf75c754

features/steps/steps.py
Ghesselink: For example, the values for all instances of IfcGeometricRepresentationContext within a model could be (for example)
`[(#24=IfcMapConversion(#11,#23,316131.64,5690966.11,$,1.,0.,$),), (#25=IfcRigidOperation(#22,#23,$,$,$),), (#27=IfcMapConversion(#26,#23,316131.64,5690966.11,$,1.,0.,$),)]`

features/steps/steps.py
Ghesselink: Converting things to string is now used to compare entity types with each other. For example:
```
>>> f.by_type("IfcMapConversion")
[#24=IfcMapConversion(#11,#23,316131.64,5690966.11,$,1.,0.,$), #27=IfcMapConversion(#26,#23,316131.64,5690966.11,$,1.,0.,$)]
>>> f.by_type("IfcMapConversion")[0] == f.by_type("IfcMapConversion")[1]
False
>>> f.by_type("IfcMapConversion")[0].is_a() == f.by_type("IfcMapConversion")[1].is_a() 
True #'IfcMapConversion" == "IfcMapConversion"
```

Or is there a better way to compare entities?

Also, sometimes IfcGeometricRepresentationContext gives back no value (or, to be more precise, an empty tuple). In that case, it would be more informative to return 'None' instead of '()'. But again, maybe this function is too specific for this rule and therefore should be modified to be more reusable. 

I could also leave the work for error messaging out and leave it for the non-rule sprint. For example, since relating (#23IfcMapConversion... and the value IfcMapConversion) are so closely related, I left the entity in which the value was found out of the error message. In GEM003 for example, stating at which entity value 'Axis' or 'Body' for RepresentationIdentifier was found is very informative, but in this rule less (I think ..). What do you think?

features/steps/steps.py
Ghesselink: WIP

features/steps/steps.py
Ghesselink: #36

features/steps/steps.py
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/pull/35

features/steps/steps.py
Ghesselink: I also realized we did something comparable with the project for Schöck, by using hashing. We can also compare instances that we and be as specific as we want. 

It also doesn't take into inheritance now I see now. I'll change it that it will take that into consideration, so it is in line with the approach we agreed earlier on (entity.is_a(i)). And take a note for a next meeting, I can imagine comparing entities will come back in the future, and I don't know if simple is good or simplistic..
```
>>> mapconv[0].is_a('IfcRigidOperation')
False
>>> mapconv[0].is_a('IfcMapConversion')
True
>>> mapconv[0].is_a('IfcCoordinateOperation')
True
```

features/steps/steps.py
Ghesselink: I realize that it is consistent with ALB002 indeed, but consistently wrong .. 

> In case of the 1-to-many relationship, the related side of the relationship shall be an aggregate being a set 1:N

It was good initially, but it got reversed with the Nests - IsNestedBy confusion. Will change it here and make a jira ticket to set it correct for the whole file.

Ghesselink: Reversed in https://github.com/buildingSMART/ifc-gherkin-rules/pull/35 and added a note at other lines were it needs to be reversed.

features/steps/steps.py
Ghesselink: It's now more general and readable, I think https://github.com/buildingSMART/ifc-gherkin-rules/commit/ca6d9d38e552cb690f6ccf3f32eb413ec577767b

features/steps/steps.py
Ghesselink: There is an option to consider inheritance now that we can reuse in future rules https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/7237bcaf90edf74e82548037b7ed2220903dbbc4

features/steps/steps.py
Ghesselink: The variable names are changed to more sensible names, I think. It should be clearer now. https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/2997a98dc1769acf00a07a37cf62d971b12aa7dd

features/steps/steps.py
aothms: Hold on a sec. When we say "The values must be identical" it's not just that they should both be *of type IfcMapConversion*, they should point to the exact same instance or have exact identical values. You can use `get_info(recursive=True, include_identifiers=False)` to get a native python dict that you can use for "deep" equality testing.

features/steps/steps.py
Ghesselink: In https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/e03228c7c4cf10dd8ad366dfdeebceab12a195df

Ghesselink:
For clearer information before merging, last changes after meeting last Thursday:

- The two given statements 'Given an entity' or 'Given all instances of X' (within the model) are combined.  https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/e03228c7c4cf10dd8ad366dfdeebceab12a195df
- As suggested, a  pyparsing appraoch is used to check whether or not to include subtypes. https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/6a125d364887b9b539df9be48fd7d56175c54a5c
- Instead of the 'magical' ;) .csv , information about with attributes to include is kept within the python implementation itself. In this case, by using the function `inst.get_info(recursive=True, include_identifier=False, ignore='SourceCRS')` https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/d9aea80d1932ef7cc820117bd142370a45ef4a74
- Resolve (potential) merge conflicts (e.g. https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/a4e621d953ce16ce9126de47b6c34af4730a6edc)

features/steps/steps.py
```diff
@@ -331,16 +406,25 @@ def step_impl(context, relationship_type, entity):
...
-
+ 
     context.applicable = getattr(context, 'applicable', True) and applicable
+    
+@given('Its values')
+def step_impl(context):
+    context._push()
+    instances_unpacked = unpack_sequence_of_entities(
+        context.instances)  # '(#23IfcWall..)' to '#23IfcWall'
+    values = [do_try(lambda : inst.get_info(recursive=True, include_identifier=False, ignore='SourceCRS'), None)
```
aothms, line 425: Does this work?

`ignore` is a sequence of attribute names. but you supply a string. Unfortunately this doesn't throw an exception in python because a string is a sequence of characters, but you'd be trying to ignore the attributes S, o, u, ... effectively.

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/entity_instance.py#L344

features/GRF001_identical-coordinate-operations.feature
```diff
@@ -0,0 +1,17 @@
...
+The rule verifies that the same coordinate system is used within an IFC model. 
+And not, for example, an IfcMapConversion in one representation context and IfcRigidOperation in another.
+"""
+
+  Scenario: IfcGeometricRepresentationContext
+
+      Given A file with Schema Identifier "IFC4X3" or "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4x1"
+        And All instances of IfcGeometricRepresentationContext without subtypes
+        And Its Attribute HasCoordinateOperation
+        And Its values
```
aothms, line 14: I had this in mind:

```suggestion
        And Its values excluding SourceCRS
```

features/steps/steps.py
```diff
@@ -331,16 +406,25 @@ def step_impl(context, relationship_type, entity):
...
         applicable = conditional_lowercase(get_mvd(context.model)) in values
     elif field == "Schema Identifier":
         applicable = context.model.schema.lower() in values
     else:
         raise NotImplementedError(f'A file with "{field}" is not implemented')
-
+ 
     context.applicable = getattr(context, 'applicable', True) and applicable
+    
+@given('Its values')
```
aothms, line 420: And then you can just say

```suggestion
@given('Its values')
@given('Its values excluding {excluding}')
def step_impl(context, excluding=()):
```

And just pass that to ignore directly.

features/steps/steps.py
```diff
@@ -331,16 +406,25 @@ def step_impl(context, relationship_type, entity):
...
+def step_impl(context):
+    context._push()
+    instances_unpacked = unpack_sequence_of_entities(
```
aothms, line 423: Instead of this `unpack_sequence_of_entities`, I'd actually expect we'd be using this `map_state()` function. Did you try? What were the issues?

aothms (changes requested):

features/steps/steps.py
Ghesselink: It should not work not indeed, but it does: 
```
>>> f.by_type("IfcRigidOperation")[0].get_info()['SourceCRS']
#11=IfcGeometricRepresentationContext($,'Model',3,1.E-05,#9,#10)
>>> f.by_type("IfcRigidOperation")[0].get_info(ignore='SourceCRS')['SourceCRS']
KeyError: 'SourceCRS'
```
I'll change it to be in a list regardless to be on the safe side. 

Could it be that it has something to do with the commented out lines (in ifcopenshell -> entity_instance.get_info())
```
def get_info_(inst):
                            # for ty in ignore:
                            #     if inst.is_a(ty):
                            #         return None
```
The only lines where ignore seems to be used is: 
```
if self.wrapped_data.get_attribute_names()[i] in ignore:
                        continue
```
 -> It will pass, even with only a string
```
>>> 'SourceCRS' in 'SourceCRS'
True
```

features/steps/steps.py
Ghesselink: Yes, I had a bit of trouble figuring it out. For example take the following two instances, both packed in a tuple
`
[(#22=IfcRigidOperation(#11,#21,IfcLengthMeasure(35010.),IfcLengthMeasure(1560.),$),), (#24=IfcRigidOperation(#23,#21,IfcLengthMeasure(23684.),IfcLengthMeasure(1564.),$),)]`

With the desired output: 
`
[#22=IfcRigidOperation(#11,#21,IfcLengthMeasure(35010.),IfcLengthMeasure(1560.),$), #24=IfcRigidOperation(#23,#21,IfcLengthMeasure(23684.),IfcLengthMeasure(1564.),$)]`


I tried it with
`instances_unpacked =list(map_state(context.instances, lambda i: i)) # or even without list(...)`
which returns (the same as context.instances)
`[(#22=IfcRigidOperation(#11,#21,IfcLengthMeasure(35010.),IfcLengthMeasure(1560.),$),), (#24=IfcRigidOperation(#23,#21,IfcLengthMeasure(23684.),IfcLengthMeasure(1564.),$),)]`

However, when trying
`instances_unpacked = list(map_state(context.instances, lambda i: i[0]))`
it returns
`[(#11=IfcGeometricRepresentationContext($,'Model',3,1.E-05,#9,#10),), (#23=IfcGeometricRepresentationContext($,'Model',3,1.E-05,#9,#10),)]`
Which suggest that it gets to the ifcopenshell.entity_instance, of which i[0] refers to the first item in that instance
I tried some more things, but I could not seem to unpack it proper this way. And with the unpack_sequence_of_entities function it worked. Do I miss something?

aothms: I'm not convinced there is an issue with that.

Yes, `HasCoordinateOperation` is a `set[0:1]` so in the step1 below is where the additional "list-wrapping" is introduced.

```
[IfcGeometricRepresentationContext_0, IfcGeometricRepresentationContext_1, IfcGeometricRepresentationContext_2]
[[IfcCoordinateOperation_0], [IfcCoordinateOperation_1], []]
[[{TargetCRS:IfcCoordinateReferenceSystem_0, Eastings: 1, ...}], [{TargetCRS:IfcCoordinateReferenceSystem_1, Eastings:2 , ...}]]
```

But the constraint on identical values is still the same according to the python semantics. Trivial example:

```
>>> 1 == 1
True
>>> [1] == [1]
True
```

So even if the values are wrapped in a list of length one (in the bigger list of all instance value), we'd still expect the same result:

```
>>> set([(1,),(1,)])
{(1,)}
>>> set([(1,),(2,)])
{(1,), (2,)}
```

But in the case above I had to use tuples, because lists are mutable in python and therefore not hashable and therefore not able to be converted to set.

```
>>> set([[1],[2]])
TypeError: unhashable type: 'list'
```

But you can still do:

```
>>> vs = [[1], [1], [1]]
>>> all(v == vs[0] for v in vs[1:])
True
```

> With the desired output:

So this I'm basically trying to question. The values wrapped in a tuple or list doesn't change in this case the ability to test them to be identical.

Ghesselink: I understand, thanks for the explanation. 
Even is the output is not desired in this case, I still don't fully understand why map_state returns a tuple. 
But I'll wrap my head around that outside this PR. 

```
The unpack_sequence_of_entities function is replaced by map_state. 
    context.instances = map_state(context.instances, lambda inst: do_try(
        lambda: inst.get_info(recursive=True, include_identifier=False, ignore=excluding), None))
# as you suggested, the output is [(A),(B)] ==  [(A),(B)]  == True 
```
https://github.com/buildingSMART/ifc-gherkin-rules/pull/35/commits/498a5ca9b9acc0190784ba7a76d423e54dc30de8

aothms (approved):

aothms (approved):

PR #37, opened 2023-01-22
Alb999 rail alignment value constraints

aothms:
# Rule implementation

> An IfcAlignmentHorizontal shall only nest a list of IfcAlignmentHorizontalSegment (through IfcAlignmentSegment) that has PredefinedType set to LINE, CIRCULARARC or CLOTHOID.

NB The validation service validates requirements that are universally true and not organization or context-specific. For that purpose:

- This rule still needs to be updated to be universally true, with
  - context-specific requirements encoded in IDS (this is possible because the non-rooted entity IfcAlignmentHorizontal can be selected with an attribute constraint on PredefinedType)
  - the rule remainig universal but individual stakeholders "forking" the repository to make organization-specific adaptations
- (or) the rule will remain just an example and will not be merged to the main branch (but can still be invoked through the "sandbox" functionality) 

Considering the following example instantiation diagram

alig dot

the feature has been annotated with the context (name of instance) being operated on in square brackets (`[  ...  ]`)
  
```
@implementer-agreement
@ALB
Feature: Rail Alignment value constraints

  Background: Rail alignment
  
[AL1, AL2]           Given An IfcAlignment
[AL1]                And A relationship IfcRelReferencedInSpatialStructure to IfcAlignment from IfcRail
    
  Scenario: Agreement on allowed values for horizontal segment types
    
[AL1_H]              Given A relationship IfcRelNests from IfcAlignment to IfcAlignmentHorizontal and following that
[AL1_H_S1, AL1_H_S2] Given A relationship IfcRelNests from IfcAlignmentHorizontal to IfcAlignmentSegment and following that
[..., ...]           Given Its attribute DesignParameters
[LINE, HELMERTCURVE] Given Its attribute PredefinedType
    
[raises error        Then the value must be "LINE" or "CIRCULARARC" or "CLOTHOID"
on HELMERTCURVE]
```

The current validation report for a population model like this is according to:

```
Rail Alignment value constraints/Agreement on allowed values for horizontal segment types.v1
• the value must be "LINE" or "CIRCULARARC" or "CLOTHOID"
The value 'HELMERTCURVE' on #40=IfcAlignmentHorizontalSegment('17KlfolOr6ReLWMUjUKk$j',$,$,$,$,$,$,$,.HELMERTCURVE.) is not one of 'LINE', 'CIRCULARARC', 'CLOTHOID'
```

In principal the entire history of "stack frames" is available when building the report so we can recover the full path from alignment to segment type.

### Background: Rail alignment

This rule uses the background statement, which is functionally equivalent to having the same `given` statements prepended to each of the scenarios, but offers additional advantages later in categorizing such rules based on their background. We might be referring to this background selection as an "activation mechanism", i.e a graph pattern (in this case) that triggers validation constraints to become applicable.

### to/from

```
A relationship IfcRelReferencedInSpatialStructure to IfcAlignment from IfcRail
                                                  ^^
```

`to` refers to the Relating(Object/...) side of the relationship node

`to` and `from` can be positioned arbitrarily but the convention is that the first end of the relationship should correspond to the previously established context.
                                                 
### corresponding nodes to build a graph

```
Given An IfcAlignment
         ^^^^^^^^^^^^
  And A relationship IfcRelReferencedInSpatialStructure to IfcAlignment from IfcRail
                                                           ^^^^^^^^^^^^
```

The first part of the relationship definition should correspond to the previously established context

### traversal

```
Given An IfcAlignment
         ^^^^^^^^^^^^
  And A relationship IfcRelReferencedInSpatialStructure to IfcAlignment from IfcRail
                                                           ^^^^^^^^^^^^
...

Given A relationship IfcRelNests from IfcAlignment to IfcAlignmentHorizontal and following that
                                      ^^^^^^^^^^^^

```

In the first case the context stays on alignment, so it gets repeated in the third statement.

```
Given A relationship IfcRelNests from IfcAlignment to IfcAlignmentHorizontal and following that
                                                                             ^^^^^^^^^^^^^^^^^^
Given A relationship IfcRelNests from IfcAlignmentHorizontal to IfcAlignmentSegment and following that
```

However with the `and following that` clause we cause the context to switch to `IfcAlignmentHorizontal` (in this case).


aothms:
Sandbox ok ca8bf07eb1868efc9dd68eb4a9c50cb234542212

aothms:
Sandbox ok c3b4a2c9bb1ee41c57827ed7225315c922ef7fc1

aothms:
Sandbox ok e2d5cfa1a3ec668925ecba5880a781b34d3e2a22

features/ALB999_rail_alignment_value_constraints.feature
```diff
@@ -0,0 +1,17 @@
...
+  
+    Given An IfcAlignment
+      And A relationship IfcRelReferencedInSpatialStructure to IfcAlignment from IfcRail
+    
+  Scenario: Agreement on allowed values for horizontal segment types
+    
+    Given A relationship IfcRelNests from IfcAlignment to IfcAlignmentHorizontal and following that
+    Given A relationship IfcRelNests from IfcAlignmentHorizontal to IfcAlignmentSegment and following that
+    Given Its attribute DesignParameters
+    Given Its attribute PredefinedType
```
pjanck, line 15: Did you meant `And` in the last three lines? Or are all four conditions to be interpreted independently of one another?

features/OJP001_Relative-placement-for-elements-aggregated-to-another-element.feature
```diff
@@ -0,0 +1,11 @@
...
+@OJP
+Feature: OJP001 - Relative placement for elements aggregated to another element
+
+  Scenario: Agreement on the relative placement of IfcElements being a part of another IfcElement through the relationship IfcRelAggregates
+
+      Given A file with Schema Identifier "IFC2X3"
+      And An IfcElement
+      And A relationship IfcRelAggregates from IfcElement to IfcElement
+      Then The relative placement of that IfcElement must be provided by an IfcLocalPlacement entity
+      And The PlacementRelTo attribute must point to the IfcLocalPlacement of the container element established with IfcRelAggregates relationship
```
pjanck, line 11: What if the container is placed by `IfcGridPlacement`? As the sentence reads right now, it is not allowed for `IfcRelAggregates.RelatingObject` to ever be placed by anything else as `IfcLocalPlacement`.

features/ALB999_rail_alignment_value_constraints.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  Background: Rail alignment
+  
+    Given An IfcAlignment
+      And A relationship IfcRelReferencedInSpatialStructure to IfcAlignment from IfcRail
+    
+  Scenario: Agreement on allowed values for horizontal segment types
+    
+    Given A relationship IfcRelNests from IfcAlignment to IfcAlignmentHorizontal and following that
+    Given A relationship IfcRelNests from IfcAlignmentHorizontal to IfcAlignmentSegment and following that
```
pjanck, line 13: What is `and following that`?

features/OJP001_Relative-placement-for-elements-aggregated-to-another-element.feature
```diff
@@ -0,0 +1,11 @@
+@implementer-agreement
+@OJP
+Feature: OJP001 - Relative placement for elements aggregated to another element
+
+  Scenario: Agreement on the relative placement of IfcElements being a part of another IfcElement through the relationship IfcRelAggregates
+
+      Given A file with Schema Identifier "IFC2X3"
+      And An IfcElement
+      And A relationship IfcRelAggregates from IfcElement to IfcElement
+      Then The relative placement of that IfcElement must be provided by an IfcLocalPlacement entity
```
pjanck, line 10: Which `that` does this sentence refer to? In the sentence above, there are two `IfcElement`s ...

pjanck (changes requested):
Initial comments for Gherkin.

features/ALB999_rail_alignment_value_constraints.feature
aothms: There is no functional difference between that. `and` is just a shortcut for the previously used gherkin keyword. But I agree it should be used consistently (maybe it's actually clearer to never use `and`).

features/ALB999_rail_alignment_value_constraints.feature
aothms: Good question, I've updated the PR description. You can find the answer under `traversal`. In short it means that, in the first case without `and following that` the `a relationship ...` statement checks whether such a relationship exists, but keeps the context on the previously active node (alignment). In the second case we actually descend according to the nesting structure, so that the attribute DesignParameters is read from the IfcAlignmentSegment.

features/OJP001_Relative-placement-for-elements-aggregated-to-another-element.feature
aothms: I'm not going to reply here, this feature shows up here, because my PR was based on two un-merged PRs. See #27.

aothms:
Sandbox ok 08ae8d6ba372801573d28edb45ddd2631d8edebf

aothms:
Sandbox ok 45690e3608a06451e39261669131b6ef5c6d6831

aothms:
Sandbox ok 427fbf0ceeadbc760656dc2b857bbba7ec5368fd

features/environment.py
```diff
@@ -3,5 +3,13 @@
 from behave.model import Scenario
 
 def before_feature(context, feature):
+    # @tfk we have this strange issue between stack frames blending over
+    # between features so we need to preserve only the bottom two stack
+    # frames when beginning a new feature.
+    context._stack = context._stack[-2:]
```
Ghesselink, line 9: Aah, that explains it .. I tried to add context._push() with at the 'Its values' statement to retrieve it later and run into exactly this (another feature somehow not working anymore ..).

civilx64:
Cleaning up stale draft PRs.

PR #39, opened 2023-01-29
BS000 - Run tests only within feature

JTurbakiewicz:


test/test_main.py
```diff
@@ -9,29 +9,33 @@
...
 test_files = glob.glob(os.path.join(os.path.dirname(__file__), "files/**/*.ifc"), recursive=True)
+
+
 @pytest.mark.parametrize("filename", test_files)
 def test_invocation(filename):
     results = list(run(filename))
     base = os.path.basename(filename)
-    
+    rule_code = base.split('-')[1].strip()
+    results_filtered_by_rule_code = [result for result in results if rule_code.lower() in result[0].lower()]
```
aothms, line 22: Instead of filtering posthoc, actually the `behave` command line usage also allows to specify as a positional argument a feature directly by filename. Maybe it would be cleaner to add an optional feature_name to `run()`?

https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/main.py#L50

aothms (changes requested):

test/test_main.py
JTurbakiewicz: Probably makes sense, was a little bit afraid to change the run() function itself. 

I am not quite sure what to do with run() invocation in the __main__.py file: https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/__main__.py#L20

At the moment left it as was.

aothms (approved):

PR #41, opened 2023-02-01
Update main.py

aothms:


JTurbakiewicz:
Hi, 

Sorry, I was thinking that only our rules will be checked, so didn't really think about the `split()[1]` crashing. 
To be 100% on the safe side, added a commit with exception logging. Probably useless, at least I hope so, but won't harm.

JTurbakiewicz (approved):

PR #42, opened 2023-02-02
small fixes/documentations improvements

Ghesselink:
A couple of small improvements that were noted. They are not specific to a rule and, on their own, would be too small for a separate PR. Most have mainly to do with readability/documentation

The following is added:

- On a couple of occasions,`if len(set()): do X ` is changed to simply `if set(): do X`. In these cases, we are interested in the existence of a certain set or list and not the length of it. 
- Type annotations for better documentation. For example, in the case of points, it is clearer that these are ifcopenshell.entity_instances and not a list with integers. 
- In ALB002: Reverse related & relating. They are now in the wrong order in the steps.py file. I think it got reversed during the confusion of the alb nesting Nest/IsNestedBy.


Ghesselink:
Updated with restructure of Jakub in https://github.com/buildingSMART/ifc-gherkin-rules/pull/61

PR #47, opened 2023-02-06
GEM004 merge errors fix

Ghesselink:
There was a lot of merging back and forth and some things did not pass through correctly. 

 -> The testfiles of GEM004 were not merged into main -> Deleted in https://github.com/buildingSMART/ifc-gherkin-rules/commit/275f1bc4fa22f967f9bf61036b99694809e4ecb4
 -> In the 'then' implementation, the constraint did not get parsed correctly


features/steps/steps.py
```diff
@@ -684,8 +684,10 @@ def step_impl(context, constraint, num=None):
 
     within_model = getattr(context, 'within_model', False)
 
-    if constraint.startswith('be ') or constraint.startswith('in '):
-        constraint = constraint[3:]
+    #to account for order-dependency of removing characters from constraint
+    for startswith, length in itertools.chain.from_iterable(itertools.permutations([('not ', 4), ('be ', 3), ('in ', 3)])):
```
aothms, line 688: Why do we have 'not' in here suddenly? That would completely change the meaning of the statement but it would simply be silently discarded here from the constraint.

aothms: I don't see a reason to manually provide `length` why not simply compute it from `startswith` (which I would call `prefix`).

aothms: Did you check what from_iterable(permutation(...)) actually obtains:

```
>>> list(itertools.chain.from_iterable(itertools.permutations([('not ', 4), ('be ', 3), ('in ', 3)])))
[('not ', 4), ('be ', 3), ('in ', 3), ('not ', 4), ('in ', 3), ('be ', 3), ('be ', 3), ('not ', 4), ('in ', 3), ('be ', 3), ('in ', 3), ('not ', 4), ('in ', 3), ('not ', 4), ('be ', 3), ('in ', 3), ('be ', 3), ('not ', 4)]
```

That would be one hello of a sentence.

Why not rather use a while loop:

```suggestion
    while constraint.startswith('be ') or constraint.startswith('in '):
        constraint = constraint[3:]
```

aothms (changes requested):

aothms:
I just observed this as well. That was an unlucky combination of absence of test files and incorrect behaviour after merging. Please proceed with these suggestions as soon as feasible and we hotfix it in production.

features/steps/steps.py
Ghesselink: > Why do we have 'not' in here suddenly? That would completely change the meaning of the statement but it would simply be silently discarded here from the constraint.
In another PR (GEM005) there is the sentence 'The value must **not** be body' . I see that discard it from the constraint like that can cause undesired side effects, I'll change it there as well so it is explicit/less silent.

Ghesselink: > That would be one hello of a sentence.

It's a long sentence, but the computation time is very low so I thought that would not be a problem. 

> Why not rather use a while loop

Because it is sometimes easier to complicate things than to find simple solutions. Thanks :)

aothms (approved):

PR #50, opened 2023-02-09
VAL142 - split steps.py file

JTurbakiewicz:


aothms:
@JTurbakiewicz Could you resolve the conflicts? 

Editors tend to get hopelessly lost in these kind of conflicts, so a trick I always use:

    git diff `git merge-base main val142-split-steps-file`..main -- features/steps/steps.py

This will print the changes on the main branch to steps.py (you need to use a bash based shell for this such as git bash) that will allow you to manually apply to changes then onto the feature branch.

aothms (approved):

aothms:
Looks great, the ast analysis script is very elegant

PR #51, opened 2023-02-13
[V - Non-Rule] Rule-specific testing

Ghesselink:
There are many files at the moment (78 in a newly developed branch) and there is no good integration (as far as I know) between behave in the command line and gherkin. 
This functionality allows to test for specific rules (e.g. 'alb001, 'alb002') as well as categories (e.g. 'alb') just by running 'pytest -sv'

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -15,14 +15,33 @@ except ImportError:
 rule_code_pattern = re.compile(r"^[a-zA-Z]{3}\d{3}$")
 rule_codes = list(filter(lambda arg: rule_code_pattern.match(arg), sys.argv[1:]))
 
-test_files = []
-for code in rule_codes:
-    paths = glob.glob(os.path.join(os.path.dirname(__file__), "files/", code.lower(), "*.ifc"))
-    if not paths:
-        print(f"No IFC files were found for the following rule code: {code}. Please provide test files or verify the input.")
-    test_files.extend(paths)
+def get_test_files():
+    """
+    Option -> Example
+    Test files for a rule -> 'python3 test_main.py alb001'
+    Also applies for multiple rules -> 'python3 test_main.py alb001 alb002'
+    Test files for a single file -> 'python3 test_main.py <path>.ifc
+    Also applies for multiple files -> 'python3 test_main.py <path1>.ifc <path2>.ifc'
+    Codes and rules can also be combined -> 'python3 test_main.py alb001 <path>.ifc'
+    """
+    rule_code_pattern = re.compile(r"^[a-zA-Z]{3}\d{3}$")
+    rule_codes = list(filter(lambda arg: rule_code_pattern.match(arg), sys.argv[1:]))
 
-@pytest.mark.parametrize("filename", test_files)
+    test_files = []
+    for code in rule_codes:
+        paths = glob.glob(os.path.join(os.path.dirname(__file__), "files/", code.lower(), "*.ifc"))
+        if not paths:
+            print(f"No IFC files were found for the following rule code: {code}. Please provide test files or verify the input.")
+        test_files.extend(paths)
+
+    file_pattern =  r".*\.ifc(\')?$" #matches ".ifc" and "ifc'"
+    test_files.extend([s.strip("'") for s in sys.argv if re.match(file_pattern, s)])
+
+    if not test_files: # for example, with 'pytest -sv'
+        test_files = glob.glob(os.path.join(os.path.dirname(__file__), "files/**/*.ifc"), recursive=True)
+    return test_files
+
+@pytest.mark.parametrize("filename", get_test_files())
 def test_invocation(filename):
     results = list(run(filename))
     base = os.path.basename(filename)
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -11,7 +11,15 @@ except ImportError:
     sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
     from main import run
 
+
 test_files = glob.glob(os.path.join(os.path.dirname(__file__), "files/**/*.ifc"), recursive=True)
+
+# test only for specific rule(s)
+DEVELOPMENT = os.environ.get('environment', 'production').lower() == 'development'
+if DEVELOPMENT:
+    test_rules = ['alb001', 'alb002'] # can also be modified to, for example, 'alb' if all the rules of a category should be tested
+    test_files = [test_file for test_file in test_files if any(x in test_file.lower() for x in test_rules)]
+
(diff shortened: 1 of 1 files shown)


aothms:
Instead of hardcoding this I'd rather look for a way to supply this as a command line argument (`python test_main.py alb`). See for inspiration: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/test/test_rules.py#L15 Also see Jakub's earlier PR https://github.com/buildingSMART/ifc-gherkin-rules/pull/39

Ghesselink:
It's now possible with a command line argument (e.g. `python test_main.py alb`, but also multple arguments as in `python test_main.py alb002 alb003`)

```
test_for_rules = any(re.match(r"^[a-zA-Z]{3}\d{3}$", arg) for arg in sys.argv[1:])
test_files = [fn for fn in glob.glob(os.path.join(os.path.dirname(__file__), 
                "files/**/*.ifc"), recursive=True) if len(sys.argv) < 2 or any(x in fn.lower() for x in sys.argv[1:]) or not test_for_rules]
```

test/test_main.py
```diff
@@ -11,7 +12,10 @@
     sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
     from main import run
 
-test_files = glob.glob(os.path.join(os.path.dirname(__file__), "files/**/*.ifc"), recursive=True)
+test_for_rules = any(re.match(r"^[a-zA-Z]{3}\d{3}$", arg) for arg in sys.argv[1:])
+test_files = [fn for fn in glob.glob(os.path.join(os.path.dirname(__file__),
```
aothms, line 16: I still find it a bit confusing:

- `test_for_rules`: why is this a boolean, why not do `filter()` instead of `any()` to filter out all non-rule code argv and use them below
- the regex is lower or upper case but by doing `x in fn.lower()` actually an uppercase argument will not work
- some of the rule codes such as `alb` will arbitrarily match word fragments such as `albino` which will be rather confusing. Test files are in separate directories so why not filter only on that directory name instead of entire file path?

aothms (changes requested):

test/test_main.py
Ghesselink: It's modified in https://github.com/buildingSMART/ifc-gherkin-rules/pull/51/commits/377f4e65b7fc0fb5f023b74dc4e07c37180a327f

> why is this a boolean, why not do filter() instead of any() to filter out all non-rule code argv and use them below

Done.

> some of the rule codes such as alb will arbitrarily match word fragments such as albino which will be rather confusing. Test files are in separate directories so why not filter only on that directory name instead of entire file path?

Done. 

> the regex is lower or upper case but by doing x in fn.lower() actually an uppercase argument will not work

The aim was to also allow for uppercase input, such as `'Alb001'` or` 'ALB001' `instead of `'alb001'` It should be good now: the regex match will still allow for this uppercase and it later converts into lowercases to match the folder name (which are always in lowercases, such as in 'alb001'). 

Furthermore, there is a warning added in the case there is someone wants to test a non-existing rule code (e.g. `'alb005'`). This could be either a typo or it could mean that there are no test files added yet (or something else even). Nothing too spectecular, but it might help someone writing rules and testing it locally.

test/test_main.py
```diff
@@ -11,8 +12,36 @@
     sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
     from main import run
 
-test_files = glob.glob(os.path.join(os.path.dirname(__file__), "files/**/*.ifc"), recursive=True)
-@pytest.mark.parametrize("filename", test_files)
+rule_code_pattern = re.compile(r"^[a-zA-Z]{3}\d{3}$")
```
aothms, line 15: You have `rule_code_pattern` and `rule_codes` duplicated in the function body on line 27.

test/test_main.py
```diff
@@ -11,8 +12,36 @@
...
+    """
+    Option -> Example
+    Test files for a rule -> 'python3 test_main.py alb001'
+    Also applies for multiple rules -> 'python3 test_main.py alb001 alb002'
+    Test files for a single file -> 'python3 test_main.py <path>.ifc
+    Also applies for multiple files -> 'python3 test_main.py <path1>.ifc <path2>.ifc'
+    Codes and rules can also be combined -> 'python3 test_main.py alb001 <path>.ifc'
+    """
+    rule_code_pattern = re.compile(r"^[a-zA-Z]{3}\d{3}$")
+    rule_codes = list(filter(lambda arg: rule_code_pattern.match(arg), sys.argv[1:]))
```
aothms, line 28: Instead of blindly taking sys.argv I would filter out all the "flags" e.g `-sv` 

```suggestion
    rule_codes = list(filter(lambda arg: rule_code_pattern.match(arg), [a for a in sys.argv[1:] if not a.startswith('-')]))
```

test/test_main.py
```diff
@@ -11,8 +12,36 @@
...
+    for code in rule_codes:
+        paths = glob.glob(os.path.join(os.path.dirname(__file__), "files/", code.lower(), "*.ifc"))
+        if not paths:
+            print(f"No IFC files were found for the following rule code: {code}. Please provide test files or verify the input.")
+        test_files.extend(paths)
+
+    file_pattern =  r".*\.ifc(\')?$" #matches ".ifc" and "ifc'"
+    test_files.extend([s.strip("'") for s in sys.argv if re.match(file_pattern, s)])
+
+    if not test_files: # for example, with 'pytest -sv'
```
aothms, line 40: Instead of deciding for running all tests at the end on empty `test_files`, I would decide for this based on an empty `[a for a in sys.argv[1:] if not a.startswith('-')]`. That way, running with an non-existant rule code will run nothing instead of everything.

test/test_main.py
```diff
@@ -11,8 +12,36 @@
...
+
+def get_test_files():
+    """
+    Option -> Example
+    Test files for a rule -> 'python3 test_main.py alb001'
+    Also applies for multiple rules -> 'python3 test_main.py alb001 alb002'
+    Test files for a single file -> 'python3 test_main.py <path>.ifc
+    Also applies for multiple files -> 'python3 test_main.py <path1>.ifc <path2>.ifc'
+    Codes and rules can also be combined -> 'python3 test_main.py alb001 <path>.ifc'
+    """
```
aothms, line 26: I think we need to test that an argument is either a valid rule code or an existing IFC file. In other cases we should do nothing and throw a big error. Otherwise it's easy to think you have ran the tests while you actually did nothing because you misttyped a filename.

aothms (changes requested):

PR #56, opened 2023-03-04
[X - Rule] VAL-186 - Add passed rule reporting functionality

JTurbakiewicz:


features/environment.py
```diff
@@ -14,7 +14,7 @@ def before_feature(context, feature):
     # between features so we need to preserve only the bottom two stack
     # frames when beginning a new feature.
     context._stack = context._stack[-2:]
-    
+    context.error_on_passed_rule = '--tags=-error_on_passed_rule' in sys.argv
```
aothms, line 17: I think we should use `--input` for this rather than `--tags`.

main.py
```diff
@@ -45,22 +50,21 @@ def run(filename, instance_as_str=True, rule_type=RuleType.ALL):
...
     # If this is a test file from the repository filter only the relevant scenarios
     feature_filter = []
-    try:    
+    try:
         rule_code = os.path.basename(filename).split('-')[1].strip().upper()
         if re.match(r'[A-Z]{3}[0-9]{3}', rule_code):
             feature_filter = ["-i", rule_code]
     except Exception as e:
         print(e)
+    proc = subprocess.run([sys.executable, "-m", "behave", *feature_filter, *tag_filter, "--tags=-run_through_pytest", "--define", f"input={os.path.abspath(filename)}", "-f", "json", "-o", jsonfn], cwd=cwd, capture_output=True)
```
aothms, line 66: I think we should use `--input` for this rather than `--tags`.

main.py
```diff
@@ -36,6 +40,7 @@ def do_try(fn, default=None):
         traceback.print_exc()
         return default
 
+
 def run(filename, instance_as_str=True, rule_type=RuleType.ALL):
```
aothms, line 44: I think rather than filtering later on 'Rule disabled' and 'Rule passed'. I'd rather have keyword arguments here, `include_disabled=False, include_passed=False`. Maybe not for now though, what do you think?

aothms (changes requested):

main.py
JTurbakiewicz: Yes, that would be cleaner for sure. 
I'll add this in a batch with passed rules cleanup (including variable naming change, better messages, and stuff like that).

features/environment.py
JTurbakiewicz: I am not sure if I get that 100%. As I understood it, `input` is the file path, so something that can be exactly the same, whether we use Gherkin as a script or as a test_main.py / validation repo. How do we recognize which instance invoked the script by `input`?

The --tags was an idea since it uses the negative tag functionality. That way it's "invisible" in the feature files, feature file devs don't need to think about it, but nonetheless, it explicitly defines the expected behavior.

aothms: Sorry I meant `--define` for passing arbitrary variables to behave rather than tags which have a specific meaning already. You're now relying on the fact that we don't (ok, and will not have) error_on_passed_rule as a regular tag. But using --define would be more pure I'd say.

We use --define for specifying `input` [0] hence my confusing. Thanks for asking.

[0] https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/main.py#L62

Issue #17, opened 2022-12-06
Typo


ciro-vendrame:
https://github.com/buildingSMART/ifc-gherkin-rules/blob/81b07c03b8c9726b4c0b580eb0efd63c1899ab2e/features/ALB002_Alignment-layout.feature#L52-L59

In these Scenarios the word "horizontal" in the agreement statement should be replaced by "vertical" and "cant" respectively

Ghesselink:
Fixed in https://github.com/buildingSMART/ifc-gherkin-rules/pull/6/commits/66f3c3c4c014422b1816bfd077ed089798ed1dc2

PR #57, opened 2023-03-14
RI14_Alignment-cant-shape-representation

JTurbakiewicz:


features/steps/thens/attributes.py
```diff
@@ -73,3 +73,31 @@ def _():
     misc.handle_errors(context, list(_()))
 
 
+@then('The type of attribute {attribute} must be {value}')
```
pjanck, line 76: Discussion needed: how is this function different from the one in line 64-73? What is the difference between 'must' and 'should', if any in this context?

features/steps/thens/attributes.py
```diff
@@ -73,3 +73,31 @@ def _():
...
+def step_impl(context, attribute, value):
+    if getattr(context, 'applicable', True):
+        errors = []
+        for instance in context.instances:
+            if isinstance(instance, tuple):
+                attribute_value = getattr(instance[0], attribute, 'Attribute not found')
+            else:
+                attribute_value = getattr(instance, attribute, 'Attribute not found')
+
+            if attribute_value != value:
```
pjanck, line 99: How does this behave with different underlying types (e.g. int, string, instance)?

pjanck (changes requested):

features/steps/thens/attributes.py
```diff
@@ -61,43 +61,31 @@ def step_impl(context, entity, other_entity):
...
 
 
-@then('The type of attribute {attribute} should be {expected_entity_type}')
+@then('The type of attribute {attribute} must be {expected_entity_type}')
 def step_impl(context, attribute, expected_entity_type):
 
     def _():
         for inst in context.instances:
+            if isinstance(inst, tuple):
+                inst = inst[0]
```
pjanck, line 70: Just curious: What does this do?

JTurbakiewicz: The `@given('Its attribute {attribute}')` given uses the below line in order to setup the scope for the following `@thens`
https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/features/steps/givens/attributes.py#L51

It returns either an attribute or an attribute wrapped in tuple/list. In this example it returns for example:

Captured stdout:
(#518=IfcShapeRepresentation(#17,'Axis','Curve3D',(#383)),)

So the above is unpacked from the touple.

features/steps/thens/attributes.py
JTurbakiewicz: `@then('The value of attribute {attribute} must be {value}')`

`{value}` is casted to a string by behave parser. For the purposes of the rules currently developed that's ok. 
For general purpose it's more complex, but I think it's very difficult to introduce type checks without assuming the .feature file writer is familiar with types.

PR #61, opened 2023-04-21
[V - Updated] Small fixes

Ghesselink:
Continuation of https://github.com/buildingSMART/ifc-gherkin-rules/pull/42

The following is added:

-`if len(set()): do X ` is changed to `if set(): do X`. In these cases, point of interest is existence and not length
- Type annotations for better documentation.
- In ALB002: Reverse related & relating. They are now in the wrong order in the steps.py file (I think due to the confusion of Nest/IsNestedBy).

### DIFF
--- a/features/steps/errors.py
+++ b/features/steps/errors.py
@@ -64,11 +64,11 @@ class IdenticalValuesError:
 
 @dataclass
 class InstanceCountError:
-    insts: ifcopenshell.entity_instance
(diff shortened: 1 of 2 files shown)


Ghesselink:
Most of this is fixed. If we'd like to change type annotations, I'd be better to open a new PR for all new implementations rather than fixing the merge conflicts.

PR #62, opened 2023-04-23
ALS004_Alignment-segment-shape-representation 

JTurbakiewicz:


Ghesselink (commented):
Nice work :) 

Should we formulate the rule (and the one of ALB002 I wrote before as well) slightly different to allow for more reuse of the already written code, e.i. 'The value must be X' https://github.com/buildingSMART/ifc-gherkin-rules/blob/47b6a14abb41da91b97d5ca65dc3c1f1ab243429/features/steps/thens/values.py#L11-L66

An idea would then be (something like) : 

_Scenario 1 : _Agreement on Representation being Axis__
```
Given a file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
And An IfcAlignmentSegment
And Its attribute RepresentationIdentifier
Then Its value must be 'Axis'
```

_Scenario 2 : Agreement on RepresentationType being Segment_
```
Given Repeat steps 1,2,3,4,5 #maybe change this to 1-5?
Then The value must be Segment
```
The functionality of this 'Repeat steps' can be found in the open PR for GEM005 https://github.com/buildingSMART/ifc-gherkin-rules/blob/0f2a25634cde3ea348faac6aa1be8845b2a82019/features/steps/giventhens/repeat_return.py#L6-L17

_Scenario 3 : Agreement on items being IfcCurveSegment_
```
Given repeat steps 1 - 5
Then The value must be 'IfcCurveSegment' #or change to 'items' and add @ to the step impl
```

features/RI009_Alignment-segment-shape-representation.feature
```diff
@@ -0,0 +1,14 @@
+@implementer-agreement
+@RI
+Feature: RI009 - Alignment segment shape representation
+The rule verifies, that each IfcAlignmentSegment uses correct representation.
```
pjanck, line 4:
```suggestion
The rule verifies that each IfcAlignmentSegment uses correct representation.
```

features/steps/thens/attributes.py
```diff
@@ -61,15 +61,30 @@ def step_impl(context, entity, other_entity):
...
+def step_impl(context, attribute, value):
+    if getattr(context, 'applicable', True):
+        errors = []
+        for instance in context.instances:
+            while isinstance(instance, (tuple, list)):
+                instance = instance[0]
+            attribute_value = getattr(instance, attribute, 'Attribute not found')
+            if attribute_value != value:
+                errors.append(err.InvalidValueError(instance, attribute, attribute_value))
+        misc.handle_errors(context, errors)
```
pjanck, line 89: I fear that it will introduce merge conflicts with #63 . Let's see.

test/files/ri009/pass-ri009-alignment-segment-shape-representation.ifc
```diff
@@ -0,0 +1,824 @@
+ISO-10303-21;
+/*  Testfile: pass-ri009-alignment-segment-shape-representation.ifc */
+/*  IFC-Requirement: ri009; Alignment segment shape representation in IFC */
+HEADER;
+FILE_DESCRIPTION(('ViewDefinition [Ifc4X3NotAssigned]'),'2;1');
+FILE_NAME ('pass-ri009-alignment-segment-shape-representation.ifc', '2023-03-14T15:36:48', ('bSI Railway Room - Technical Services'), ('aIFC4Rail'), 'Automatic Pass File Production', 'Python+IfcOpenShell', 'BC-002');
+FILE_SCHEMA(('IFC4X3'));
```
pjanck, line 7: I believe that this schema identifier is incorrect (similarly for other test files).

pjanck (changes requested):

features/RI009_Alignment-segment-shape-representation.feature
evandroAlfieri, line 1: Please replace RI009 with ALS004

features/RI009_Alignment-segment-shape-representation.feature
```diff
@@ -0,0 +1,14 @@
+@implementer-agreement
+@RI
+Feature: RI009 - Alignment segment shape representation
```
evandroAlfieri, line 3: Please replace RI009 with ALS004

test/files/ri009/README.MD
evandroAlfieri, line 1: Please replace RI009 with ALS004

test/files/ri009/fail-ri009-scenario01-wrong-items-type.ifc
evandroAlfieri, line 1: Please replace RI009 with ALS004

test/files/ri009/fail-ri009-scenario01-wrong-representationidentifier-value.ifc
evandroAlfieri, line 1: Please replace RI009 with ALS004

test/files/ri009/fail-ri009-scenario01-wrong-representationtype-value.ifc
evandroAlfieri, line 1: Please replace RI009 with ALS004

test/files/ri009/pass-ri009-alignment-segment-shape-representation.ifc
evandroAlfieri, line 1: Please replace RI009 with ALS004

evandroAlfieri (commented):
Looks good. Just the unfortunate renaming of all RI009.

features/steps/thens/attributes.py
JTurbakiewicz: For sure, but should be trivial to solve. I don't think there's much we can do if we want to have those rules split.

JTurbakiewicz:
> 



> Nice work :)
> 
> Should we formulate the rule (and the one of ALB002 I wrote before as well) slightly different to allow for more reuse of the already written code, e.i. 'The value must be X'
> 
> https://github.com/buildingSMART/ifc-gherkin-rules/blob/47b6a14abb41da91b97d5ca65dc3c1f1ab243429/features/steps/thens/values.py#L11-L66
> 
> An idea would then be (something like) :
> 
> _Scenario 1 : _Agreement on Representation being Axis__
> 
> ```
> Given a file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
> And An IfcAlignmentSegment
> And Its attribute Representation
> And Its attribute Representations
> And Its attribute RepresentationIdentifier
> Then Its value must be 'Axis'
> ```
> 
> _Scenario 2 : Agreement on RepresentationType being Segment_
> 
> ```
> Given Repeat steps 1,2,3,4,5 #maybe change this to 1-5?
> Then The value must be Segment
> ```
> 
> The functionality of this 'Repeat steps' can be found in the open PR for GEM005
> 
> https://github.com/buildingSMART/ifc-gherkin-rules/blob/0f2a25634cde3ea348faac6aa1be8845b2a82019/features/steps/giventhens/repeat_return.py#L6-L17
> 
> _Scenario 3 : Agreement on items being IfcCurveSegment_
> 
> ```
> Given repeat steps 1 - 5
> Then The value must be 'IfcCurveSegment' #or change to 'items' and add @ to the step impl
> ```
> 
> This way we get rid of some duplicate code and reuse `'The value must be {constraint}'` often. We could even distinquish between 'type' and 'value' in the gherkin implementation by adding` 'The {type/value/items} must be {constraint}` and perhaps the feature file itself becomes more readable (we could potentially use the scenario description as a description of what is being tested for the platform itself). But maybe I am making it too complicated, I'm not sure .. What do you think?

Hi!

Thank you for a great comment. 

I think to setup multiple scenarios with repeatable code we could also use the Background feature, as experimented here: https://github.com/buildingSMART/ifc-gherkin-rules/pull/37/files#diff-f12e75ce63549024543ab0c71a93b3cc880178205d26599482a386aa6041d205R5

I find it a bit more readable than` Given repeat steps 1 - 5`, but that's debatable of course. 

As for using the already present step: 'The value must be {constraint}' (thens/values.py). 
I think this is a good idea, but I would have to add a decorator, to have something like:

```
@then("The value must be equal to: {constraint}")  # HERE CHANGE
@then("The value must {constraint}")
@then("The values must {constraint}")
@then('At least "{num:d}" value must {constraint}')
@then('At least "{num:d}" values must {constraint}')
def step_impl(context, constraint, num=None):
```

And I would add the logic checking the value being equal to the provided value to this step, as currently {constraint} checks for 'identical', 'unique' mostly.  But I do agree that it would perhaps be cleaner to use this step.

What do you think? @Ghesselink

pjanck:
> An idea would then be (something like) :
> _Scenario 1 : _Agreement on Representation being Axis__
> ```
> Given a file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
> And An IfcAlignmentSegment
> And Its attribute Representation
> And Its attribute Representations
> And Its attribute RepresentationIdentifier
> Then Its value must be 'Axis'
> ```
> _Scenario 2 : Agreement on RepresentationType being Segment_
> ```
> Given Repeat steps 1,2,3,4,5 #maybe change this to 1-5?
> Then The value must be Segment
> ```

> I think to setup multiple scenarios with repeatable code we could also use the Background feature, as experimented here: https://github.com/buildingSMART/ifc-gherkin-rules/pull/37/files#diff-f12e75ce63549024543ab0c71a93b3cc880178205d26599482a386aa6041d205R5

I agree; the way I understand the proposal, I also think that this reinvents an existing Gherkin feature: the [Background](https://cucumber.io/docs/gherkin/reference/#background) keyword. 

> I would have to add a decorator

This is a code-architecture decision. I don't have any strong opinions on this.

Ghesselink:
> I also think that this reinvents an existing Gherkin feature

>  think to setup multiple scenarios with repeatable code we could also use the Background feature

Agree with both of you @pjanck @JTurbakiewicz . In this case it seems clearer. 

So then it would be something like : 
```
Background : AlignmentSegment Representation

Given a file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
And An IfcAlignmentSegment
And Its attribute RepresentationIdentifier

Scenario : Agreemen on Representation being Axis
Then The value must be 'Axis'

Scenario: Agreement on RepresentationType being Segment 
Then: The value must be 'Segment' 

Scenario: Agreement on IfcCurveSegment
Then: The value must be 'IfcCurveSegment'
```
It looks like if someone reads it for the first time they'll immediately know what's being tested, nice :-) 
Does this work? 

> I think this is a good idea, but I would have to add a decorator, to have something like

That's not required I think as it's already covered by the `@The value must {constraint} ` . In the example of '`The value must be 'Axis',` the constraint is 'be Axis' and 'be' part will be stripped from the constraint, leaving only 'Axis' in this case. https://github.com/JTurbakiewicz/ifc-gherkin-rules/blob/bb4a7f3c2b3db3a65e20f5b93e5a3e3fa39cd165/features/steps/thens/values.py#L21-L22 . 
Then there seems to be two options of implementation: 
- If there is nothing 'extra' stated in the constraint (such as identical/unique), as is the case above, implement it as 'equal to'. That also leaves the option open to implement the specific 'equal to' as actually equality between to values if this use case comes up. Then we'll need to check for the combination of a _literal 'be' followed by a value._ 
- Require for the gherkin statement to specifically state 'value must _be equal to_ Axis' and implement it in a similar way as `identical` and `unique`. https://github.com/JTurbakiewicz/ifc-gherkin-rules/blob/bb4a7f3c2b3db3a65e20f5b93e5a3e3fa39cd165/features/steps/thens/values.py#L29 . 
Now I'm writing this down, I think I'm leaning towards the first option. It seems to me that it will more understandable for someone else reading or making own gherkin statements (including future us).

But we could decide to be more specific about this, in which case I think it'll be better just to leave it in the {constraint} variable and treat in a similar way as 'identical' and 'unique https://github.com/JTurbakiewicz/ifc-gherkin-rules/blob/bb4a7f3c2b3db3a65e20f5b93e5a3e3fa39cd165/features/steps/thens/values.py#L29

JTurbakiewicz:
> > I also think that this reinvents an existing Gherkin feature
> 
> > think to setup multiple scenarios with repeatable code we could also use the Background feature
> 
> Agree with both of you @pjanck @JTurbakiewicz . In this case it seems clearer.
> 
> So then it would be something like :
> 
> ```
> Background : AlignmentSegment Representation
> 
> Given a file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
> And An IfcAlignmentSegment
> And Its attribute Representation
> And Its attribute Representations
> And Its attribute RepresentationIdentifier
> 
> Scenario : Agreemen on Representation being Axis
> Then The value must be 'Axis'
> 
> Scenario: Agreement on RepresentationType being Segment 
> Then: The value must be 'Segment' 
> 
> Scenario: Agreement on IfcCurveSegment
> Then: The value must be 'IfcCurveSegment'
> ```
> 
> This would look very clear indeed, nice :-)
> 
> > I think this is a good idea, but I would have to add a decorator, to have something like
> 
> That's not required I think as it's already covered by the `@The value must {constraint} ` . In the example of '`The value must be 'Axis',` the constraint is 'be Axis' and 'be' part will be stripped from the constraint, leaving only 'Axis' in this case. https://github.com/JTurbakiewicz/ifc-gherkin-rules/blob/bb4a7f3c2b3db3a65e20f5b93e5a3e3fa39cd165/features/steps/thens/values.py#L21-L22 . Then there seems to be two options of implementation:
> 
> * If there is nothing 'extra' stated in the constraint (such as identical/unique), as is the case above, implement it as 'equal to'. That also leaves the option open to implement the specific 'equal to' as actually equality between to values if this use case comes up. Then we'll need to check for the combination of a _literal 'be' followed by a value._
> * Require for the gherkin statement to specifically state 'value must _be equal to_ Axis' and implement it in a similar way as `identical` and `unique`. https://github.com/JTurbakiewicz/ifc-gherkin-rules/blob/bb4a7f3c2b3db3a65e20f5b93e5a3e3fa39cd165/features/steps/thens/values.py#L29 .
> 
> While I am writing this down, I think I'm leaning towards the first option. It seems to me that it will more understandable for someone else reading or making own gherkin statements (including future us). The implementation in python would then be to change https://github.com/JTurbakiewicz/ifc-gherkin-rules/blob/bb4a7f3c2b3db3a65e20f5b93e5a3e3fa39cd165/features/steps/thens/values.py#L21-L22 With something like :
> 
> ```
> while constraint.startswith('be'):
>    constraint = constraint[3:]
>    if re.search(r'\s', constraint): #check if str containts one value, i.e. not containing whitespaces
>       if not constraint in context.instances: # untested yet, perhaps needs to be modified to know which instances is causing the error). 
>           errors.append(some_error_category) #tbd 
>  while constraint.startswith('in'):
>      constraint = constraint[3:]
> ```
> 
> I'll test and work this out further, but would first like to know if you agree/I make sense :) @JTurbakiewicz

Makes sense. I'll play with it a little bit as well.

The additional constraint of ours is to have all the scenarios, so:
```

    Then The value of attribute RepresentationIdentifier must be Axis
    And The value of attribute RepresentationType must be Segment
    And The type of attribute Items must be IfcCurveSegment
```
    
   Be present in the same time. I'll check if background scenario lets us do that.

PR #64, opened 2023-04-23
RI007_Stationing-property

JTurbakiewicz:


pjanck (changes requested):
I wonder: do the test files fail *only* the rules proposed in this PR?

evandroAlfieri:
This rule is i) partially test-specific (cannot be part of the Validation Service); ii) partially touching on topics currently under discussion by the Implementers Forum. For these reasons, my suggestion is to put the development of this rule on hold. I can inform when the IF reaches a conclusion on this topic.

civilx64:
Cleaning up stale draft PRs.

PR #65, opened 2023-05-01
SPS002 - IfcFacility-part

JTurbakiewicz:


test/files/sps2/pass-sps002-ifcfacility-aggregation.ifc
```diff
@@ -0,0 +1,824 @@
+ISO-10303-21;
+/*  Testfile: pass-als003-alignment-cant-shape-representation.ifc */
+/*  IFC-Requirement: als003; Alignment cant shape representation in IFC */
+HEADER;
+FILE_DESCRIPTION(('ViewDefinition [Ifc4X3NotAssigned]'),'2;1');
+FILE_NAME ('pass-als003-alignment-cant-shape-representation.ifc', '2023-03-14T15:36:48', ('bSI Railway Room - Technical Services'), ('aIFC4Rail'), 'Python+IfcOpenShell', 'BC-002; Stationing on Alignment');
```
pjanck, line 6: Simile.

features/SPS002_IfcFacility-aggregation.feature
```diff
@@ -0,0 +1,10 @@
+@implementer-agreement
+@SPS
+Feature: SPS002 - IfcFacility aggregation
+The rule verifies that each IfcFacility is a part of an IfcProject, IfcSite or another IfcFacility.
+
+  Scenario: Agreement on each IfcFacility being a part of an IfcProject, IfcSite or another IfcFacility.
+
+      Given A file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
+      And An IfcFacility
+      Then Each IfcFacility must be a part of IfcProject or IfcSite or IfcFacility
```
pjanck, line 10: Are the `or`-s really the way to go, or should we rather use `xor` (the real intention in my view)? Is an IfcFacility being part of two or all three of these marked correct as well?

features/steps/thens/relations.py
```diff
@@ -40,6 +40,24 @@ def step_impl(context, entity, condition, directness, other_entity):
 
     misc.handle_errors(context, errors)
 
+@then('Each {entity} {condition} be a part of {other_entities}')
+def step_impl(context, entity, condition, other_entities):
+    stmt_to_op = ['must', 'must not']
+    assert condition in stmt_to_op
+    expect_part = True if condition == 'must' else False
```
pjanck, line 47: `else False` makes it potentially dangerous for future work, e.g. if new `stmt_to_op` will be introduced and this conditional not amended accordingly. Proposal: default should throw a not implemented exception (or similar).

features/steps/thens/relations.py
```diff
@@ -40,6 +40,24 @@ def step_impl(context, entity, condition, directness, other_entity):
...
+
+    errors = []
+
+    if context.instances and getattr(context, 'applicable', True):
+        other_entities = other_entities.split(' or ')
+        for ent in context.model.by_type(entity):
+            aggregates_relation = ent.Decomposes[0] # TODO check if relation is 1:1
```
pjanck, line 54: Is this TODO a remnant from the coding, or is it still a TODO?

pjanck (changes requested):
Initial comments.

Ghesselink:
Looks good so far, not sure whether to look really in-depth since it is a draft PR (still WIP?). 

I took the freedom to change the folder to 'sps002' and the filename for the fail file, since this caused the tests to fail. Now all the marks are green again and the autistic part of my brain can relax.

features/steps/thens/relations.py
```diff
@@ -40,6 +39,51 @@ def step_impl(context, entity, condition, directness, other_entity):
...
+            aggregated_table[applicable_entity].append(tbl_relationship_object)
+        else:
+            aggregated_table[applicable_entity] = [tbl_relationship_object]
+    errors = []
+
+    if context.instances and getattr(context, 'applicable', True):
+        for ent in context.model.by_type(entity):
+            for applicable_entity in aggregated_table.keys(): # check which applicable entity the currently processed entity is (inheritance), e.g IfcRailway -> IfcFacility
+                if ent.is_a(applicable_entity):
+                    break # TODO - > is the hierarchy important?
```
JTurbakiewicz, line 68: What the code does here is essentially running through the dict similar to the one below (simplified):

```
{IfcFacilityPartCommon : IfcFacility,
IfcFacility : [IfcProject, IfcSite, IfcFacility],
IfcRailway : [IfcProject, IfcSite, IfcRailway]}
```

Imagine the entity in question is IfcRailway which is clearly a IfcRailway but also a IfcFacility. 
The dicts in Python are not ordered in some Python versions + the .csv list in not ordered on input either.

pjanck: > in favor of having a more deterministic behavior of the rule

+1 from me. 

---

> aggregated to other IfcSpatialStructureElement

expands to 

> aggregated to IfcProject or ...

Note that `IfcProject` **is not** an `IfcSpatialStructureElement`. This is wrong.

---

> on the .csv implemented by the user

Is it planned that a user would be able to change these `.csv`? How do you ensure validity of such input?

---

> Should it be done in inheritance order... ?

I don't have any opinion on this. This seems to me as an implementation specific design choice. In my view, the stricter the rule is, the better the checking results are - see also my first point above.

---

Is this code any different from the one I reviewed in #67 ?

features/SPS002_Correct_spatial_breakdown.feature
```diff
@@ -0,0 +1,11 @@
...
+@SPS
+Feature: SPS002 - Correct spatial breakdown
+The rule verifies that spatial elements are aggregated as per the Spatial Composition Table.
+The possible allowed breakdown can be found in the csv file in the folder 'features/resources/spatial_CompositionTable.csv'
+
+  Scenario: Agreement on each IfcSpatialElement being aggregated to other IfcSpatialStructureElement as per spatial composition table.
+
+      Given A file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
+      And An IfcSpatialElement
+      Then Each IfcSpatialElement must be aggregated to other IfcSpatialStructureElement as per spatial_CompositionTable.csv
```
pjanck, line 11: Isn't this rule applicable to all version of the schema? @evandroAlfieri

features/resources/spatial_DecompositionTable.csv
```diff
@@ -0,0 +1,46 @@
...
+IfcMarinePart,IfcSpace
+IfcProject,IfcSite
+IfcProject,IfcBuilding
+IfcProject,IfcFacility
+IfcProject,IfcBridge
+IfcProject,IfcMarineFacility
+IfcProject,IfcRailway
+IfcProject,IfcRoad
+IfcProject,IfcExternalSpatialElement
+IfcProject,IfcSpace
```
pjanck, line 28: `IfcProject` **is not** an `IfcSpatialElement`. This contradicts the rule's documentation.

features/resources/spatial_DecompositionTable.csv
pjanck, line 1: I think that this `csv` is incomplete. For example, I believe that `IfcFacilityPart` can be decomposed in `IfcFacilityPart` as well. Where does this `csv` originate from?

pjanck (changes requested):

features/resources/spatial_DecompositionTable.csv
JTurbakiewicz: It comes from: https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Composition/Aggregation/Spatial_Decomposition/content.html#General-Usage

I'm happy to add to the csv. Is there a better source for that?

features/SPS002_Correct_spatial_breakdown.feature
```diff
@@ -0,0 +1,11 @@
+@implementer-agreement
+@SPS
+Feature: SPS002 - Correct spatial breakdown
+The rule verifies that spatial elements are aggregated as per the Spatial Composition Table.
+The possible allowed breakdown can be found in the csv file in the folder 'features/resources/spatial_CompositionTable.csv'
+
+  Scenario: Agreement on each IfcSpatialElement being aggregated to other IfcSpatialStructureElement as per spatial composition table.
```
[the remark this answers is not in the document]
JTurbakiewicz: Fair. Changing to:

Scenario: Agreement on each IfcSpatialElement being aggregated to other IfcSpatialElement as per spatial composition table.

features/resources/spatial_DecompositionTable.csv
JTurbakiewicz: Good point. Changing the documentation to:
```

  Scenario: Agreement on each IfcSpatialElement being aggregated as per spatial composition table.

      Given A file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
      And An IfcSpatialElement
      Then Each IfcSpatialElement must be aggregated as per spatial_CompositionTable.csv
```

features/steps/thens/relations.py
JTurbakiewicz: > 
> Note that `IfcProject` **is not** an `IfcSpatialStructureElement`. This is wrong.
> 


Fair point. Changed that in the documentation. 



> Is it planned that a user would be able to change these `.csv`? How do you ensure validity of such input?

No. I intended that only the rule developers will be able to change the csv, bad phrasing on my part. 
I meant that if there are some developments to the spatial composition / decomposition tables in the future, it would be quite easy for us (devs) to change the .csv in comparison to changing the python code. 

> I don't have any opinion on this. This seems to me as an implementation specific design choice. In my view, the stricter the rule is, the better the checking results are - see also my first point above.

I'll try to come up with the inheritance order solution.


> 
> Is this code any different from the one I reviewed in #67 ?

Very similar if not the same. I just needed it in both PRs in the same time.

Ghesselink (commented):
I feel we need a re-organization regarding relationships. This PR, ALB004 and some other PRs are very similar in functionality which make it quite confusing. Right now we have : 

- nesting, aggegrating, containing, assigning relationships
- An option to specify a relationship with a custom csv file
- A couple of nuances (e.g., directly/indirectly, constraint on number of relating elements)
- Slightly unstructured error classes

Maybe we could make a start with all punt the third point (the nuances) and place them into a single @then statement. Something like : 
```
@then('Each {entity} must be {relationship} to {other_entity}
@then('Each {entity} must be {relationship}{if_asper} {condition}
```
The structure is then always: 
Entity -> Relationship -> entity -> optional condition
And the more nuanced, 'advanced', statements are then : 
Entity -> Relationship -> advanced requirement -> Entity
Some examples of optional condition : 
- '`As per {table}'`
- `'If {another_entity} is present'`

And advanced requirements : 
'at most 3 instance(s) of'
'only by the following entities:'
This condition we could parse as in https://github.com/buildingSMART/ifc-gherkin-rules/pull/48

I feel this would result in clearer implementation and ease writing future rules and it's better to start with the cases that are the most similar (without the advanced statements). On the other hand, it will perhaps add more complication on github, as there are now many open PR and not a clear overview.
What do you think? @JTurbakiewicz

features/steps/thens/relations.py
```diff
@@ -40,6 +39,51 @@ def step_impl(context, entity, condition, directness, other_entity):
...
+
+    aggregated_table = {}
+
+    for d in tbl:
+        applicable_entity = d[ent_tbl_header]
+        tbl_relationship_object = d[relationship_tbl_header]
+        if applicable_entity in aggregated_table:
+            aggregated_table[applicable_entity].append(tbl_relationship_object)
+        else:
+            aggregated_table[applicable_entity] = [tbl_relationship_object]
```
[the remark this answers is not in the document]
Ghesselink: On a side note: I think the PR is failing the tests due to an error in the main repository (if you haven't noted already) https://github.com/buildingSMART/ifc-gherkin-rules/pull/69

features/SPS002_Correct_spatial_breakdown.feature
evandroAlfieri: Very much. Please @JTurbakiewicz extend the Given with "IFC2X3" and "IFC4". Also nice if you can add a couple of test files to corroborate this behavior. Thanks

features/steps/thens/relations.py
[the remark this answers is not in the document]
JTurbakiewicz: Good idea, done!

features/steps/thens/relations.py
JTurbakiewicz: Added sorting by inheritance.

JTurbakiewicz:
> 

Added something along those lines as a test here: https://github.com/buildingSMART/ifc-gherkin-rules/pull/67
What do you think? @Ghesselink 

I'll be adding some more test cases here, so I'll most likely change some stuff still, but it's perhaps a reasonable start. 

> I feel we need a re-organization regarding relationships. This PR, ALB004 and some other PRs are very similar in functionality which make it quite confusing. Right now we have :
> 
> * nesting, aggegrating, containing, assigning relationships
> * An option to specify a relationship with a custom csv file
> * A couple of nuances (e.g., directly/indirectly, constraint on number of relating elements)
> * Slightly unstructured error classes
> 
> Maybe we could make a start with all punt the third point (the nuances) and place them into a single @then statement. Something like :
> 
> ```
> @then('Each {entity} must be {relationship} to {other_entity}
> @then('Each {entity} must be {relationship}{if_asper} {condition}
> ```
> 
> The structure is then always: Entity -> Relationship -> entity -> optional condition And the more nuanced, 'advanced', statements are then : Entity -> Relationship -> advanced requirement -> Entity Some examples of optional condition :
> 
> * '`As per {table}'`
> * `'If {another_entity} is present'`
> 
> And advanced requirements : 'at most 3 instance(s) of' 'only by the following entities:' This condition we could parse as in #48
> 
> I feel this would result in clearer implementation and ease writing future rules and it's better to start with the cases that are the most similar (without the advanced statements). On the other hand, it will perhaps add more complication on github, as there are now many open PR and not a clear overview. What do you think? @JTurbakiewicz

PR #69, opened 2023-05-17
change scope emitted_one_passing

Ghesselink:
Latest commit https://github.com/buildingSMART/ifc-gherkin-rules/commit/2ccae8f4d1f3e263888d11fad7e86ba99c0f1b00 was failing the test due to a function not having access to a variable that is defined outside its scope. Defining it inside the scope fixes it. 

A bit silly one line PR, but I cannot push it directly.


Ghesselink:
Fixed in https://github.com/buildingSMART/ifc-gherkin-rules/commit/9ea10f680f60b821084a1f198bc6e421bd63eb87

PR #71, opened 2023-05-19
Concept template: Main points

aothms:
# Validation service

### Background

Concept templates are a core part of the IFC specification. They represent instance diagrams to indicate intended usage. The precise semantics of these diagrams are rather weak.

Gherkin is a plain-text format for writing structured specifications. It helps bridge the gap between technical and non-technical team members by providing a shared conceptualization of domain rules. In order for the specification to be executable the statements are paired with an implementation in an imperative programming language, Python in the case of the buildingSMART validation service. The advantages of this approach are the possibilities for parametrization and orchestration, as implementation function are assigned parameters and context. The shared conceptualization of domain rules helps with reporting and benefits from the flexibility of imperative programming.

### Main points

For this specific assessment we have looked at the main points in infrastructure alignment models. They are to be modeled as IfcReferents, using an IfcRelPositions relationship to IfcAlignmentSegments.

The starting point has been the following template.

main_points_template

Where traditionally concept templates were described in the mvdXML format, in the ifc4.3 specification and validation service, we have opted for a Graphviz .DOT notation format.

### Approach in validation service

The precise semantics of concept templates are rather weak, there is no indication of required cardinalities and no separation between antecedent and consequent common in rule languages (or applicability and requirement in IDS).

The approach in this assessment is to use concept templates as a graph query on the IFC input file and propose further constraints on the set of result tuples by means of gherkin statements. This set of result tuples is formed by taking every valid subgraph for the query input and "projecting" this to a flat tuple by taking the optional name bindings associated to nodes in the graph. `"PositionsName"` and `"SegmentName"` are examples of such name bindings in this case.

Such a projection is a common step in query execution, for example also in SPARQL (the query language for the semantic web) a SPARQL rendition of this template is provided below for comparison.

```
SELECT ?URI ?GlobalId ?Type ?MainPointName ?MainPointPredefinedType ?Positions ?PositionsName ?RelatedProducts ?DesignParameters ?StartTag ?EndTag ?SegmentName
WHERE {
  ?URI rdf:type ifcowl:IfcLinearElement .
  ?URI ifcowl:globalId_IfcRoot/express:hasString ?GlobalId .
    ?var4 ifcowl:relatingObject_IfcRelNests ?URI .
      ?var4 rdf:type ifcowl:IfcRelNests .
        ?var4 ifcowl:relatedObjects_IfcRelNests ?Type .
          ?Type rdf:type ifcowl:IfcReferent .
            ?Type ifcowl:name_IfcRoot ?MainPointName .
[...]

PR #72, opened 2023-05-28
Decorate handling of errors

Ghesselink:
EDIT 29/09 @todo add decorator to avoid repetition of `for inst in context.instances ` and handling of 'no errors' in `errors.RuleSuccessInst` . Open PR again when finished or other more suitable solution is found and opened in other PR. 

See conversation in https://github.com/buildingSMART/ifc-gherkin-rules/pull/71

> It would also be interesting to wrap the then step error handling in a decorator, so that the need for manually calling misc.handle_errors() is eliminated. Conceptually:

Finding a solution for this situation is not always this straightforward, for example when the errors are counted within the function scope. I'm not certain whether attempting to address this issue would provide benefits or instead introduce unnecessary complexity.


Ghesselink:
Outdated due to https://github.com/buildingSMART/ifc-gherkin-rules/pull/119

PR #74, opened 2023-07-27
add rule creation protocol package

Ghesselink:
Package to ensure adherence to the conventions for rule creation, as detailed in the README.md file.

https://github.com/buildingSMART/ifc-gherkin-rules/tree/main/features


test/rule_creation_protocol/protocol.py
```diff
@@ -0,0 +1,181 @@
...
+                message = f"expected {valid_separators} but found the following unvalid seperators {unvalid_separators}"
+            )
+        return value, values
+
+
+class RuleCreationConventions(ConfiguredBaseModel):
+    """Validate feature conventions
+    """
+    feature: Naming  # feature as in .feature file
+    dotfeature_file: Naming  # either user input or test file
```
JTurbakiewicz, line 98: As far as I noticed, those are the only spots where the `Naming` class is created. I thought of the grammar `feature: Naming` as type hinting only and was quite surprised to see it's the way the class is invoked. 

Is it a pydantic pattern? Or am I missing something?

JTurbakiewicz (approved):
I like this very much @Ghesselink. 
Would be happy to add a check or two just to try it myself :)

test/rule_creation_protocol/protocol.py
```diff
@@ -0,0 +1,301 @@
...
+        """must include a description of the rule that start with "The rule verifies that..."""  # allow for comma's
+        if not any(value.startswith(f"{prefix} rule verifies{optional_comma} that") for prefix in ("This", "The") for optional_comma in ("", ",")):
+            raise ProtocolError(
+                value=value,
+                message=f"The description must start with 'The rule verifies that', it now starts with {value}"
+            )
+        return value
+
+    @field_validator('steps')
+    def validate_steps(cls, value):
```
JTurbakiewicz, line 135: There are multiple "rules" checked here within one validate_steps method. While it has the drawback, that not all (only the first) step errors are listed at once per run, I am ok with this. It seems elegant to have all the steps check in one place. 

What do you think @Ghesselink ?

JTurbakiewicz, line 135: This set of validations is checked for every test file within a rule. I think it would be enough to check it once per rule (feature), not once per test file per rule.

While this can be changed, I think it's not terrible to leave it as is, because our test_main.py procedure is based on looping through every test file. @Ghesselink , what do you think?

test/test_main.py
```diff
@@ -42,13 +44,25 @@ def get_test_files():
...
-def test_invocation(filename):
-    results = list(run(filename))
+def test_invocation(request, filename):
+
+    def step_attrs_to_file(*attrs): 
+        def get_nested_item(d, keys):
+            for key in keys:
+                d = d.get(key, {})
+            return d
+        return list({get_nested_item(item, attrs) for item in behave_results})[0]
```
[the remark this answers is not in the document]
Ghesselink: Yes, I think so too. It would maybe be better to remove the whole logic to the protocol main package and just pass the behave_results to the function to keep it clean. Something like 
```
if '--validate-dev' in sys.argv or request.config.getoption("--validate-dev"):
   protocol.enforce(behave_results)
```

test/rule_creation_protocol/protocol.py
Ghesselink: Both :)

I've created a seperate Naming pydantic class because a lot of 'Naming' is equal in terms of implementation
GEM001 - Closed shell edge usage 
GEM001_Closed-shell-edge-usage.feature

What is tested here are the seperators; `-` `_` and ` `, `-`, respectively. I thought it would be easier to create a seperate pydantic class for this. In which the following elements are validated:

- Before : This is used to parse the name in elements that are tested seperately. If something goes wrong with parsing, an error is raised. The parsing is done with pyparsing with a pre-defined structure (First the rule code, then a seperator, than a collection of words and an optional .feature). The parsed results are then tested by the other `field_validators` 
- Rule code : test whether 'GEM001' is correct. And not GE001 or GEM01
- First seperators and second seperators. Checks if `_` is used between GEM001 and Closed-shell-edge usage and not `-` or `!`

test/rule_creation_protocol/protocol.py
Ghesselink: I'm not sure if I understand. `value` here is a list of step dictionaries and all the steps within a rule are tested here.
`all(d['keyword'] in ['Given', 'Then', 'And'] for d in value)`

for instance with ALS004
```
    Given A file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3" or "IFC4X3_ADD2"
    And An IfcAlignmentSegment
    Then The value of attribute RepresentationIdentifier must be Axis
    And The value of attribute RepresentationType must be Segment
    And The type of attribute Items must be IfcCurveSegment
```

```
[d['keyword'] for d in value]
['Given', 'And', 'And', 'And', 'Then', 'And', 'And']
```

Ghesselink: That would be enough for all the conventions that are tested in the `.feature` file. 
However, there are also conventions related to each test file (the naming for instance, and having .ifc as an extention). 

The place in test_main.py seemed to be the most natural. The costs is that each .feature file is tested multiple times (with each test file), but it's quite cheap since it runs very quick.
An alternative is check whether a `.feature `file is already checked and register that somewhere so it passes that validation for the next run (with the next testfile). But maybe we shouldn't complicate it too much, it's already quite challenging to keep a good overview of all the conventions already.

Ghesselink:
Closing. Continued by 
[convention_checking_rewrite](https://github.com/buildingSMART/ifc-gherkin-rules/tree/convention_checking_rewrite)

Issue #54, opened 2023-02-20
missing IfcShapeRepresentation.RepresentationType


evandroAlfieri:
**Segment** and **SectionedSurface** are missing from https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/features/resources/valid_RepresentationType.csv

PR #77, opened 2023-09-20
Als006

aothms:
nb: pull request includes ALS005, please ignore in review


features/steps/thens/attributes.py
```diff
@@ -69,12 +69,14 @@ def step_impl(context, entity, other_entity):
...
+    expected_entity_types = tuple(map(str.strip, expected_entity_type.split(' or ')))
+
     def _():
         for inst in context.instances:
             related_entity = misc.map_state(inst, lambda i: getattr(i, attribute, None))
             errors = []
             def accumulate_errors(i):
-                if not i.is_a(expected_entity_type):
+                if not any(i.is_a().lower() == x.lower() for x in expected_entity_types):
                     misc.map_state(inst, lambda x: errors.append(err.AttributeTypeError(False, x, [i], attribute, expected_entity_type)))
```
Ghesselink, line 80:
```suggestion
                    misc.map_state(inst, lambda x: errors.append(err.AttributeTypeError(False, x, i, attribute, expected_entity_type)))
```
Change `[i] `to` i`

It's not code that is new in this PR I see. The tests for
_'fail-als006-sectioned-solid-horizontal-wrong-item.ifc'_ 
returns the error
`expected type IfcCompositeCurve .... but found [#79=IfcGradientCurve((#8..#54,$)]`

Which gives the false impression that the error is due to the entity being in a list `['IfcGradientCurve']` rather than the  entity itself` 'IfcGradientCurve'`. 
When running the test, it appears that it does evaluate the value 'IfcGradientCurve' (in line 79, the line above), so it's just the text output. 

I'm almost sure that this usage of the list pertains to an outdated error rendering class, possibly related to one of the alignment rules. Omitting it has no impact when executing pytest.

Ghesselink (approved):
Looks good. It's great to see that we're effectively reusing most of the Python code we've previously written.

It was only missing documentation for the testfiles. I've commited it to the branch of this PR.

features/ALS006_Alignment-horizontal-shape-representation.feature
```diff
@@ -0,0 +1,12 @@
...
+Feature: ALS006 - Horizontal alignment shape representation
+  Scenario: Agreement on each IfcAlignmentHorizontal using correct representation
+
+      Given A file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
+      And An IfcAlignmentHorizontal
+      And Its attribute Representation
+      And Its attribute Representations
+      Then The value of attribute RepresentationIdentifier must be Axis
+      And  The value of attribute RepresentationType must be Curve2D
+      And  The type of attribute Items must be IfcCompositeCurve or IfcIndexedPolycurve or IfcPolyline
```
pjanck, line 12: Comparing lines 6 and 12: how come one needs `"`, while the other doesn't? This seem inconsistent and imho unnecessarily complicates the rule writing / reading.

test/files/als006/README.md
```diff
@@ -0,0 +1,6 @@
...
+|-------|---------|--------|-------|
+| pass-als006-sectioned-solid-horizontal.ifc| success |    | |
+| fail-als006-sectioned-solid-horizontal-wrong-identifier.ifc| fail | On instance #200001=IfcShapeRepresenta...,(#54)) the following invalid value for RepresentationIdentifier has been found: Footprint| The value of attribute Representation must be Axis, but is Footprint instead  |
+| fail-als006-sectioned-solid-horizontal-wrong-item.ifc| fail |The instance #20001=IfcShapeRepresentation(#1001, 'Axis', 'Curve2D',(#79)) expected type 'IfcCompositeCurve or IfcIndexedPolycurve or IfcPolyline' for the attribute Items, but found #79=IfcGradientCurve((#8..#54,$) | On entity #20001=IfcShapeRepresentation(#1001, 'Axis', 'Curve2D',(#79)). The type of attribute Items must be IfcCompositeCurve or IfcIndexedPolycurve or IfcPolyline, but found IfcGradientCurve instead |
+| fail-als006-sectioned-solid-horizontal-wrong-type.ifc| fail | On instance #20001=IfcShapeRepresenta...,(#54)) the following invalid value for RepresentationType has been found: Curve3D |The value of attribute RepresentationType must be Curve2D, but found Curve3D instead |
```
pjanck, line 6: This table seems to have more information than the one for ALS005 (#76). Can this be unified?

Meta: are there any guidelines, how these tables should look like?

pjanck (approved):
Ignored ALS005 content as instructed.

test/files/als006/README.md
Ghesselink: The guidelines are in : 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/features/README.md

A template of how these tables should look like specifically in https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/features/README.md#table-template-for-unit-test-files

The error column should display the error message that is generated by the validation service. In 'Description' there's the option to add something specific to that file if necessary. But ideally, the error message that is generated automatically should be descriptive enough that this extra information is not necessary. 

The table in als005 is generated programmatically with the script in ; https://github.com/buildingSMART/ifc-gherkin-rules/pull/83

features/ALS006_Alignment-horizontal-shape-representation.feature
Ghesselink: I understand how it might appear inconsistent. We're currently experimenting to find a balance between ease of Python implementation and the human readability of Gherkin statements.

Currently, it's due to the following given statement
https://github.com/buildingSMART/ifc-gherkin-rules/blob/05329ef21559fc1de7113626f716ead32a5b0320/features/steps/givens/attributes.py#L34

This implementation results in 'Schema Identifier' being parsed into the 'field' variable, and the IFC versions into the 'values' variable. Additionally, the 'field' variable is used to parse the value 'Model View Definition' within the same function. The " symbol is essential for Gherkin to delineate where the 'field' variable concludes and 'values' begins.

For instance, with 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/05329ef21559fc1de7113626f716ead32a5b0320/features/SPS001_Basic-spatial-structure-for-buildings.feature#L15-L16

I'll merge this PR and move the discussion to the 'custom parsing' issue and PR. I hope that's ok. 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/48
https://github.com/buildingSMART/ifc-gherkin-rules/issues

PR #81, opened 2023-09-21
Sps004

aothms:
Only review https://github.com/buildingSMART/ifc-gherkin-rules/commit/f159851fe08aadcc01260f83398d252559488d06 thanks


features/SPS004_No_combination_of_containment_and_positioning.feature
```diff
@@ -0,0 +1,18 @@
+@implementer-agreement
+@SPS
+Feature: SPS004 - No combination of containment and positioning
+Each IfcProduct shall be either be contained in one and only one IfcSpatialStructureElement, or contained in one and only one IfcLinearPositioningElement/
```
pjanck, line 4:
```suggestion
Each IfcProduct can either be contained in one and only one IfcSpatialStructureElement, or contained in one and only one IfcLinearPositioningElement, but not both.
```

The way I interpret the implementation below, it is only mutually exclusive for both to be present; but they both might be empty as well.

test/files/sps004/fail-sps004-scenario01-with_position_with_contain.ifc
```diff
@@ -0,0 +1,38 @@
+ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION(('ViewDefinition [CoordinationView]'),'2;1');
+FILE_NAME('','2023-09-21T11:25:23',(''),(''),'IfcOpenShell-v0.7.0-64d98f129','IfcOpenShell-v0.7.0-64d98f129','');
+FILE_SCHEMA(('IFC4X3_ADD1'));
```
pjanck, line 5: CoordinationView with IFC4X3_ADD1? Intentional?

pjanck (changes requested):

test/files/sps004/fail-sps004-scenario01-with_position_with_contain.ifc
aothms: This probably applies to the majority of files. Let's save this for later #91

features/SPS004_No_combination_of_containment_and_positioning.feature
aothms: Good point, also created a test case for that (which indeed passes without code changes).

PR #83, opened 2023-09-21
handle markdown for testfiles

Ghesselink:
Hopefully a useful tool to generate markdown files for testfiles. 

Usage
Include the rule_code
`python3 generate-markdown.py sys001`
basic

For a single test file, if there are multiple results pertaining to one rule, the table will have additional rows to accommodate each result.
complete

Descriptions can be added or modified by passing the --file-desc argument to the command, followed by the filename and its respective description. Descriptions for multiple files can be provided, and the rule_code can also still be specified if needed.
`python3 generate
-markdown.py sys001 --file-desc pass-sys001-sink_source.ifc example pass file --file-desc fail-sys001-scenario02-sink.ifc example fail file`
update_description

Alternatively, you can update the description for an individual file in a already existing markdown table. Please note that you'll need to provide the updated_file.ifc twice: first, to gather the test files, and secondly, after the --file-desc flag, followed by the updated description."
`python3 generate-markdown.py sys001 fail-sys001-scenario02-no_ports.ifc --file-desc fail-sys001-scenario02-no_ports.ifc updated description single file`
single_file_update


test/test_main.py
```diff
@@ -42,10 +43,65 @@ def get_test_files():
...
+            file.write(pd.DataFrame([markdown_result_testfile]).to_markdown(index=False))
+
+
 @pytest.mark.parametrize("filename", get_test_files())
 def test_invocation(filename):
     results = list(run(filename))
     base = os.path.basename(filename)
+
+    if '--generate-markdown' in sys.argv:
+        handle_testfile_markdown(results, base)
```
aothms, line 103: I think I'd rather move this to a separate file altogether. The way it is currently written requires you to run the tests to generate markdown which I think is unnecessary.

aothms (changes requested):

aothms:
For SYS001 it generated some weird formatting (due to multiline errors?)

afbeelding

https://github.com/buildingSMART/ifc-gherkin-rules/commit/1311f58bc5fe3e7fbc1ae7411e1cf11dc738cc9b

Ghesselink:
Answering @aothms 


> And for the pass results it generated a "Rule passed" message.

I followed the conventions in  https://github.com/buildingSMART/ifc-gherkin-rules/tree/main/features#readme . 
It makes somse sense to include the 'pass' files in, as that way it's still possible to add a description (for instance with 'ALB002'). 

> I tested your script on the GRP001 PR [...]  I didn't get any content in the error column for the fail files. Is that intended?

> For SYS001 it generated some weird formatting (due to multiline errors?)

I think for both GRP001 and SYS001 the error was due to multiline errors. The handling of multiple lines in one markdown cell seems to be problematic in github 
https://github.com/markdown-it/markdown-it/issues/406
https://stackoverflow.com/questions/47324785/how-can-i-add-a-table-with-multi-row-cells-to-a-readme-in-vsts/47335483#47335483

There are some workarounds, but the most straightforward and often used I think is just using html. This should be rendered well in Github. In case of multiple results pertaining to one rule, the table will have additional rows to accommodate each result. Furthermore, descriptions are adjustable even after generating a markdown file. 
The old markdown table will still be parsed in the script (for instance, when updating a description after 'Error')

After modifying the script (see also on the top of the PR). 
`SYS001`

complete

`GRP001`

grp001

PR #84, opened 2023-09-21
SYS001 / RI-309 cable signal flow

aothms:
This contains some overlap with #37 which is in an unmergeable state currently. So some temporary excerpts have been extracted in the interest of time.

Feel free to propose something else than `SYS`.


features/SYS001_Cable_signal_flow.feature
```diff
@@ -0,0 +1,17 @@
+@implementer-agreement
+@ALB
```
JTurbakiewicz, line 2: Possibly SYS here?

JTurbakiewicz:
Same comment about README.MD in tests, if you think it's required, you can ping me and I'll fill that in.

features/SYS001_Cable_signal_flow.feature
aothms: Thanks

PR #94, opened 2023-09-29
add guidelines

Ghesselink:
From https://github.com/Ghesselink/Rule_Creation


Ghesselink:
Outdated due to https://bsi-technicalservices.atlassian.net/browse/VAL-280
Backup of info in https://github.com/Ghesselink/Rule_Creation

PR #95, opened 2023-09-30
fix testing disabled rules with pytest

Ghesselink:
Failed error https://github.com/buildingSMART/ifc-gherkin-rules/commit/eedb41d84b549daeafb8d28379f0fd1aba409d7a

Is due to 
```
    if base.startswith("fail-"):
        assert len(results) > 0
```
In combination with 
   ` results = [result for result in results if result[4] != 'Rule disabled']`

https://github.com/buildingSMART/ifc-gherkin-rules/blob/eedb41d84b549daeafb8d28379f0fd1aba409d7a/test/test_main.py#L62C1-L63C32

The rules with 'Rule Disabled' as message gets filtered out. This way, even with files that start with 'fail' :` len(results) == 0`, 

A solution is to not overwrite the original results  from` main.run()` and check whether the rule is available before the assertion : 
`    if base.startswith("fail-") and not any(description == 'Rule disabled' for description in [result[4] for result in gherkin_results]):`

This way in combination with using the original results in tabulate, we still get a meaningful error message with testing without triggering `assert`

example

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -44,22 +44,24 @@ def get_test_files():
 
 @pytest.mark.parametrize("filename", get_test_files())
 def test_invocation(filename):
-    results = list(run(filename))
+    gherkin_results = list(run(filename))
     base = os.path.basename(filename)
     # if base.startswith("pass-"):
-    results = [result for result in results if result[4] != 'Rule disabled']
-    results = [result for result in results if 'Rule passed' not in result[4]]
+    results = [result for result in gherkin_results if result[4] != 'Rule disabled']
+    results = [result for result in results if 'Rule passed' not in results[4]]
     print()
     print(base)
     print()
     print(f"{len(results)} errors")
-    if results:
+
+    if gherkin_results:
         print(tabulate.tabulate(
-            [[c or '' for c in r] for r in results],
-            maxcolwidths=[30] * len(results[0]),
+            [[c or '' for c in r] for r in gherkin_results],
+            maxcolwidths=[30] * len(gherkin_results[0]),
             tablefmt="simple_grid"
         ))
-    if base.startswith("fail-"):
+
+    if base.startswith("fail-") and not any(description == 'Rule disabled' for description in [result[4] for result in gherkin_results]):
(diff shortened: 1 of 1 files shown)


aothms:
Oh yes that's right. For the other rule which is disabled `alb001` I patched it by just changing every file from `fail-` to `pass-`. Which is also not ideal.

https://github.com/buildingSMART/ifc-gherkin-rules/tree/main/test/files/alb001

Maybe another way would be to inpect which scenarios are disabled and not to use any of their test files? But I guess it's a larger effort.

Ghesselink:
See https://github.com/buildingSMART/ifc-gherkin-rules/pull/95/commits/4e64e7308af537a8e02769414d3bfd51d08cde9a

It appears that this can be accomplished during the selection of test files. While we cannot currently, at that point, access tags via  `context`, we can parse the feature files to determine if `@rule-disabled` is present. If it is, we can exclude the respective filenames containing the disabled rule.

```
def exclude_disabled_rules(filepath):
    tags = []

    with open(filepath, 'r') as file:
        for line in file:
            stripped_line = line.strip()  # Remove any leading or trailing whitespace
            if stripped_line.startswith("@"):
                tags_on_line = stripped_line.split()
                tags.extend(tags_on_line)
    return '@disabled' in tags

def rule_disabled(code: str) -> bool:
    feature_paths = glob.glob(os.path.join(os.path.dirname(__file__), "../features/**/*.feature"), recursive=True)
    return next((exclude_disabled_rules(path) for path in feature_paths if os.path.basename(path).lower().startswith(code)), False)
```

`Rule_disabled` checks, as the name suggests, whether the rule is disabled or not. 
```
       paths = glob.glob(os.path.join(os.path.dirname(__file__), "files/", code.lower(), "*.ifc"))
        if not paths:
            print(f"No IFC files were found for the following rule code: {code}. Please provide test files or verify the input.")
        elif rule_disabled(code):
            print(f"The rule identified by code '{code}' is currently marked as 'disabled'. Any associated test files will not be taken into consideration")
            continue #exclude testfiles associated
```

PR #89, opened 2023-09-24
include ifc4x3_add2

Ghesselink:
https://github.com/buildingSMART/ifc-gherkin-rules/issues/88

PR #102, opened 2023-10-22
Pse001 more checks

JTurbakiewicz:


buildingsmart-private (approved):

berlotti (approved):

features/steps/errors.py
```diff
@@ -152,6 +143,46 @@ class InvalidValueError(RuleState):
     def __str__(self):
         return f"On instance {misc.fmt(self.inst)} the following invalid value for {self.attribute} has been found: {self.value}"
 
+@dataclass
+class InvalidPropertySetDefinition(RuleState):
+    inst: ifcopenshell.entity_instance
+    object: ifcopenshell.entity_instance
+    name: str = None
+    types: list = None
+    template_type_enum: str = None
```
Ghesselink: This is a bit contradictory. When 'name' is not given, it default to None. But this is not a string, so this will lead to an AttributeError. 
A way around it is Typing.Optional[str] = None. 

I've modified it in https://github.com/buildingSMART/ifc-gherkin-rules/pull/102/commits/b5d11133e964d69e66f9b35c419b2d95ba8686c5 . Hope that's ok

features/steps/errors.py
```diff
@@ -152,6 +143,46 @@ class InvalidValueError(RuleState):
...
+
+
+@dataclass
+class InvalidPropertyDefinition(RuleState):
+    inst: ifcopenshell.entity_instance
+    property: ifcopenshell.entity_instance
+    accepted_values: list = None
+    accepted_type: str = None
+    accepted_data_type_value: str = None
+    value: str = None
```
Ghesselink: Same here as mentioned above. 
Fixed in [change type hints other error class](https://github.com/buildingSMART/ifc-gherkin-rules/commit/2ad27ef770bc14b5d9941c8be03d66db5c91385e)

features/PSE001_Ifcpropertyset-validation.feature
```diff
@@ -0,0 +1,39 @@
...
+      Then The IfcPropertySet Name attribute value must use predefined values according to the IFC4_definitions.csv table
+      Then The IfcPropertySet must be assigned according to the property set definitions table IFC4_definitions.csv
+      Then Each associated IfcProperty must be named according to the property set definitions table IFC4_definitions.csv
+      Then Each associated IfcProperty must be of type according to the property set definitions table IFC4_definitions.csv
+      Then Each associated IfcProperty value must be of data type according to the property set definitions table IFC4_definitions.csv
+
+
+    Scenario: Agreement on each IfcPropertySet correctly defining an applicable entity.
+
+      Given A file with Schema Identifier "IFC4X3"
```
Ghesselink, line 32: I think we should be careful here with ifc 4 versioning, but that's a topic for at the other PR perhaps 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/89

features/resources/property_set_definitions/IFC4x3_definitions.csv
Ghesselink, line 1: What is the difference between IFC4X3 and IFC4 in terms of property sets? 

That would be useful to know, given the conversation we had two weeks ago about IFC versioning and making a rule to check for versioning in general.

features/steps/thens/relations.py
```diff
@@ -131,6 +126,157 @@ def step_impl(context, related, relating, other_entity, condition):
                         yield(err.RuleSuccessInst(True, inst))
 
 
+@then('The IfcPropertySet Name attribute value must use predefined values according to the {table} table')
+@then('The IfcPropertySet must be assigned according to the property set definitions table {table}')
+@then('Each associated IfcProperty must be named according to the property set definitions table {table}')
+@then('Each associated IfcProperty must be of type according to the property set definitions table {table}')
+@then('Each associated IfcProperty value must be of data type according to the property set definitions table {table}')
```
Ghesselink, line 133: I think we could use something more elegant to parse these sentences. But that would be for another PR 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/48

features/steps/thens/relations.py
```diff
@@ -131,6 +126,157 @@ def step_impl(context, related, relating, other_entity, condition):
...
+            for property_def in make_obj(property_set_attr['property_definitions']):
+                accepted_values['property_names'].append(property_def['property_name'])
+                accepted_values['property_types'].append(property_def['property_type'])
+                accepted_values['data_types'].append(property_def['data_type'])
+
+            accepted_values['applicable_entities'] = make_obj(property_set_attr['applicable_entities'])
+
+            try:
+                accepted_values['applicable_type_values'] = make_obj(property_set_attr.get('applicable_type_value', '')).split(',')
+            except json.decoder.JSONDecodeError:
```
Ghesselink, line 172: Can you maybe explain what happens here? With `json.decoder.JSONDecodeError:`

features/steps/thens/relations.py
```diff
@@ -131,6 +126,157 @@ def step_impl(context, related, relating, other_entity, condition):
...
+
+            accepted_values = establish_accepted_pset_values(name, property_set_definitons)
+
+            if accepted_values:  # If not it's a custom Pset_ prefixed attribute, e.g. Pset_Mywall (no need for further Pset_ checks),
+
+                if 'IfcPropertySet must be assigned according to the property set definitions table' in context.step.name:
+                    try:
+                        relations = inst.PropertyDefinitionOf  # IFC2x3
+                    except AttributeError:
+                        relations = inst.DefinesOccurrence # IFC4x3
```
Ghesselink, line 196: I think this should be documented a bit better and linked to the 'Given' statement. 
But, again, this is another discussion related to https://github.com/buildingSMART/ifc-gherkin-rules/pull/89

Also adding these notes a bit for my own reference if you don't mind :)

Do you have more implementations here in the code that is depending of IFC versioning?

Ghesselink (approved):
Nice work. 
A couple of minor remarks about type hinting, I've done some commits regarding this myself. 

Also some remarks about reusability and ifc versioning. I think that's a topic other discussions though and should not block this PR. If you don't mind, I also added these notes for my own reference. 

I think if you resolve the conversations you can merge it, as there is then 1 approving review.

features/steps/errors.py
JTurbakiewicz: Hi, good point. Thank you very much <3

features/steps/errors.py
JTurbakiewicz: Thank you!

features/PSE001_Ifcpropertyset-validation.feature
JTurbakiewicz: That's a good point. I'm changing `IFC4X3` to `IFC4X3_ADD2`. Thank you.

features/resources/property_set_definitions/IFC4x3_definitions.csv
JTurbakiewicz: I got the property sets information from:

IFC4: https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/annex/annex-a/general-usage/IFC4_ADD2-psd.zip_ (mind the _ after zip)
IFC 4.3 ADD2: https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/annex-a-psd.zip, based on https://technical.buildingsmart.org/standards/ifc/ifc-schema-specifications/

I wasn't comparing the differences, but from a quick glimpse now, there's a whole bunch of changes, for example different Pset_ present (e.g. Pset_Address), different template_type (e.g. Pset_ActionRequest), different applicable types/entities (e.g. Pset_WallCommon). I'm pretty sure the IfcProperty requirements changed as well.

features/steps/thens/relations.py
JTurbakiewicz: Agreed. I have multiple then's here, as it was requested to have the checks listed in the .feature files, so:

      Then The IfcPropertySet Name attribute value must use predefined values according to the IFC4X3_ADD2_definitions.csv table
      Then The IfcPropertySet must be assigned according to the property set definitions table IFC4X3_ADD2_definitions.csv
      Then Each associated IfcProperty must be named according to the property set definitions table IFC4X3_ADD2_definitions.csv
      Then Each associated IfcProperty must be of type according to the property set definitions table IFC4X3_ADD2_definitions.csv
      Then Each associated IfcProperty value must be of data type according to the property set definitions table IFC4X3_ADD2_definitions.csv
      
      Instead of 
      
      Then The IfcPropertySet must use predefined values according to the IFC4X3_ADD2_definitions.csv table
      
      
Which makes sense I think, as it's way more explicit and clear for the user what is checked. 
But then again I don't want to write multiple steps in .py, as it's so compact of a check, that it makes sense in my opinion to do it in one function. That's why I have multiple decorators.

JTurbakiewicz: I will resolve the conversation, although I agree, that something better can be done here (but I don't know what without major code changes to the whole repository)

JTurbakiewicz:
Checks failed for IFC4X3_ADD2. The error root is most likely known (ifcopenshell version related). I'll look into it.

Ghesselink:
It seems ifcopenshell version related indeed. Probably when you merge main into this branch it'll work. 

But I think it'd be better to leave it as it was and handle all the versioning together in a single PR. And merge this one.

But it's nice to have notes in the code implementation which attributes were used for which IFC version :)

features/steps/thens/relations.py
JTurbakiewicz: Good question as I had to remind myself why it's there :) 

The reason was that I was trying to make a python object from string like: `"['IfcActor', 'IfcBuilding', 'IfcSite']"`. If a string was '', json.loads() produced an error. Perhaps it was cleaner to check if a string was empty instead of the try/except. 

I've checked though, and for `applicable_type_values `there's no need to use json at all, as it's always in a nice string format, like: `"IfcActor,IfcBuilding,IfcSite"`. 

I've changed those lines removing this unnecessary redundancy. Thank you for a very valid point.

features/steps/thens/relations.py
JTurbakiewicz: Here I am not sure what to do. I've added a bit more description, that's a good first step for sure. 

I don't know how to nicely link it to the 'Given' statement, without complicating or duplicating the python function. I still am tempted to keep the current solution, as it's just one try/except block. But I am happy to discuss more, I'm leaving the conversation unresolved for a day or two. 

There are no more places depending on IFC versioning that I'm aware of.

Ghesselink: I think for now it's more than fine. Thanks for all the helpful comments :-) 

I don't think there's anything you can do. I'm trying to implement a warning when there is another version than IFC4X3_ADD2 (to be edited with newer versions in the future) and uniform the other IFC versions to simplify the versioning, unless there is a rule designed for a specific version. That's why the comments are really helpful.

For the rest the code looks really good. Maybe some minor things, for instance 

 ```
          def make_obj(s):
                if s:
                    return json.loads(s.replace("'", '"'))
                else:
                    return ''
```
Could be simply
          def make_obj(s):
                return json.loads(s.replace("'", '"')) if s else ''

File: features/steps/givens/attributes.py [hunk 7]
```diff
@@ -25,6 +25,19 @@ def step_impl(context, attribute, value):
     )
 
 
+@given("Its attribute {attribute} {condition} with {prefix}")
+def step_impl(context, attribute, condition, prefix):
+    assert condition in ('starts', 'does not start')
+
+    if condition == 'starts':
+        context.instances = list(
+            filter(lambda inst: hasattr(inst, attribute) and str(getattr(inst, attribute)).startswith(prefix), context.instances)
```
aothms, line 34:
```suggestion
            filter(lambda inst: hasattr(inst, attribute) and str(getattr(inst, attribute, '')).startswith(prefix), context.instances)
```

I would probably add a default to the getattr just in order to not have it throw an exception in case of a non-existing attr. That would be our mistake in the scenario writing though, but still.

features/steps/thens/relations.py
```diff
@@ -131,6 +126,155 @@ def step_impl(context, related, relating, other_entity, condition):
...
+                    related_objects = inst.DefinesType
+                    for obj in related_objects:
+                        if accepted_values['template_type'] and accepted_values['template_type'] in ['PSET_OCCURRENCEDRIVEN', 'PSET_PERFORMANCEDRIVEN']:
+                            yield (err.InvalidPropertySetDefinition(False, inst=inst, object=obj, name=name, template_type_enum=accepted_values['template_type']))
+
+                        correct = [obj.is_a(accepted_object) for accepted_object in accepted_values['applicable_type_values']]
+                        if not any(correct):
+                            yield (err.InvalidPropertySetDefinition(False, inst, obj, name, accepted_values['applicable_type_values']))
+                        elif context.error_on_passed_rule:
+                            yield (err.RuleSuccessInst(True, inst))
```
aothms, line 220: This monster block for all the rules is a bit funky. I think it's fine for now. My main point would be that it's not clear when the RuleSuccessInst is supposed to be emitted, but anyway that is going to have to be refactored with the other PR coming in.

I would probably prefer 5 differnet functions that make use of some utility function, but either way is fine.

aothms (approved):

features/steps/thens/relations.py
JTurbakiewicz: I agree. 

My plan of action:
 1. Merge this to main.
 2. Merge main into PR decorator branch. 
 3. Change PSE001 part in the decorator branch. 
 
 I think it's the cleanest way to tackle it.

PR #114, opened 2023-11-19
check schema version

Ghesselink:
```
Given any IFC file
Then the schema identifier must be IFC2X3 or IFC4 or IFC4X3_ADD2 
```
For reference:
| Version | Name          | Schema id   | Common jargon |
|---------|---------------|-------------|---------------|
| 4.3.2.0 | IFC4.3 ADD2   | IFC4X3_ADD2 | IFC4.3        |
| 4.0.2.1 | IFC4 ADD2 TC1 | IFC4        | IFC4          |
| 2.3.0.1 | IFC2x3 TC1    | IFC2X3      | IFC2x3        |


features/GEM005_Building-shape-representation.feature
```diff
@@ -6,8 +6,7 @@ The rule verifies that an IfcBuilding has a correct representation
 
   Scenario: Agreement on empty IfcBuilding using correct representation
 
-      Given A file with Schema Identifier "IFC2X3" or "IFC4" or "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
-      And An IfcBuilding
+      Given An IfcBuilding
```
JTurbakiewicz, line 9: While I'm of course fine with leaving this as you propose, just a thought. Do we care what happens when IFC 5.0 comes out?  The rule written without the Schema identifier will fire for every subsequent IFC version. For most (all?) rules this might be what we want, for some perhaps not.

features/OJP001_Relative-placement-for-elements-aggregated-to-another-element.feature
```diff
@@ -7,7 +7,7 @@ with an PlacementRelTo attribute pointing to the IfcLocalPlacement of the contai
 
   Scenario: Agreement on the relative placement of IfcElements being a part of another IfcElement through the relationship IfcRelAggregates
 
-      Given A file with Schema Identifier "IFC2X3"
+      Given A file with Schema Version "IFC2X3"
```
JTurbakiewicz, line 10: Is there a reason to use `Version `here?

features/SPS004_No-combination-of-containment-and-positioning.feature
```diff
@@ -6,14 +6,14 @@ The rule verifies that each IfcProduct can either be contained in one and only o
 
   Scenario: Agreement on elements being contained cannot be positioned
 
-      Given A file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3"
+      Given A file with Schema Version "IFC4.3"
```
JTurbakiewicz, line 9: The same question about `Version`. In some places, the keyword is used, in some not.

features/steps/thens/attributes.py
```diff
@@ -109,3 +113,18 @@ def step_impl(context, attribute, value):
...
+@then('The {field} of the {model_or_field} must be "{values}"')
+@err.handle_errors
+def step_impl(context, field, model_or_field, values):
+    values = misc.strip_split(values, strp='"', splt=' or ')
+    model_schema = context.model.schema_identifier.lower()
+    if field == "Schema Identifier":
+        s = context.model.schema_identifier
+        if not s.lower() in values:
+            yield err.IncorrectSchemaError(False, s, values)
+    elif field == "Schema" and not context.model.schema in values:
```
JTurbakiewicz, line 127: Perhaps add an else statement with RulePassed? Does it make sense for this particular rule?

test/files/ifc001/pass-ifc001-IFC4.ifc
JTurbakiewicz, line 1: Could you please also add README about test files?

JTurbakiewicz:
I like it very much. I was struggling a little bit with what IFC version to specify, now it will be way easier <3 

Btw, would it make sense to make this IFC001 rule an exception and to run it for every test file, and not only those in the ifc001 test directory?

features/SPS003_Correct-containment-of-assemblies.feature
evandroAlfieri, line 1: Do I understand correctly that when no version is made explicit, it means that the rule apply to all (currently existing) versions? This is related to @JTurbakiewicz question. Meaning:
- if we list all versions, when a new is out we need to evaluate applicability and, if needed, add version
- if we do not list all version, when a new one is out we need to evaluate applicability and, if needed, add the exception
I guess the message is: when a new version is out, the Validation Service (as any other tool implementing IFC) will have to do its homework

features/OJP001_Relative-placement-for-elements-aggregated-to-another-element.feature
Ghesselink: No, I overlooked it; it's from an older commit. 
It should be just `'Schema'`
Changed it in 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/114/commits/f479248f5606cb74e06fc2f0acdb86a7bb525512

features/GEM005_Building-shape-representation.feature
Ghesselink: I'm not sure. On the one hand, IFC 5 might take a while to come. On the other hand, that also means that we might have long forgotten about this. 

In alternative implementations, if the `Given` clause begins with an IFC instance, such as `IfcBuilding`, it is understood that the clause applies to all IFC schemas.

Should we use our decorator for this purpose? We could check in the stack frame whether the first `Given` statement is one of the statements referring to versions. If not, use a default set of versions. In our case, that would be IFC Schema `IFC2X3`, `IFC4` and `IFC4X3`. 
When IFC5 comes, or a new IFC4 version, we'll be able to make modifications at this central 'handling' location instead of spreaded accross `.feature` files.

test/files/ifc001/pass-ifc001-IFC4.ifc
Ghesselink: Added. https://github.com/buildingSMART/ifc-gherkin-rules/pull/114/commits/d06ca4b8c0da62f20a2f5fb0832eaab3666fa424

features/steps/thens/attributes.py
Ghesselink: Maybe not, but let's be consistent. 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/114/commits/26365972fa5a68a6b7012d077dcbc0ab9a1ae337

There's a 'weird' caveat in this rule. That is, all the other rules so far work with `context.instances`. But this one doesn't necessarily. I took `Given An IfcProject` to match the logic of our fail/success implementation. But maybe it's too risky to assume there is always an IfcProject.

Ghesselink:
> I like it very much. I was struggling a little bit with what IFC version to specify, now it will be way easier <3
> 
> Btw, would it make sense to make this IFC001 rule an exception and to run it for every test file, and not only those in the ifc001 test directory?

That would mean we'd have to update the test files regularly and manually in order to not fail the tests, which was one of the reasons we discussed using this approach.

features/IFC001-official-schema-identifier.feature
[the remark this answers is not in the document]
Ghesselink: That one is not merged yet, so I'm still safe .. :-) 

Jokes aside, thanks for noting. 

I was also wondering. I think it would be useful to also add a `@critical` tag. (or smth like that) We discussed that this rule should run first and, if failing, throw a big error.

features/IFC001-official-schema-identifier.feature
```diff
@@ -0,0 +1,10 @@
+@implementer-agreement
+@IFC
+Feature: CRI - Official Schema Identifier
+This rule verifies that the IFC model conforms to one of the official versions released by buildingSMART.
```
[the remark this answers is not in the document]
Ghesselink: I can't commit the suggestion directly on Github, probably because the `git diff`is outdated.
Modified it in https://github.com/buildingSMART/ifc-gherkin-rules/pull/114/commits/84d9d824a834a02a000bc4fa27d479570e353a16

features/SPS002_Correct-spatial-breakdown.feature
```diff
@@ -6,6 +6,6 @@ The possible allowed breakdown can be found in the csv file in the folder 'featu
 
   Scenario: Agreement on each IfcSpatialElement and IfcProject being aggregated as per spatial composition table.
 
-      Given A file with Schema Identifier "IFC4X3_TC1" or "IFC4X3_ADD1" or "IFC4X3" or "IFC4"
+      Given A file with Schema "IFC4" or "IFC4.3"
```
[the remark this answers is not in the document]
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/pull/114/commits/aee14fa8faff2347dcfae74f26bb3451f8e9d44e

features/SPS001_Basic-spatial-structure-for-buildings.feature
```diff
@@ -6,13 +6,13 @@ The rule verifies that there's maximum one instance of IfcSite and at least one
   Scenario: Agreement141 - Agreement on having maximum of one instance of IfcSite
 
       Given A file with Model View Definition "CoordinationView"
```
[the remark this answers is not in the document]
Ghesselink: Changed in https://github.com/buildingSMART/ifc-gherkin-rules/pull/114/commits/aee14fa8faff2347dcfae74f26bb3451f8e9d44e

> need change to python?

No, I've already implemented some 'forgiving' lines of code that allows for a rule-maker to use either `file` or `model` interchangeably. 
But let's stick using `model` in the statements.

features/SPS001_Basic-spatial-structure-for-buildings.feature
```diff
@@ -6,13 +6,13 @@ The rule verifies that there's maximum one instance of IfcSite and at least one
...
-        And A file with Schema Identifier "IFC2X3"
+        And A file with Schema "IFC2X3"
```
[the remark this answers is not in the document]
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/pull/114/commits/aee14fa8faff2347dcfae74f26bb3451f8e9d44e

features/SPS001_Basic-spatial-structure-for-buildings.feature
```diff
@@ -6,13 +6,13 @@ The rule verifies that there's maximum one instance of IfcSite and at least one
...
        Then There must be at most 1 instance(s) of IfcSite
 
   Scenario: Agreement142 - Agreement on having at least one instance of IfcBuilding as part of the spatial structure 
 
       Given A file with Model View Definition "CoordinationView"
```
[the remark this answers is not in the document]
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/pull/114/commits/aee14fa8faff2347dcfae74f26bb3451f8e9d44e

features/SPS003_Correct-containment-of-assemblies.feature
Ghesselink: > Do I understand correctly that when no version is made explicit, it means that the rule apply to all (currently existing) versions?

Yes, implicitly. The gherkin implementation just simply doesn't look at it. It's more a matter of applicability I guess.
Right now, say we run the validation service on a model with `IfcAlignment`. If we would not specifiy a version, the statement 'Given an IfcAlignment' will always be triggered. Explicitly expressing that it requires IFC version `IFC4.3` allows us to look for applicability. 

I took a note to implement 'default-schemas' with the decorator Jakub and me are working on. Meaning, when there is no schema's defined in the set of `Given` statements it will be one of the versions we listed. This way, I'd say it's good that schema versions are very explicit on one location and if there's an error regarding this (e.g. with a new version) we'd know where to update it easily.

features/IFC001-official-schema-identifier.feature
```diff
@@ -0,0 +1,10 @@
...
+Specifically, IFC2x3 TC1 (version 2.3.0.1), IFC4 ADD2 TC1 (version 4.0.2.1), or IFC4.3 ADD2 (version 4.3.2.0).
+
+  Scenario: Verifying Current Schema Identifier for IFC version
+  
+    Given An IfcProject
```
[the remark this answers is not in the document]
Ghesselink: Another option would be to leave the Given statement out. This gives practically the same output.  `'Given an IFC model' `is probably clearer for the reader.

evandroAlfieri (approved):
well done 🦾

PR #117, opened 2023-11-21
change add to given or then

Ghesselink:
As discussed in meeting 

```
Given An IfcElement
And Something else
Then Condition must be true
```

Should be changed to 

```
Given An IfcElement
Given Something else
Then Condition must be true
```

Furthermore, indentation is equal across all `.feature `files now.
No Python changes.


evandroAlfieri (approved):
💪

Issue #105, opened 2023-11-01
ALS010 - IfcShapeRepresentation with Type=Segment should have exactly 1 item


aothms:
From: https://github.com/IfcOpenShell/IfcOpenShell/pull/3945#issuecomment-1788133258

![afbeelding](https://github.com/buildingSMART/ifc-gherkin-rules/assets/1096535/6b2746a5-ce6c-46b7-9192-110f7fc57515)

The concept template language is not able to specify this.

evandroAlfieri:
ALS010?

civilx64:
Agree on ALS prefix.  Not sure about next available index.  Currently it appears that ALS008 would be next available but that would change after 006 and 007 are renamed to 001 and 002 respectively per review comments on #106.

civilx64:
@evandroAlfieri please rename from `ALSXXX` to `ALS010` per discussion this week.

civilx64:
ALS010 has been merged and will be present in the next update of the validation service.

PR #125, opened 2024-01-04
ALS010 - Alignment segment shape representation must have exactly one instance

civilx64:
Addresses #105


features/ALS010_Alignment-segment-shape-representation-has-exactly-one-item.feature
```diff
@@ -0,0 +1,23 @@
...
+
+      Given Its attribute RepresentationType
+      Then The value must be "placeholder_invalid"
+
+
+  @E00040
+  Scenario: Agreement on each IfcAlignmentSegment having correct number of representation items
+
+      Given Its attribute Items
+      Then There must be one representation item
```
aothms, line 23: I think we should remove the attribute name from the statement in order to make it more general.

```suggestion
      Then There the number of values should be 1
```

It's a bit of similar predicate to https://github.com/buildingSMART/ifc-gherkin-rules/blob/development/features/steps/thens/values.py#L31-L47 except there we count the occurrences of a certain value in the list. Here we only count the items in the list.

features/steps/thens/existance.py
```diff
@@ -22,3 +22,10 @@ def step_impl(context, inst, constraint, num, entity):
 
         if not op(len(instances_in_model), num):
             yield StepResult(expected=num, observed=len(instances_in_model))
+
+
+@validate_step('There must be one representation item')
+def step_impl(context, inst):
+
+    count = len(inst.Representation.Items)
+    yield StepResult(expected=142, observed=count)
```
aothms, line 31:
```suggestion
    yield StepResult(expected=1, observed=count)
```

But as said above, we should probably find a way to make this more generic. I'm also a bit confused actually on why you do `inst.Representation.Items` here. I would have imagined that by the time you would evaluate this statement, `inst` would already be the value of `Items` due to `Its attribute Items`.

aothms (changes requested):

features/ALS010_Alignment-segment-shape-representation-has-exactly-one-item.feature
```diff
@@ -0,0 +1,23 @@
...
+    Given A model with Schema "IFC4.3"
+    Given An IfcAlignmentSegment
+    Given Its attribute Representation
+    Given Its attribute Representations
+
+  @E00020
+  Scenario: Agreement on each IfcAlignmentSegment using correct representation - Type
+
+      Given Its attribute RepresentationType
+      Then The value must be "placeholder_invalid"
```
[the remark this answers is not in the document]
civilx64: This has been corrected.  Initially I was trying to get the rule to fail which is why I had "placeholder_invalid".

features/ALS010_Alignment-segment-shape-representation-has-exactly-one-item.feature
civilx64: This has been revised based on today's discussion.  As for making this step more general, I would prefer to leave it as-is, in the spirit of YAGNI.

features/steps/thens/existance.py
civilx64: Correct, at this point in the process `inst` is the Items attribute.  It appears that inst will always be a list of length 1 but the list object will contain one or more `Items` in a tuple.  Therefore I have revised the step implementation to check both the length of the list and and the length of the first list item.

features/steps/thens/existance.py
```diff
@@ -22,3 +24,14 @@ def step_impl(context, inst, constraint, num, entity):
 
         if not op(len(instances_in_model), num):
             yield StepResult(expected=num, observed=len(instances_in_model))
+
+
+@validate_step('There must be {num:d} representation item(s)')
+def step_impl(context, inst, num):
+    # inst is a list that always has one item but that item may contain multiple tuples
```
aothms, line 31: I don't think we can get away with it like that. The structure comes from the IFC data model. If we print the instances after every given statement and interleave these line by line we get the following structure:

```
#2288=IfcAlignmentSegment('3FHEGq6711TulOKVgmhD$$',#1,'h1',$,$,#2289,#2458,#2293),
    #2458=IfcProductDefinitionShape($,$,(#2460)),
        (#2460=IfcShapeRepresentation(#2459,'Axis','Segment',(#2450)),),
                ((#2450),),
#2295=IfcAlignmentSegment('2NHsvN8xnFnetsLAyLgoE6',#1,'h2',$,$,#2296,#2469,#2300),
    #2469=IfcProductDefinitionShape($,$,(#2470)),
        (#2470=IfcShapeRepresentation(#2459,'Axis','Segment',(#2461,#24610)),),
                ((#2461,#24610),),
#2302=IfcAlignmentSegment('1$forlm9r0XeFCaEQu3k18',#1,'h3',$,$,#2303,#2479,#2307),
    #2479=IfcProductDefinitionShape($,$,(#2480)),
        (#2480=IfcShapeRepresentation(#2459,'Axis','Segment',(#2471)),),
                ((#2471,),)
#2309=IfcAlignmentSegment('3gVJrcRzD8yPNgP0ikTfG9',#1,'h4',$,$,#2310,#2489,#2314),
    #2489=IfcProductDefinitionShape($,$,(#2490)),
        (#2490=IfcShapeRepresentation(#2459,'Axis','Segment',(#2481)),),
                ((#2481,),)
```

So `len(inst)` is actually the number of shape representations, `len(inst[0])` is the amount of items for the first shape representation.

--> So, quick and dirty way to move on would be to change it to `all(len(v) == num for v in inst)`. All shape representations should have length of `num`.

But maybe we should also think about what this means for the semantics of this rule.

Are multiple representations allowed? Are representations with a different identifier allowed to have multiple segments?

Is the rule:

a)

- All representation identifiers should be Segment
- All sequences of items should be length=1

(this allows for the case of a segment without a representation)

b)

- An alignmentseg should have a at least one Segment representation
- For the representation that has Id=Segment only a single item is allowed

(this disallows a segment without a representation - I don't think we can do that given businesslogic-only exchanges)

or c)

- If an alignmentseg has a representation, then
- it should have a at least one Segment representation
- and then representation should have a sequence of items should be length=1

This is a good example of where a simple rule gets complex due to the variability of the IFC schema.

My recommendation:

- Stick with option (a) this is closest to what's currently implemented. I don't foresee other representation Ids on segments for now (but it is a bit against the nature of IFC to forbid supplementary reps, we can move to ) The case of a segment without a representation should be allowed.
- Create some more tests:
  - Segment without representation
  - Segment with two representations

civilx64: The implementation of scenario 2 has been adjusted.  I used a for loop so that we could extract the actual number of items and report that back as "observed".  I also added more tests for the two cases listed above.

Ghesselink: For a separate PR perhaps, but now it's fresh in my head. 
We could use this (as well as similar step_impl functions in `values.py`) more with reuse in mind, so for example 
```
      Given Its attribute Items
      Then  It must be of length [{pred}]{num:d}
```
  
  pred (abbreviation for 'predicament'?) can then be one of: 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/f6d10cf03d0acc7a0a94383185402e642803e114/features/steps/utils/misc.py#L104-L117
Used in the step_impl function https://github.com/buildingSMART/ifc-gherkin-rules/blob/f6d10cf03d0acc7a0a94383185402e642803e114/features/steps/thens/existance.py#L20-L30
  
` [{pred}]` is written in brackets for behave to parse match correctly. 
 An alternative would be using a `register_type`, but I'm not sure how to do this in this case. 
 
 But, as said, it would probably be better to do this outside of this work.

evandroAlfieri (approved):
Only looked at the feature file, and that's good for me. Well done

aothms (approved):

PR #128, opened 2024-01-13
implement ALS015

civilx64:


features/steps/givens/attributes.py
```diff
@@ -71,6 +73,19 @@ def step_impl(context, attribute):
     setattr(context, 'attribute', attribute)
 
 
+@validate_step('Its final segment')
+def step_impl(context):
+    context._push()
+    context.instances = list()
+    for curves in context._stack[1]["instances"]:
```
Ghesselink, line 80: Why context._stack[1]["instances'], doesn't simply `context.instances` work?
I just tried in my debugger, and it indeed doesn't. I'd expect them to be the same. 
What kind of errors did you encounter here?

features/steps/thens/attributes.py
```diff
@@ -72,4 +77,14 @@ def step_impl(context, inst, field, file_or_model, values):
     elif field == "Schema" and not context.model.schema in values:
         s = context.model.schema
         if not s.lower() in values:
-            yield StepResult(expected=values, observed=s)
\ No newline at end of file
+            yield StepResult(expected=values, observed=s)
+
+
+@validate_step('The {length_attribute} of the final segment must be 0')
```
Ghesselink, line 83: I'd say make the '0' a function attribute, so 
`'must be {num:d}'` . Or does it have to be a float in this case?

features/ALS015_Alignment-representation-zero-length-final-segment.feature
```diff
@@ -0,0 +1,23 @@
...
+  Given An IfcAlignment
+  Given Its attribute Representation
+  Given Its attribute Representations
+  Given Its attribute Items
+  Given Its attribute Segments
+  Given Its final segment
+
+@E00020
+Scenario:
+  Then The SegmentLength of the final segment must be 0
```
Ghesselink, line 18: Would the following work? 
```
Given Its final segment 
Given its attribute SegmentLength or Length

Then The value must be equal to 0
```

Ghesselink: But maybe it's too simplistic due to the involvement of the geometric tolerance (which isn't the case with just plain integer to integer comparisons)

features/ALS015_Alignment-representation-zero-length-final-segment.feature
```diff
@@ -0,0 +1,23 @@
...
+Background: Validating final segment of alignment geometry (representation).
+  Given A model with Schema "IFC4.3"
+  Given An IfcAlignment
+  Given Its attribute Representation
+  Given Its attribute Representations
+  Given Its attribute Items
+  Given Its attribute Segments
+  Given Its final segment
+
+@E00020
```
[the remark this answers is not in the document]
civilx64: The first scenario checks length and the second checks the value of the attribute `Transition`.  Given the content and the description of the feature itself I assumed that it was ok as-is.

features/ALS015_Alignment-representation-zero-length-final-segment.feature
civilx64: I guess I could add the additional `Given` statement.  It would eliminate a few lines from the `then` step implementation.  Is that the preferred approach stylistically?

features/steps/givens/attributes.py
civilx64: I didn't encounter any errors.  This was the first way I put it together and it worked, so I moved on :smiley:.  Note L78 where context is pushed, meaning `context.instances` then becomes `None`, then turned into an empy list on L79..

features/steps/thens/attributes.py
civilx64: Yes, it has to be a float.  There's no value in making it a function attribute because of the presence of `final segment` in the statement.  Perhaps revising to `final alignment segment` will help indicate that this implementation is specific to alignments.

features/ALS015_Alignment-representation-zero-length-final-segment.feature
evandroAlfieri, line 1: @Ghesselink would your "naming convention script" raise any flag for the consistency between the file name and the feature title? Or is it case insensitive?

PR #130, opened 2024-01-14
ALB015

civilx64:
`ALB015` should be reviewed only after `ALS015` is complete and merged, given that the latter builds off of the former.  The rules are similar; just applied to different parts of the alignment model.


features/steps/validation_handling.py
```diff
@@ -222,6 +222,17 @@ def handle_nested(instance):
 def is_list_of_tuples_or_none(var):
     return isinstance(var, list) and all(item is None or isinstance(item, tuple) for item in var)
 
+def select_final_instance(instances):
+    """"
+    The 'its final' predicament seems to be a statement that can be applied to either
```
aothms, line 227:
```suggestion
    The 'its final' predicate seems to be a statement that can be applied to either 
```

features/steps/givens/attributes.py
```diff
@@ -91,27 +87,18 @@ def step_impl(context, inst, attribute, tail="single"):
...
-        # processing an ALS rule
-        for curves in context._stack[1]["instances"]:
-            for curve in curves:
-                for segments in curve:
-                    last_seg = segments[-1]
-                    context.instances.append(last_seg)
-                    setattr(context, 'applicable', True)
-    elif segment_type.upper() in business_logic_types:
+    if segment_type == "segment":
+        return [segments[-1] for curve in inst for segments in curve]
```
aothms, line 95: 1.

I don't think we should `return` and `yield` in the same function. This has a high potential of confusing. In python 2.x this wasn't even allowed.

2.

It seems to me that 

- @validate_step('Its final {segment_type}')
- @validate_step('Its final segment')

are redundant.

features/steps/validation_handling.py
```diff
@@ -241,7 +253,8 @@ def inner(context, **kwargs):
                     context.instances =  flatten_list_of_lists([fn(context, inst=inst, **kwargs) for inst in flatten_list_of_lists(context.instances)])
                 else:
                     context.instances = list(map(attrgetter('inst'), filter(lambda res: res.severity == OutcomeSeverity.PASS, itertools.chain.from_iterable(fn(context, inst=inst, **kwargs) for inst in context.instances))))
-                pass
+                    if 'its final' in context.step.name.lower() and 'segment_type' in fn_args:
```
aothms, line 256: I'm worried about this. I really prefer not to have a tight coupling between the invocation in the decorator and the semantics of individual statements. This is really hard to maintain and understand.

The addition of `'its final'` should be handled in the step implementation function.

civilx64: Agree.
This has been resolved elsewhere and `its final` is no longer in the decorator.

features/steps/givens/attributes.py
civilx64: Agree. The former was removed in a different edit and is no longer present.

features/steps/validation_handling.py
civilx64: This line is no longer present after re-syncing with `development`.  However a search across files located it in another spot, which I have addressed in the most recent push.

civilx64:
Spoke too soon; tests not passing 😠.  Will resolve before pushing the update.

features/steps/givens/attributes.py
civilx64: As it turned out the best solution was two separate implementation functions: one for ALS015 and a separate one for ALB015.

civilx64:
@evandroAlfieri there were some recent tweaks to the ALB015 feature file, in case you would like to review.

features/ALB015_Alignment_business_logic_zero_length_final_segment.feature
```diff
@@ -0,0 +1,35 @@
+@ALB
+@version1
+Feature: ALB015 - Alignment Business Logic Zero-Length Final Segment
+  The rule verifies that each aspect (horizontal, vertical, cant) of the alignment business logic
```
evandroAlfieri, line 4:
```suggestion
  The rule verifies that each layout (horizontal, vertical, cant) of the alignment business logic
```
for consistency with IFC documentation, where they're always called layouts

features/ALB015_Alignment_business_logic_zero_length_final_segment.feature
evandroAlfieri, line 1: Please make the file name and the feature title consistent (case sensitive). @Ghesselink I probably already asked this, shouldn't your script pick up this?

features/ALS015_Alignment-representation-zero-length-final-segment.feature
evandroAlfieri, line 1: Minor. The description says _segment with length_ and _Transition_ (capital T), while the attributes are _SegmentLength_ and _Transition_

For consistency, please chose one of the following for the description:

- ... a discontinuous segment with length = 0
- ... a segment with Transition = DISCONTINUOUS and SegmentLength = 0

Probably the first reads better.

evandroAlfieri (approved):
Great work. Minor tweaks.

features/ALB015_Alignment_business_logic_zero_length_final_segment.feature
Ghesselink: > shouldn't your script pick up this?

The script is not coupled yet. It was with the old repository, but we had many fundamental changes about how data is passed around. I've made some updates to be up-to-date with the development branch https://github.com/buildingSMART/ifc-gherkin-rules/pull/135

I've run the script on this repository. What gets triggered is the `Zero-length` with `'-'` as a separator. This raises an error due to not all separators being similar.

features/ALS015_Alignment-representation-zero-length-final-segment.feature
civilx64: Agree.  I have revised to use the first option.

features/ALB015_Alignment_business_logic_zero_length_final_segment.feature
civilx64: My apologies.  I have revised for case sensitivity and removed the `'-'` character pert Geert's feedback.  Thanks for the tip on "layout"... I think I invented "aspect" somewhere along the line.

PR #138, opened 2024-01-29
ALA001

civilx64:
Implement ALA001, the first rule in a new functional part for checking alignment consistency
between the representation (geometry) and business logic.

The implementation of this rule adds a new module `ifc43x_alignment_validation` that is used
primarily to load alignment data from an `IfcAlignment` entity,
then parse all subcomponents such as business logic layouts and shape representations in an
object-oriented manner.


features/ALA001_Alignment-overall-agreement-of-business-logic-and-geometry.feature
```diff
@@ -0,0 +1,38 @@
+@ALA
+@version1
+Feature: ALA001 - Alignment overall agreement of business logic and geometry
+  The rule verifies that when an Alignment has both business logic and geometry (representation)
+  that the representation entity type corresponds to the aspects present in the business logic.
```
evandroAlfieri, line 5:
```suggestion
  The rule verifies that when an Alignment has both business logic and geometry (representation), the representation entity type must correspond to the layouts present in the business logic.
```

evandroAlfieri (changes requested):
Very good!

evandroAlfieri (approved):

PR #140, opened 2024-01-29
IBP001 - Stair decomposition

Ghesselink:
Code found on my local machine that could serve as a best-practice


features/resources/spatial_DecompositionTable.csv
```diff
@@ -43,4 +43,8 @@ IfcSite,IfcRailway
 IfcSite,IfcRoad
 IfcSite,IfcExternalSpatialElement
 IfcSite,IfcSpace
-IfcSpace,IfcSpace
\ No newline at end of file
+IfcSpace,IfcSpace
```
aothms, line 46: I think this should be a separate table. IfcStair is not a spatial element. Or the table can be renamed to simply decomposition table. But people are going to complain if we don't have consistent vocabulary on these kind of things.

test/files/ibp001/generate.py
```diff
@@ -0,0 +1,36 @@
+import os
+import ifcopenshell
+import ifcopenshell.template
+
+for validity, elem in [
+    ("pass", 'IfcRailing'),
+    ("pass", 'IfcSlab'),
+    ("pass", 'IfcStairFlight'),
+    ("pass", None),
+    ("fail", "IfcBeam"),
```
aothms, line 10: I would say a Beam should/could be allowed, same as an IfcMember.

aothms (changes requested):

evandroAlfieri:
> I'm not immediately falling in love with the rule code ASS101, but maybe I'm being too conservative.

I expect @Ghesselink naming convention script to flag it. And not because it's conservative too...but because the feature title, i.e. ASS101, is inconsistent with the file name, i.e. IBP001 😄

Ghesselink:
I changed it too IPB (Industry Best Practice) for the same reason 😛

test/files/ibp001/generate.py
Ghesselink: And what about IfcRoof and IfcWall, for instance? Should they be allowed? 

The documentation I found about it only mention flights, railings and slabs. e.g.; https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcStair.htm

features/resources/spatial_DecompositionTable.csv
Ghesselink: Moved it to a separate table for now

evandroAlfieri (approved):
Nice. Let's merge!

PR #142, opened 2024-02-01
fix feature_filter hack for for prod

Ghesselink:
Msg from Slack from Raphael

> is it possible to change this block by the way, it’s a bit hacky and throwing errors in std.out (‘list index out of range’) which I have to explicitly ignore now https://github.com/buildingSMART/ifc-gherkin-rules/blob/900b2e93fb872e7daadddc3a4786d79be76a476d/main.py#L61


aothms:
Ref: https://github.com/buildingSMART/ifc-gherkin-rules/pull/145/files#diff-b10564ab7d2c520cdd0243874879fb0a782862c3c902ab535faabe57d5a505e1L65-R71

PR #145, opened 2024-02-13
Attempts at integration into VS

aothms:
Hi team, I've been working my way around some issues I encountered:

- Only do feature matching when in production mode. In pre-06 our file names where just unique codes so they would never trigger the matching behaviour. In the new platform we retain the basename of the file.
- Don't capture output in production mode (I think this is what @rw-bsi also encountered, it silences all things like import errors)
- And then a whole lot of ugly import try catching that was needed on my end for the databases to be in sync. Currently we're running as a subprocess but I think the future we might want to use the actual celery worker interpreter to do the actual behave evaluation as all the python interpreter startups with heavy module imports are deadslow) in that case it becomes even more important the sys.module paths line up properly.


Ghesselink:
I've worked around why the tests are failing. I think it's because of this line 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/4adabb8a714c9dd3888741cf559fc00f5aeec234/validation_results.py#L17

 `Path(ifc_validation_models.__file__) `is referring to a file, while current_script_dir / 'ifc_validation_models' is referring to folder or python package (i.e. the submodule). If we stick with the Path module, a solution would be to compare with
  ```python
  Path(ifc_validation_models.__file__).parent
  ```
   instead . 
  
  Then it will fail because of the next line 
  ```python
  ifc_validation_models.apps.IfcValidationModelsConfig.name = 'ifc_validation_models'
  ```
This seems because the apps folder is not imported in the __init__.py file of `ifc_validation_models` package. 
Locally, the tests pass when I import this specifically
https://github.com/buildingSMART/ifc-validation-data-model/blob/main/__init__.py
```python
from . import apps 
```

Does this make sense? Since I don't have a local version of the validation service running yet, I cannot see if I accidentally break something with this solution.

validation_results.py
```diff
@@ -5,16 +5,29 @@
...
+# @todo importing from random directories is a security hazard, this should be properly passed as a configuration param
+current_script_dir = Path(os.path.dirname(os.path.abspath(__file__)))
+sys.path.insert(0, str(current_script_dir.parent.parent.parent.parent))
+
+try:
+    import apps.ifc_validation_models as ifc_validation_models
+except:
+    import ifc_validation_models
+
+if Path(ifc_validation_models.__file__) == current_script_dir / 'ifc_validation_models':
```
Ghesselink, line 17:
```suggestion
if Path(ifc_validation_models.__file__).parent == current_script_dir / 'ifc_validation_models':
```

Ghesselink (changes requested):

aothms:
Thanks for looking into this. Python modules can be really funky (and annoying).

> This seems because the apps folder is not imported in the **init**.py file of `ifc_validation_models` package. Locally, the tests pass when I import this specifically https://github.com/buildingSMART/ifc-validation-data-model/blob/main/__init__.py
> 
> ```python
> from . import apps 
> ```

Can we figure out a way to make this work with only changes on this side of the repo? Maybe:

```python

try:
    import apps.ifc_validation_models as ifc_validation_models
    import apps.ifc_validation_models.apps as ifc_validation_models_apps
except:
    import ifc_validation_models
    import ifc_validation_models.apps as ifc_validation_models_apps

...

ifc_validation_models_apps
```

Or is that even uglier? I'm also fine with your proposal GH!

Ghesselink:
> Thanks for looking into this. Python modules can be really funky (and annoying).
> 
> > This seems because the apps folder is not imported in the **init**.py file of `ifc_validation_models` package. Locally, the tests pass when I import this specifically https://github.com/buildingSMART/ifc-validation-data-model/blob/main/__init__.py
> > ```python
> > from . import apps 
> > ```
> 
> Can we figure out a way to make this work with only changes on this side of the repo? Maybe:
> 
> ```python
> try:
>     import apps.ifc_validation_models as ifc_validation_models
>     import apps.ifc_validation_models.apps as ifc_validation_models_apps
> except:
>     import ifc_validation_models
>     import ifc_validation_models.apps as ifc_validation_models_apps
> 
> ...
> 
> ifc_validation_models_apps
> ```
> 
> Or is that even uglier? I'm also fine with your proposal GH!

The line 'from . import apps' was for a slightly different problem. 
It's a bit confusing, but I think there are two alike but distinct 'apps' 
- The `apps` that refers to the folder within the validation repository. https://github.com/buildingSMART/validate/tree/development-0.6/backend/apps
- The `apps` that refers to apps.py within the datamodel. [apps.py](https://github.com/buildingSMART/ifc-validation-data-model/blob/2b93ab4196b99197d0b45c8bd5ddf2cd4f29bc6d/apps.py)

This solution is for the latter. But anyhow, I think we can solve it with only changes on this side of the repo. Simply by 
`from ifc_validation_models import apps`

So 

```python
if Path(ifc_validation_models.__file__).parent == current_script_dir / 'ifc_validation_models':
    # we are using our own submodule
    try:
        ifc_validation_models.apps.IfcValidationModelsConfig.name = 'ifc_validation_models'
    except:
        from ifc_validation_models import apps
        apps.IfcValidationModelsConfig.name = 'ifc_validation_models'
    os.environ['DJANGO_SETTINGS_MODULE'] = 'ifc_validation_models.independent_worker_settings'
else:
    os.environ['DJANGO_SETTINGS_MODULE'] = 'apps.ifc_validation_models.dependent_worker_settings'
```

This also passes the tests and we don't have to touch the datamodel submodule

aothms:
>  It's a bit confusing, but I think there are two alike but distinct 'apps'

yes that's confusing, but I was aware of that.

I think this looks good as is.

rw-bsi:
Is the file `'apps.ifc_validation_models.dependent_worker_settings'` available somewhere? Currently the integration with Validation Service - development-0.6 branch is broken

PR #146, opened 2024-02-14
Implements ALA002

civilx64:
- feature file
- step implementation
- test data


evandroAlfieri (approved):
Feature file looks good

features/steps/thens/alignment.py
```diff
@@ -51,6 +51,23 @@ def is_3d(entity: ifcopenshell.entity_instance) -> bool:
...
+        logic_count = None
+    try:
+        rep_count = len(representation.segments)
+    except AttributeError:
+        rep_count = None
+
+    return logic_count, rep_count
+
+
 @gherkin_ifc.step('A representation by {ifc_rep_criteria} requires the {ifc_layout_criteria} in the business logic')
```
Ghesselink, line 71: Can we split the `ifc_layout_criteria` up?  
So that we get 
`' A representation by {ifc_rep_criteria} requires the {existence:absence_or_presence} of {entities} in the business logic'`

To allow for only presence/absence as string input (on top in the file, similar as in values.py)
```
from parse_type import TypeBuilder
register_type(value_or_type=TypeBuilder.make_enum(dict(map(lambda x: (x, x), ("presence", "absence"))))) 
```

features/steps/thens/alignment.py
```diff
@@ -63,8 +80,8 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
                                                 severity=OutcomeSeverity.ERROR)
 
             case ("IfcGradientCurve", "presence of IfcAlignmentVertical"):
```
Ghesselink, line 82: This will then match the following
```python
match (ifc_rep_criteria, existance, entities):
  case ("IfcGradientCurve", "presence", "IfcAlignmentVertical")
```

features/steps/thens/alignment.py
```diff
@@ -63,8 +80,8 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
...
-                if align.gradient_curves is not None:
-                    if align.verticals is None:
+                if align.gradient_curve is not None:
+                    if align.vertical is None:
                         yield ValidationOutcome(inst=inst, expected=ifc_layout_criteria, observed=None,
```
Ghesselink, line 85: And depending on whether `existence` equals `presence` or `absence` we'll say:

In case of presence, we expected {'entity':'IfcGradientCurve'} 
With the new function that parses the output into json that Raphael can work with, that means simply returning `expected=entities` and `observed=None`

In the case of absence, it'll be vice versa. We expected None, but observed the presence of an entity (or one of the entities)

I guess this will be good enough for the first version.

features/steps/thens/alignment.py
```diff
@@ -93,7 +110,7 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
                         if (item.is_a().upper() == "IFCCOMPOSITECURVE") and (
```
Ghesselink, line 110:
```python
case ("IfcCompositeCurve as Axis", "absence of IfcAlignmentVertical and IfcAlignmentCant"):
``` 
In line 106 (I think it's still part of ALA001). 

`observed = "', '".join(entities.split(" and "))
`

This will return the string 
'expected=None'
'observed={'OneOf' : ['IfcAlignmentVertical', 'IfcAlignmentCant']}

Which I think would be clear

features/ALA002_Alignment-same-number-of-segments-in-business-logic-and-geometry.feature
Ghesselink, line 1: I'd say we could write the Then statements shorter, as in 

`The alignment's horizontal layout must have the same number of segments as the shape representation
`
to 

`The layout must have the same number of segments as the shape representation`

That is, we already have the information about the layout being horizontal, vertical or cant; this is the type of the instance that is returned by the Given statement.

features/steps/thens/alignment.py
```diff
@@ -110,3 +127,43 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
...
+@gherkin_ifc.step(
+    'The alignment\'s {layout_type} layout must have the same number of segments as the shape representation')
+def step_impl(context, inst, layout_type):
+    for layout_ent in context.instances:
+        for rel in layout_ent.Nests:
+            ent = rel.RelatingObject
+            if ent.is_a() == "IfcAlignment":
+                align = ifc43x_alignment_validation.entities.Alignment().from_entity(ent)
+
+                match layout_type:
```
Ghesselink, line 141:
```suggestion
                match inst.is_a()
```

features/steps/thens/alignment.py
```diff
@@ -110,3 +127,43 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
...
+                    case "horizontal":
+                        logic_count, rep_count = count_segments(
+                            logic=align.horizontal,
+                            representation=align.composite_curve,
+                        )
+                    case "vertical":
```
Ghesselink, line 147:
```suggestion
                    case "IfcAlignmentVertical":
```

features/steps/thens/alignment.py
```diff
@@ -110,3 +127,43 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
...
+    'The alignment\'s {layout_type} layout must have the same number of segments as the shape representation')
+def step_impl(context, inst, layout_type):
+    for layout_ent in context.instances:
+        for rel in layout_ent.Nests:
+            ent = rel.RelatingObject
+            if ent.is_a() == "IfcAlignment":
+                align = ifc43x_alignment_validation.entities.Alignment().from_entity(ent)
+
+                match layout_type:
+                    case "horizontal":
```
Ghesselink, line 142:
```suggestion
                    case "IfcAlignmentHorizontal":
```

features/steps/thens/alignment.py
```diff
@@ -110,3 +127,43 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
...
+                        logic_count, rep_count = count_segments(
+                            logic=align.horizontal,
+                            representation=align.composite_curve,
+                        )
+                    case "vertical":
+                        logic_count, rep_count = count_segments(
+                            logic=align.vertical,
+                            representation=align.gradient_curve,
+                        )
+                    case "cant":
```
Ghesselink, line 152:
```suggestion
                    case "IfcAlignmentCant":
```

features/steps/thens/alignment.py
```diff
@@ -110,3 +127,43 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
...
+
+                if logic_count != rep_count:
+                    # account for alignment that has business logic only
+                    # or contains representation only
+                    logic_only = (logic_count is not None) and (rep_count is None)
+                    rep_only = (logic_count is None) and (rep_count is not None)
+                    if not (logic_only or rep_only):
+                        yield ValidationOutcome(inst=inst, expected="same count of segments",
+                                                observed="differing count of segments",
+                                                severity=OutcomeSeverity.ERROR)
```
Ghesselink, line 169: I'd say just add `logic_count` and `rep_count` instead of the string that is currently parsed to `expected` and `observed`
Additionally with the feature file 'same count' it should be possible for the end-user to derive what the error is. 

Or what do you think? I'm not completely sure

features/steps/thens/alignment.py
```diff
@@ -110,3 +127,43 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
                                 yield ValidationOutcome(inst=inst, expected=ifc_layout_criteria,
                                                         observed="IfcAlignmentCant",
                                                         severity=OutcomeSeverity.ERROR)
+
+
+@gherkin_ifc.step(
+    'The alignment\'s {layout_type} layout must have the same number of segments as the shape representation')
```
Ghesselink, line 133:
```suggestion
@gherkin_ifc.step('The layout must have the same number of segments as the shape representation')
```

Perhaps we can write this even simpler

We have the layout type also implicitly as information, i.e. `inst.is_a()`

features/steps/thens/alignment.py
```diff
@@ -110,3 +127,43 @@ def step_impl(context, inst, ifc_rep_criteria, ifc_layout_criteria):
...
+def step_impl(context, inst, layout_type):
```
Ghesselink, line 134:
```suggestion
def step_impl(context, inst):
```
and remove `layout_type`

Ghesselink (changes requested):
Looks very good :) 

I have some suggested some slight changes that, I think, will make it easier to work with in the future and return a bit clearer json for Raphael and, eventually, the end user. 

What do you think?

features/steps/thens/alignment.py
civilx64: I updated `observed` to report the number of segments found in the business logic and in the shape representation.  Example
```
Observed value: {"value": "10 segments in business logic and 8 segments in representation"}
```

features/steps/thens/alignment.py
civilx64: Agree!  I implemented this per your suggested change.

civilx64:
Thanks for the review!  I have incorporated all of your suggestions and also synced this branch with `development` before they get a chance to diverge. 😃

PR #147, opened 2024-02-19
Outcome unification

aothms:
Some notes:

- I ended up using most of the `expected` columns. Because I felt they duplicated too much of the step definition. But on the other hand I don't know how obvious the failing step is to the end user. But on the other hand I don't know how expected=1 would really help understanding which step is failing in case of many steps. I think maybe we come to the conclusion that we do need the failing step in the database and maybe the simplest way is to do that generically just the full step text as we get it from behave in the decorator.
- For observed I applied two changes in direction as well: (a) in case of counts, instead of listing the count as observed I referened the list of instances (b) in case of partial results "at least two items of type X" we test the length of [i for i in values if i is X] but I referenced the full list in observed, not just those of value X, because I think that is closer to what is actually observed.
- I removed one case of explicitly yielding OutcomeSeverity.PASSED. I think that should be handled automatically now
- I no longer selectively yield outcomes (like only the first). I think with the database backend it should be efficient enough to capture all?

### DIFF
--- a/features/steps/thens/geometry.py
+++ b/features/steps/thens/geometry.py
@@ -13,16 +13,8 @@ def step_impl(context, inst, clause):
     entity_contexts = ifc.recurrently_get_entity_attr(context, inst, 'IfcRepresentation', 'ContextOfItems')
     precision = ifc.get_precision_from_contexts(entity_contexts)
     points_coordinates = geometry.get_points(inst)
-    comparison_nr = 1
-    duplicates = set()
-    for i in itertools.combinations(points_coordinates, 2):
-        if math.dist(i[0], i[1]) < precision:
-            if clause == 'including' or (clause == 'excluding' and comparison_nr != len(points_coordinates) - 1):
-                # combinations() produces tuples in a sorted order, first and last item is compared with items 0 and n-1
-                duplicates.add(i)
-                if len(duplicates) > 2:  # limit nr of reported duplicate points to 3 for error readability
-                    break
-        comparison_nr += 1
-    if duplicates:
-        yield ValidationOutcome(inst=inst, expected="No duplicates", observed="Duplicates", severity=OutcomeSeverity.ERROR)
-
+    for i, j in itertools.combinations(range(len(points_coordinates)), 2):
+        # combinations() produces tuples in a sorted order, first and last item is compared with items 0 and n-1
+        if clause == 'including' or (clause == 'excluding' and (i, j) == (0, len(points_coordinates) - 1)):
+            if math.dist(points_coordinates[i], points_coordinates[j]) < precision:
+                yield ValidationOutcome(inst=inst, observed=(points_coordinates[i], points_coordinates[j]), severity=OutcomeSeverity.ERROR)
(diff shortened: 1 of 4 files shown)


features/steps/thens/geometry.py
```diff
@@ -13,16 +13,8 @@ def step_impl(context, inst, clause):
...
-                duplicates.add(i)
-                if len(duplicates) > 2:  # limit nr of reported duplicate points to 3 for error readability
-                    break
-        comparison_nr += 1
-    if duplicates:
-        yield ValidationOutcome(inst=inst, expected="No duplicates", observed="Duplicates", severity=OutcomeSeverity.ERROR)
-
+    for i, j in itertools.combinations(range(len(points_coordinates)), 2):
+        # combinations() produces tuples in a sorted order, first and last item is compared with items 0 and n-1
+        if clause == 'including' or (clause == 'excluding' and (i, j) == (0, len(points_coordinates) - 1)):
```
Ghesselink, line 18:
```suggestion
        if clause == 'including' or (clause == 'excluding' and (i, j) =! (0, len(points_coordinates) - 1)):
```
Shouldn't this be `!= `instead?
Changing this causes the tests to pass. If the cause is excluding and the point is equal to the last point, we'd like to return False so there is no comparison.

features/steps/thens/nesting.py
```diff
@@ -36,10 +37,7 @@ def step_impl(context, inst, other_entities):
     nested_entities = [i for rel in inst.IsNestedBy for i in rel.RelatedObjects]
     nested_entity_types = set(i.is_a() for i in nested_entities)
     if not nested_entity_types <= allowed_entity_types:
-        yield ValidationOutcome(inst=inst, expected=str(allowed_entity_types), observed=allowed_entity_types - nested_entity_types, severity=OutcomeSeverity.ERROR)
-    else:
-        yield ValidationOutcome(inst=inst, severity=OutcomeSeverity.PASSED)
```
Ghesselink, line 41: Yep, this is handled automatically. Thanks for noting

Ghesselink (changes requested):
The CI/CD fails due to, I think, a small error. 
Locally, running pytest with this change the tests pass

>  I think maybe we come to the conclusion that we do need the failing step in the database and maybe the simplest way is to do that generically just the full step text as we get it from behave in the decorator

That sounds reasonable. Implemented this for now [Return last step name if expected/observed is left empty](https://github.com/buildingSMART/ifc-gherkin-rules/pull/147/commits/ff241ec04df47cb052022d0124b342ff1021e017)
Alternatively, in the future and in case if we ended up with too many strings in the database, we could just display the scenario number and then step number. This way, we can automatically extract the related 'Then' step from the .feature file that is linked to the validating rule (I'd guess simply by text-base extraction of the .feature file).

aothms:
> This way, we can automatically extract the related 'Then' step from the .feature file that is linked to the validating rule (I'd guess simply by text-base extraction of the .feature file).

Yes. The thing that still bothers me about this is the manual versioning. So if we forget updating the version number of a feature, but insert a step somewhere, everything will be misaligned. Do you think we can have a CI/CD script that rejects updates to feature without bumping the version number?

Issue #97, opened 2023-10-12
Step catalog


aothms:
It appears behave already has something like a step catalog. Who wants to experiment with that?

https://github.com/behave/behave/blob/13893e30eeb8fd5fc4ebeef5763dfe781b6f12a1/behave/formatter/steps.py#L272-L280

PR #148, opened 2024-02-22
Check version bump

Ghesselink:
In case something is modified in a .feature file, the version number needs to be bumped. This scripts checks that automatically. In github CI/CD, it only gets triggered on pull requests and not on push. This will prevent confusing situations, for instance when a .feature file is updated multiple times within a specific branch that is used for a pull request.

Specifically, the number referenced to in the @version tag. For example in https://github.com/buildingSMART/ifc-gherkin-rules/blob/874bb0843c4bbdae0180b9c9430fd47e358082e0/features/ALA002_Alignment-same-number-of-segments-in-business-logic-and-geometry.feature#L2

Related to https://github.com/buildingSMART/ifc-gherkin-rules/pull/147

As well as related to the issue https://github.com/buildingSMART/ifc-gherkin-rules/issues/97

The purpose is to setup a rule catalog. So that we can output the correct `Then` step of the validation outcome to the end-user without unnecessarily populating the database.  However, for this we'll have to ensure we don't output an outdated version of the .feature file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,16 @@ jobs:
         flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Setup version bump env vars
+      id: envvars
+      run: |
+        echo "TARGET_BRANCH=$TARGET_BRANCH" >> $GITHUB_OUTPUT
+        echo "PULL_REQUEST=$PULL_REQUEST" >> $GITHUB_OUTPUT
+      env:
+        TARGET_BRANCH: ${{ github.event_name == 'pull_request' && github.base_ref || 'development' }}
+        PULL_REQUEST: ${{ github.event_name == 'pull_request' && 'True' || 'False' }}
+
     - name: Test with pytest
       run: |
+        pytest -sv --target_branch=${{ steps.envvars.outputs.TARGET_BRANCH }} --pull_request=${{ steps.envvars.outputs.PULL_REQUEST }}
         pytest -sv
--- a/features/rule_creation_protocol/version_check.py
+++ b/features/rule_creation_protocol/version_check.py
@@ -1,0 +1,29 @@
+import subprocess
+import re
+from .errors import ProtocolError
+
+def get_changed_feature_files(target_branch):
+    changed_files = subprocess.getoutput(f"git diff --name-only {target_branch}").split("\n")
+    feature_files = [f for f in changed_files if f.endswith(".feature")]
+    return feature_files
+
+def check_version_bump(file):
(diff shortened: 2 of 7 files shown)


features/rule_creation_protocol/version_check.py
```diff
@@ -0,0 +1,29 @@
+import subprocess
+import re
+from .errors import ProtocolError
+
+def get_changed_feature_files(target_branch):
+    changed_files = subprocess.getoutput(f"git diff --name-only {target_branch}").split("\n")
```
aothms, line 6: Please change these to `subprocess.run()` which is the "modern" way to do this, operating on a list. So that you don't need to worry about shell escaping. If `{file}` (next fn) contains a space, it will break. But if every arg is a separate str in a list passed to subprocess.run() it gets escaped automatically.

features/rule_creation_protocol/version_check.py
```diff
@@ -0,0 +1,29 @@
...
+    # Compare the file from the current branch to the target_branch
+    diff_output = subprocess.getoutput(f"git diff {file}")
+    versions = re.findall(r"@version(\d+)", diff_output)
+    versions = [int(v) for v in versions]
+    if not versions:
+        raise ProtocolError(
+            value = None,
+            message = "When changing a feature file, the version number must be bumped"
+        )
+    return versions[1] > versions[0]  # Check version bump
```
aothms, line 20: Ah you're using the diff output. I can't think of a way where it would break. I have used https://pypi.org/project/unidiff/ in the past to more properly parse the diff output, but wouldn't necessarily recommend. Also had some issues.

I would add a check though about `if len(versions) > 2`. I guess that could happen if you add multiple version tags? Or is that already checked elsewhere. Minor thing.

I wonder if we should check `return versions[1]  == versions[0] + 1`?

.github/workflows/ci.yml
```diff
@@ -38,6 +38,16 @@ jobs:
         flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Setup version bump env vars
```
aothms, line 41: I think this is not necesary. There should be a `GITHUB_BASE_REF` environment variable. I would skip most of this. And just check `os.environ.get('GITHUB_BASE_REF')` in `run()`.

https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

aothms (changes requested):

civilx64:
> In case something is modified in a .feature file, the version number needs to be bumped. 

[Minor changes such as fixing typos or re-wording the description do not increment the version](https://github.com/buildingSMART/ifc-gherkin-rules/blob/development/features/README.md?plain=1#L72)

It appears that this checking script will need some additional nuance.

Ghesselink:
PR is deprecated; all further updates are in https://github.com/buildingSMART/ifc-gherkin-rules/pull/170

PR #150, opened 2024-02-27
Minor fixes

aothms:
- I got various error messages about NoneType not iterable in after_feature. Could be a symptom of something else, but this is an easy fix
- `@disabled` should always be excluded because it raises a not implemented error now


aothms:
Add two commits. 

- https://github.com/buildingSMART/ifc-gherkin-rules/pull/150/commits/2560d788c2d9beadea84b6b1f4936a45b9eccff8 is probably an important one. Relying on after_feature() effectively renders all but the last scenario useless, `after_scenario()` is what we should probably use. This is what causes e.g GEM111 files not to be reported as failing.
- Please double check, but I think the indentation was off here which may have caused some of the duplicate outcomes we were seeing.

rw-bsi (approved):

PR #151, opened 2024-02-28
Addresses VAL-326

civilx64:
update documentation for the structure of a feature file

### DIFF
--- a/features/README.md
+++ b/features/README.md
@@ -1,5 +1,6 @@
 # How to create a rule
 Follow these steps to add a new rule to the Validation Service
+
 | n. | Step                                                                                     | Responsible                 |
 |----|------------------------------------------------------------------------------------------|-----------------------------|
 | 1  | Create a new branch in the bSI ifc-gherkin-rules repository                              | bSI Validation Service team |
@@ -14,7 +15,7 @@ Follow these steps to add a new rule to the Validation Service
 
 In the buildingSMART [GitHub repository containing all rules](https://github.com/buildingSMART/ifc-gherkin-rules), create the branch that will be used to develop the new rule.
 
-- Give the branch a meaningful name
+- Name the branch with the name of the new rule. Example: `GEM900` for a new rule in the geometry functional part
 - Add 1 rule per branch, to facilitate review (1 rule = 1 `.feature` file)
 
 ## 2. Rule development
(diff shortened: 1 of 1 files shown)


evandroAlfieri (approved):
very good

features/README.md
```diff
@@ -59,7 +60,42 @@ SPS001_Basic-spatial-structure-for-buildings.feature
 
 #### Mandatory content
 `.feature` files:
-- must start with a tag to the functional part. See [Functional parts](./Functional-parts.md)
+- must include one and only one of these three tags to classify the validation category:
+    - `@implementer-agreement`
```
aothms, line 64:
```suggestion
- must include one and only one of these tags to classify the validation category:
    - `@critical`
    - `@implementer-agreement`
```

Right?

features/README.md
```diff
@@ -284,12 +333,15 @@ When creating a new step, think about parametrisation and optimisation of the st
 Before creating a new step, check if something similar already exist.
 Try to reuse existing steps.
 
-#### Do not use "when" keyword
+#### Do not use "when" or "And" keywords
 The "when" keyword must not be used.
-Allowed keywords are: `Given`, `Then`, `And`, `*`
+The "And" keyword must not be used.
```
aothms, line 338: > The "And" keyword must not be used.

Do we currently comply with this?

aothms: @Ghesselink has been working on categorizing the various given statements we have:

- global file-based such as `given schema ...`
- intial scope selection `Given an IfcWall`
- filtering scope `Given attribute Name = ...`
- navigating scope `Given its attribute XXX`  `Given an relation IfcX to IfcY`

I think it would be very nice to add that here, but that can of course be a subsequent PR.

features/README.md
evandroAlfieri: Is this used only for the version prerequisite?

Ghesselink: > Is this used only for the version prerequisite?

So far, yes. See also
https://github.com/buildingSMART/ifc-gherkin-rules/blob/development/features/IFC101_Only-official-ifc-versions-allowed.feature

features/README.md
Ghesselink: > > The "And" keyword must not be used.
> 
> Do we currently comply with this?

Yes, I refactored all the statements a while ago. Cannot find back exactly when. 
We're not enforcing it at the moment with the conventions-check.

Ghesselink: > I think it would be very nice to add that here, but that can of course be a subsequent PR.

Think I'll use our work at the moment as a sort of R&D for these improvements and add everything to the VS after the 11th of March.

PR #152, opened 2024-02-29
Delete ALB001

evandroAlfieri:


aothms:
The history of this one is a bit funky. I'll commit it directly https://github.com/buildingSMART/ifc-gherkin-rules/commit/a1c34f42f27b68b2a86af5e5d4bf02e75d143176

afbeelding

PR #154, opened 2024-03-02
add tags to alignment feature files

civilx64:
- this will result in newer alignment rules being categorized correctly as implementer agreements
- this also corrects minor typos (extra spaces and blank lines) for better alignment with the updated README regarding format of `.feature` files
- ref [VAL-336](https://bsi-technicalservices.atlassian.net/jira/software/projects/VAL/boards/1?selectedIssue=VAL-336&text=336)


evandroAlfieri (approved):
Excellent

PR #156, opened 2024-03-11
Fix ALA002 for correct count of segment with HELMERT transition type

civilx64:
Rule ALA002 was originally developed with the assumption that each segment in the alignment business logic is represented by a single segment in the geometry representation. However, a single segment of type `HELMERTCURVE` requires two segments in the representation.

Revisions include:

- rephrasing of the rule name, description, and steps in the `.feature` file
- changed to step implementation to iterate over logic segments and create a count of expected representation segments

Other miscellaneous / housekeeping items included in this commit:

- fix enumeration name and `append` syntax in `environment.py`.
- upgrade data model subrepository to latest commit
- bump python version in gihtub actions to 3.11 for consistency with production environment Dockerfile


features/steps/thens/alignment.py
```diff
@@ -61,15 +61,20 @@ def count_segments(logic, representation):
...
     except AttributeError:
-        logic_count = None
+        expected_count = None
     try:
         rep_count = len(representation.segments)
     except AttributeError:
         rep_count = None
 
-    return logic_count, rep_count
+    return expected_count, rep_count
```
Ghesselink, line 77: I'd say be consistent with using `expected_count` instead of `logic_count`.

I don't seem to be able to pinpoint the exact line numbers in pull requests. But for example in line 164/169 we still say 
`if logic_count != rep_count:`

This should then be 
`if expected_count != rep_count`

test/files/ala002/pass-ala002-helmert_curve.ifc
```diff
@@ -0,0 +1,156 @@
...
+#15 = IFCRAILWAY('1FNFy9AJeHwuVmwDZHIYIu', #3, 'optional Railway Name', 'optional Railway Description', $, #14, $, 'Longname for Railway track geometry (alignment)', $, $);
+#16 = IFCDIRECTION((0., 1.));
+#17 = IFCGEOMETRICREPRESENTATIONCONTEXT($, 'MODEL', 3, 1.E-5, #13, #16);
+#18 = IFCRELAGGREGATES('1FNFyAAJeHwu6TwDZHIYIu', #3, 'Documentation Container', 'Container for IfcProject, collects relevant products', #1, (#15));
+#19 = IFCRELCONTAINEDINSPATIALSTRUCTURE('1FNFyBAJeHwvmywDZHIYIu', #3, 'IfcProducts contained in myIfcRailway', 'Association between IfcProducts and IfcRailway', (#20), #15);
+#20 = IFCALIGNMENT('1FNFyCAJeHwxedwDZHIYIu', #3, 'Spor', 'optional Railway Description', $, #153, #156, $);
+#21 = IFCALIGNMENTHORIZONTAL('1FNFyDAJeHwv87wDZHIYIu', $, $, $, $, #94, $);
+#23 = IFCRELNESTS('3BJTAQrjCHwvVKbERtTLTf', $, $, $, #20, (#21, #41, #61));
+#28 = IFCCARTESIANPOINT((0., 0.));
+#29 = IFCALIGNMENTHORIZONTALSEGMENT($, $, #28, 0., 300., 0., 100., $, .HELMERTCURVE.);
```
Ghesselink, line 31: It would be nice to have a file that fails for specifically for helmertcurves as well. 
Can't this be as simple as taking a file that now passes and change the predefined types of one segment to `HELMERTCURVE`. If I'm correct, it should fail then. For instance, the file `test/files/ala002/pass-ala002-business-logic_only.ifc`

.github/workflows/ci.yml
```diff
@@ -17,10 +17,10 @@ jobs:
       run: |
         git submodule init && git submodule update --recursive --remote
 
-    - name: Set up Python 3.9
+    - name: Set up Python 3.11
```
Ghesselink, line 20: Hmm, I'm actually a bit surprised that the tests were running correctly before for case matching (e.g. in thens/alignment.py)

Ghesselink (changes requested):
Looks good. Just some very small comments;.

ifc_validation_models
aothms, line 1: Submodule ref is moved. (I know they are annoying...)

aothms (changes requested):

.github/workflows/ci.yml
civilx64: Agree.  Pattern matching didn't land until python 3.10.

ifc_validation_models
civilx64: I did this on purpose, to bump the submodule ref to the latest commit on `main`.  Did I miss something?

.github/workflows/ci.yml
civilx64:
```
[Unit Test with pytest/build-linux]   ✅  Success - Main Lint with flake8
[Unit Test with pytest/build-linux] ⭐ Run Main Test with pytest
[Unit Test with pytest/build-linux]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/6] user= workdir=
| ============================= test session starts ==============================
| platform linux -- Python 3.10.12, pytest-8.1.1, pluggy-1.4.0 -- /usr/bin/python3
| cachedir: .pytest_cache
| rootdir: /home/scott/projects/ifc-gherkin-rules
| configfile: pyproject.toml
collected 239 items         
```

looking at the logs the github action appears to use system python (`/usr/bin/python3`) of the docker container, which happens to be 3.10 - which is why the tests were passing before even with the yaml file specifying 3.9.

features/steps/thens/alignment.py
civilx64: Renamed as suggested.

test/files/ala002/pass-ala002-helmert_curve.ifc
civilx64: Added a `fail` file for Helmert Curve

features/steps/thens/alignment.py
```diff
@@ -61,15 +61,20 @@ def count_segments(logic, representation):
...
     """
     try:
-        logic_count = len(logic.segments)
+        expected_count = 0
+        for seg in logic.segments:
+            if seg.PredefinedType == "HELMERTCURVE":
+                expected_count += 2
+            else:
+                expected_count += 1
     except AttributeError:
```
aothms, line 70: Later we do `(logic_only or rep_only)`. So this seems to imply that the try-except block is to catch `logic.segments`. But the block also catching `seg.PredefinedType` so if there is an unexpected instance there without a predefined type attribute, the loop is terminated and the segments that were presented are not considered. This is not really a concern (because we have schema validation in place anyway), it's just that I find the intention behind try catch a bit unclear.

I would rather see `if logic.has_segments:` which makes it much more explicit which cases we're expecting and handling. But for me it's not necessary to address this now.

aothms (approved):

features/steps/thens/alignment.py
civilx64: I went ahead and revised this while still fresh in my mind.

PR #162, opened 2024-04-10
Correct ALS005 to allow 'FootPrint-Curve2D' as well as 'Axis-Curve3D'

civilx64:
Addresses [VAL-410](https://bsi-technicalservices.atlassian.net/browse/VAL-410)


features/steps/thens/attributes.py
```diff
@@ -60,11 +60,18 @@ def step_impl(context, inst, attribute, value):
...
+        value = tuple(opts)
+        pred = misc.reverse_operands(operator.contains)
 
     if isinstance(inst, (tuple, list)):
         inst = inst[0]
     attribute_value = getattr(inst, attribute, 'Attribute not found')
-    if not pred(attribute_value, value):
+    if inst is None:
+        # nothing was activated by the Given criteria
+        yield ValidationOutcome(inst=inst, severity=OutcomeSeverity.EXECUTED)
```
aothms, line 73: @Ghesselink 

This seems to be a remaining issue that we haven't fully wrapped our head around. This sounds like something we should address in the decorator, but it's a bit the same dilemma as always.

- We don't want to check in every step for `inst is None`, so why not skip these already in the decorator.
- But we also don't know if there are then steps that do need to ability to check for None, a bit like the "Then assert existence" we had in gcert.
- We want to retain the `None`s in there in order to retain the correspondence between various levels of "the stack".
- How do we distinguish between `None`s coming from optional attributes being empty and "applicability filters not yielding anything", e.g:
  - `Given Its attribute Representation`. Representation can be $ in IFC -> None as attribute value
  - `Given Name=FreeBeer`. Name != 'FreeBeer' in IFC -> None to retrain stack level correspondence

I've created something for this in the backlog.

Ghesselink: To keep things connected; backlog item is V414. Also applicable in V-423

PR #164, opened 2024-04-10
VAL-354

JTurbakiewicz:


aothms:
The name inside the feature is still called "Feature: PSE001 - IfcPropertySet validation"

JTurbakiewicz:
@aothms, good spot, thank you and sorry. I didn't add the change to the file after renaming, my bad.

evandroAlfieri:
@aothms @Ghesselink can be merged?

Ghesselink (approved):
Just ran it locally, it looks excellent. Haven't looked a lot at this rule, but can totally understand why the error occurs.
The only small thing I could foresee is some hefty db memory use in case the list of acceptable values gets long. 
Screenshot 2024-04-12 at 14 56 34

PR #165, opened 2024-04-11
OJP001 - fix

JTurbakiewicz:


evandroAlfieri:
@aothms @Ghesselink can be merged?

features/steps/thens/attributes.py
```diff
@@ -25,7 +25,10 @@ def step_impl(context, inst, entity, other_entity, relationship):
             relating_obj_placement = relating_object.ObjectPlacement
             entity_obj_placement_rel = getattr(related_obj_placement, "PlacementRelTo", None)
             if relating_obj_placement != entity_obj_placement_rel:
-                yield ValidationOutcome(inst=inst, expected=relating_obj_placement, observed=entity_obj_placement_rel, severity=OutcomeSeverity.ERROR)
+                if entity_obj_placement_rel:
+                    yield ValidationOutcome(inst=inst, expected=relating_obj_placement, observed=entity_obj_placement_rel, severity=OutcomeSeverity.ERROR)
```
Ghesselink, line 29:
```suggestion
                     yield ValidationOutcome(inst=inst, expected={"entity":f"{relating_obj_placement.is_a()}(#{relating_obj_placement.id()})"}, observed={"entity":f"{entity_obj_placement_rel.is_a()}(#{entity_obj_placement_rel.id()})"}, severity=OutcomeSeverity.ERROR)
```

Then the error message would be 
expected = {"entity": "IfcLocalPlacement(#31)"}
observed = {"entity": "IfcLocalPlacement(#46)"}

That's maybe a bit clearer

Ghesselink (changes requested):
Looks good :) 

Two things: 
- As you noted already in Jira, the expected/observed is not so clear. 
- We don't seem to have testfiles for the first scenario (`'The relative placement of that IfcElement must be provided by an IfcLocalPlacement entity'`)

features/steps/thens/attributes.py
aothms: I would say we keep this kind of formatting in the generic place in validation_handling.py.

In expected/observed we just do `relating_obj_placement` vs `entity_obj_placement_rel`.

Ghesselink: Done in [format observed entity_instance in validation_handling rather than st…](https://github.com/buildingSMART/ifc-gherkin-rules/commit/44ee60b1098a7f40cd68aff831757dc3a4d172b4)
Screenshot 2024-04-12 at 17 24 23

This is also applied to IBP001

PR #58, opened 2023-03-19
Ri 07_agreement_on_stationing_property

JTurbakiewicz:

Issue #49, opened 2023-02-09
General dude about this repo


IbimDesarrollo:
Can someone explain me was the difference between gherkin rules and IDS standard? I see that both can check ifc models, but i dont know why development it is not focus in one system.
Thanks in advance

evandroAlfieri:
The **bSI Validation Service** (which uses Gherkin rules for some of its checks) validate an IFC files against syntax, schema, and other constraints of the IFC standard (schema + specification). It tells you if your file is a valid IFC file.

**IDS** can be use to demand (and check) information that are specific for a use-case, a project, an organisation, etc. You can use it to check if an IFC file contains the information you needed for a specific use.

Corollary: they are not "competing systems". they complement each others. That is why there was (is) effort in both directions, and not only one.

PR #43, opened 2023-02-02
add new description for alb002 and consistent description alb003

Ghesselink:
A new description is added for ALB002 and the description of ALB003 is now consistent with the other descriptions. 

Rules that don't have a description yet:

- GEM001
- GEM002
- SPS001

PR #166, opened 2024-04-13
simplify retrieval of activation instances

Ghesselink:
See VAL-423

A couple of remarks 
- Thomas' work of resetting the context stack perhaps resulted in (I think) a more easy and robust retrieval of the first layer of `instances` out of the stack. 
- Instead of filtering out None, we can keep the None in. It was mentioned before that the difficulty with this would be to distinquish between 'None because no result is found' and 'None as a result'. I'm still unsure how to handle this generically in validation_handling. Especially this line might give us some trouble 
```
            if inst is None:
                return
```


aothms:
Great! This is more or less what I was trying to write in Jira as well. But maybe we can do the same in the given handling branch. Instead of ever executing a given statement on None, just immediately propagate it to None in the stack/context.instances.

I see quite some cases of "Attribute not found" in the tabulate outcome. I think even if we previously filtered out top level `None`s in context.instances. We don't filter out `[None,None]` or even `[None]`. Which we would get in the case of multiple indirections of aggregates such as Product>ProductDefinionShape.

After deployment we can work on our own NoneType sibling that we can use to indicate the difference between NotApplicable/EmptyValueInModel.

Ghesselink:
> But maybe we can do the same in the given handling branch. Instead of ever executing a given statement on None, just immediately propgate it to None in the stack/context.instances

Done this in [given default Nonetype output](https://github.com/buildingSMART/ifc-gherkin-rules/pull/166/commits/01a6fa83533bcf017e627389a4a42c5b2755a508) . In case of None, it doesn't execute the statement but just returns None (therefore we don't get the errors that occur when executing something on inst=None).

> Which we would get in the case of multiple indirections of aggregates such as Product>ProductDefinionShape.

I'm not sure this would be a problem. How I see it currently: in `map_sate`, we recursively go over the list until we get an item that is not a sequence. So, instead of calling `apply_operation` (which in turn calls the given implementation) on `[None, None] `oncee, we'd call `apply_operation` twice, once for each None in that list. With the latest commit `[None, None]` will therefore output `[None, None]`. 
The exception for this is the `depth=1` cases. This will call the step_imp function on on [None, None]. However, only current implementation is 'Its final segment', selecting the last element in the list. This should, again, just return `None`. 
Or do you still get the same errors? So far I'm not, but I'll still be running some files in the gherkin repository this afternoon. 

> After deployment we can work on our own NoneType sibling that we can use to indicate the difference between NotApplicable/EmptyValueInModel.

Agree. I think it's a good example of what you stated 
> it's mostly just making fully explicit all our implicit understanding

aothms:
> I'm not sure this would be a problem

No it's not a problem. It's just an observation. We recursively go into `context.instances` potentially returning None. So in addition to `context.instances = [None, None]` we could have something like `context.instances = [[None, None], [None, None]]`. But we only filtered `context.instances = list(filter(None, context.instances))`. So after that filtering, in the first case, we would end up with `context.instances = []`. In the second case we would retain `context.instances = [[None, None], [None, None]]`. At least that's my interpretation.

Anyway, I'm really in favour of this PR. +7 -34 is really a good sign. I think this again greatly simplifies things.

afbeelding

But let's wait until monday to see how we proceed with this. I think in the last couple of days we've been working on a much better mental model behind how we execute our steps. It's just a bit unfortunate that all of this comes now. At some point we said "let's just merge in the decorator branch, because anyway it's better than what we have now", but we never really tied up all the loose ends after that.

test/files/als004/fail-als004-scenario01-wrong_representationidentifier_value.ifc
```diff
@@ -74,6 +73,7 @@ DATA;
 #64=IFCALIGNMENTVERTICAL('0dZodT25PB4PjSqv7D1BCl',#3,'V1',$,$,#380,$);
 #65=IFCRELNESTS('3ZLxvT5W55ARIuY$a2OpaK',#3,'name',$,#64,(#66,#68,#70,#72,#74));
 #66=IFCALIGNMENTSEGMENT('1N_pe3V5P5lBTAZb7Fr6OX',#3,'V1 001 CONSTANTGRADIENT',$,$,#310,#313,#67);
+#34=IFCALIGNMENTSEGMENT('31tSVngQb0XRlPErBWB1XW',#3,'H1 001 LINE',$,$,#179,#182,#35);
```
aothms, line 76: Oh just one more comment. Just moving an instance down in the file doesn't actually change things. At this line the order is defined:

    #33=IFCRELNESTS('0YIHzLddz2kfrHsBRhMEDr',#3,'name',$,#26,(#34,#37,#40,#43,#46,#49,#52,#55,#58,#61));


I do think it's a good idea to test with some files where the error is not on the first instance, but better create a new file and actively modify, e.g:

```
#40=IFCALIGNMENTSEGMENT('2K28_c3_bCK853aPYKlp2m',#3,'H1 003 CIRCULARARC',$,$,#205,#208,#41);
#208=IFCPRODUCTDEFINITIONSHAPE($,$,(#209));
#209=IFCSHAPEREPRESENTATION(#17,'Axis','Segment',(#204));
```

Ghesselink: Hmm, the order in which context.instances were passed to the handle_then was changed (at index 10 rather than 0). I'm not really understanding why; when I manually check `instances=file.by_type("IfcAlignmentSegment")`, the order is unchanged, as you mentioned.. (#34 still == instances[0])

Changed line #209 into '`Curve2D`' and now the error is at index 2 (as #40 is after #34 and #37; the other instances of type `IfcAlignmentSegment`). Which is both expected and understood. 
[change index fail file for testing](https://github.com/buildingSMART/ifc-gherkin-rules/pull/166/commits/8c6e37241f44844be5dcc2ba24cc0d6334b6ee2b)

PR #169, opened 2024-04-18
throw big error on false negative

Ghesselink:


aothms:
Has this been tested when used as part of the Val service? Isn't merging
this in a bit premature?

Sent from a mobile device, excuse my brevity. Kind regards, Thomas

Op do 18 apr 2024 15:07 schreef Evandro Alfieri ***@***.***>:

aothms:
@evandroAlfieri @Ghesselink 

> Has this been tested when used as part of the Val service? Isn't merging
this in a bit premature?

PR #175, opened 2024-04-26
Part composing Part

jmirtsch:
Parts should be permitted to compose of other parts

PR #176, opened 2024-04-26
fix/part-composing-part

Ghesselink:
Continuation of https://github.com/buildingSMART/ifc-gherkin-rules/pull/175

> Parts should be permitted to compose of other parts

- Changes are based on dev branch; ci/cd was failling due to outdated ifcopenshell version
- Commits of @jmirtsch are added to the resource folder
- Includes a testfile to ensure tests are running correctly.

cc @pjanck 

> lftm - related to https://github.com/buildingSMART/validate/issues/61 and https://github.com/jwouellette/TPF-5_372-Unit_Test_Suite/issues/14


evandroAlfieri (approved):
Lgtm

jmirtsch:
I have one more scenario to consider.  Perhaps it already was (and if so, if there are any documented discussions it would be great to get a reference).  Should a site be able to compose a facility?
Ie if I have a road facility, it might be composed of multiple sites.  If this isn't permitted, then the class IfcSpace would be the likely candidate to be used in a containment hierarchy.   An alternative implementation might be to have sites in their own hierarchy and have elements referenced to the site.  At least one authoring application authors multiple sites and the root element as a facility (ie ifcroad, or Ifcrail) would be a logical choice to me.

aothms:
>  An alternative implementation might be to have sites in their own hierarchy and have elements referenced to the site.

At least in the alignment that would already be possible, because it is kind of lifted out of the spatial hierarchy and directly aggregated into the project. So it's indeed a relevant question what that implies for road.

>  Should a site be able to compose a facility?

Even in the case of buildings this has been brought up. You could have a construction site on the 3rd storey of an existing building, why not?

But this really relates to the ambiguity around Site and Project. Are these geospatial reference points, the terrain, the plot, the construction area. This has never been really established and the various attributes, properties and usages all relate to various facets.

In the end I'm all for predictable and consistent exchanges though. I'm wondering if a sensible compromise would be something like the following, where lines indicate aggregation, containment, or references.

afbeelding

aothms:
Created separate issues for the discussion points. Moving these changes along in the meantime.

PR #177, opened 2024-04-27
Fix/val-247 readme adjustments

Ghesselink:
Add new up-to-date README files outputting the same data as in the frontend of the vs service (instance_id, expected, observed). 
Additionally, a modified script to automatically generate this format is added. 

Future to-do; the behave output conversion to tabulate/pytest is perhaps a bit outdated and could use a lick of paint.


evandroAlfieri (approved):

test/generate_markdown.py
```diff
@@ -0,0 +1,159 @@
...
+    from test.utils import collect_test_files
+
+
+MarkdownTableStr = NewType('MarkdownTableStr', str)
+
+def handle_markdown_table(markdown_table: MarkdownTableStr) -> pd.DataFrame:
+    """
+    Handling in case the already existing file is in non-html markdown format
+    """
+    lines = [line.strip() for line in markdown_table.split("\n")]
```
aothms, line 23: You can use `markdown.markdown()` (`pip install markdown`) to convert the markdown table to html.

test/generate_markdown.py
```diff
@@ -0,0 +1,159 @@
...
+                    df = handle_markdown_table(content)
+                elif has_html_table:
+                    df = pd.read_html(content)[0]
+                else:
+                    raise ValueError(f"The file '{readme_path}' does not contain a recognized HTML or Markdown table. Please ensure the file adheres to the expected format.")
+
+                if base not in df['File name'].values:
+                    df = update_html_markdown(df, content_testfile)
+
+                    file.write(df.to_html(index=False))
```
aothms, line 138:
```suggestion
                    file.write(df.to_markdown(index=False))
```

Kan we keep using markdown tables? Or otherwise CSV? HTML is rather difficult to hand-edit.

aothms (changes requested):

test/generate_markdown.py
Ghesselink: All converted to markdown 
[readme.md in markdown](https://github.com/buildingSMART/ifc-gherkin-rules/pull/177/commits/34ddd9e8bb125d0d1299e4cb263fb8d909b300d1)

test/generate_markdown.py
Ghesselink: We're now using html and converting it to markdown only in the end, so this is not required anymore. 
[update markdown generate script](https://github.com/buildingSMART/ifc-gherkin-rules/pull/177/commits/1ea22a7c634b248eb857d40f1831d26b57aa9c74)

Ghesselink:
Added a couple of commits : 

- Update markdown generator. It's not always straightforward to get the output message (which is a string representation of `ValidationOutcome`) back to a dict in a generic way. I think it's working better now. 
- Add a script to create a readme file for all rules with one command `(python test/run_generate_markdown.py`)
- Disabled the disabled rules from generating. 
- Run the previous mentioned script once to generate markdown README.md files

PR #178, opened 2024-04-29
fix/VAL-459 ALB015 fail reports too many entities

civilx64:
creates a set from `context.instances` to ensure unique entities


civilx64:
Superseded by #179

PR #179, opened 2024-04-29
Fix/val-459-combine-relationships-with-stacked-selection

Ghesselink:
Modified: 

- The more standardised and tested work from gcert `'A relationship X between U and Z'` with an optional `and following that ` moved over to the current gherkin repository. 
- OJP001. Now we're stating `'and following that' ` to continue with the subsequent element (and not the initial activation instance)
- ALB015. With using the statement 'its final segment at depth 1' from the previous ALS015 rule, we're selecting the last `IfcAlignmentSegment` from context.instances, thus avoiding the previous found duplication.

### DIFF
--- a/features/ALB015_Alignment-business-logic-zero-length-final-segment.feature
+++ b/features/ALB015_Alignment-business-logic-zero-length-final-segment.feature
@@ -10,7 +10,7 @@ Feature: ALB015 - Alignment business logic zero length final segment
     Given A model with Schema "IFC4.3"
     Given An IfcAlignmentHorizontal
     Given A relationship IfcRelNests from IfcAlignmentHorizontal to IfcAlignmentSegment and following that
-    Given Its final segment at depth 1
+    Given Its final element at depth 1 filtering by id
     Given Its attribute DesignParameters
     Then The SegmentLength of the final IfcAlignmentHorizontalSegment must be 0
 
@@ -18,7 +18,7 @@ Feature: ALB015 - Alignment business logic zero length final segment
     Given A model with Schema "IFC4.3"
     Given An IfcAlignmentVertical
     Given A relationship IfcRelNests from IfcAlignmentVertical to IfcAlignmentSegment and following that
-    Given Its final segment at depth 1
+    Given Its final element at depth 1 filtering by id
     Given Its attribute DesignParameters
     Then The HorizontalLength of the final IfcAlignmentVerticalSegment must be 0
 
@@ -26,7 +26,6 @@ Feature: ALB015 - Alignment business logic zero length final segment
     Given A model with Schema "IFC4.3"
     Given An IfcAlignmentCant
     Given A relationship IfcRelNests from IfcAlignmentCant to IfcAlignmentSegment and following that
-    Given Its final segment at depth 1
+    Given Its final element at depth 1 filtering by id
     Given Its attribute DesignParameters
     Then The HorizontalLength of the final IfcAlignmentCantSegment must be 0
-
--- a/features/ALB015_Alignment-business-logic-zero-length-final-segment.feature
+++ b/features/ALB015_Alignment-business-logic-zero-length-final-segment.feature
@@ -12,7 +12,7 @@ Feature: ALB015 - Alignment business logic zero length final segment
     Given A relationship IfcRelNests from IfcAlignmentHorizontal to IfcAlignmentSegment and following that
     Given Its final element at depth 1
     Given Its attribute DesignParameters
-    Then The SegmentLength of the final IfcAlignmentHorizontalSegment must be 0
+    Then The SegmentLength of the IfcAlignmentHorizontalSegment must be 0
 
   Scenario: Validating final segment of vertical alignment business logic
     Given A model with Schema "IFC4.3"
@@ -20,7 +20,7 @@ Feature: ALB015 - Alignment business logic zero length final segment
     Given A relationship IfcRelNests from IfcAlignmentVertical to IfcAlignmentSegment and following that
     Given Its final element at depth 1
     Given Its attribute DesignParameters
-    Then The HorizontalLength of the final IfcAlignmentVerticalSegment must be 0
+    Then The HorizontalLength of the IfcAlignmentVerticalSegment must be 0
 
   Scenario: Validating final segment of cant alignment business logic
     Given A model with Schema "IFC4.3"
@@ -28,4 +28,4 @@ Feature: ALB015 - Alignment business logic zero length final segment
     Given A relationship IfcRelNests from IfcAlignmentCant to IfcAlignmentSegment and following that
     Given Its final element at depth 1
     Given Its attribute DesignParameters
-    Then The HorizontalLength of the final IfcAlignmentCantSegment must be 0
+    Then The HorizontalLength of the IfcAlignmentCantSegment must be 0
--- a/features/ALB015_Alignment-business-logic-zero-length-final-segment.feature
+++ b/features/ALB015_Alignment-business-logic-zero-length-final-segment.feature
@@ -9,23 +9,24 @@ Feature: ALB015 - Alignment business logic zero length final segment
   Scenario: Validating final segment of horizontal alignment business logic
     Given A model with Schema "IFC4.3"
     Given An IfcAlignmentHorizontal
-    Given A relationship IfcRelNests from IfcAlignmentSegment to IfcAlignmentHorizontal
-    Given Its final IfcAlignmentSegment
+    Given A relationship IfcRelNests from IfcAlignmentHorizontal to IfcAlignmentSegment and following that
+    Given Its final segment at depth 1
     Given Its attribute DesignParameters
     Then The SegmentLength of the final IfcAlignmentHorizontalSegment must be 0
 
   Scenario: Validating final segment of vertical alignment business logic
     Given A model with Schema "IFC4.3"
     Given An IfcAlignmentVertical
-    Given A relationship IfcRelNests from IfcAlignmentSegment to IfcAlignmentVertical
-    Given Its final IfcAlignmentSegment
+    Given A relationship IfcRelNests from IfcAlignmentVertical to IfcAlignmentSegment and following that
+    Given Its final segment at depth 1
     Given Its attribute DesignParameters
     Then The HorizontalLength of the final IfcAlignmentVerticalSegment must be 0
 
   Scenario: Validating final segment of cant alignment business logic
     Given A model with Schema "IFC4.3"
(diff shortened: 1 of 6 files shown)


features/steps/givens/attributes.py
```diff
@@ -7,9 +7,7 @@
 from validation_handling import gherkin_ifc
 from . import ValidationOutcome, OutcomeSeverity
 
-register_type(file_or_model=TypeBuilder.make_enum(dict(map(lambda x: (x, x), ("file", "model")))))
-register_type(plural_or_single=TypeBuilder.make_enum(dict(map(lambda x: (x, x), ("plural", "single")))))
-
+register_type(first_or_final=TypeBuilder.make_enum({"first": 0, "final": 1 }))
```
aothms, line 10:
```suggestion
from enum import Enum, auto
class FirstOrFinal(Enum):
  FIRST = auto()
  FINAL = auto()

register_type(first_or_final=TypeBuilder.make_enum({"first": FirstOrFinal.FIRST, "final": FirstOrFinal.FINAL }))
```

What do you think? I didn't test this, but I find 0 and 1 too easy to mix up.

features/steps/givens/attributes.py
```diff
@@ -74,27 +72,24 @@ def step_impl(context, inst, attribute, condition, prefix):
...
+@gherkin_ifc.step('Its {ff:first_or_final} element at depth 1 filtering by {filter}')
+def step_impl(context, inst, ff=0, filter=None):
+    def get_filter_key(attr):
+        if attr == 'id':
+            return lambda entity: entity.id() # in case of a method
+        else:
+            return lambda entity: getattr(entity, attr, None) # in case of an attribute
+
+    if ff:
+        yield ValidationOutcome(instance_id =  max(inst, key=get_filter_key(filter)) if filter else inst[-1], severity=OutcomeSeverity.PASSED)
```
aothms, line 89: I don't understand the intention behind this `... filtering by id`.

Based on that you're taking the instance with the highest or lower (`min`, `max` respectively using a key transformation of inst -> inst.id() [which doesn't exist on non-instances]). But the id is only a local naming mechanism in the exchange format, it really should not have any significance outside of the exchange format.

aothms (changes requested):

features/steps/givens/attributes.py
aothms:
```
    Given A model with Schema "IFC4.3"                                                               # features/steps/givens/attributes.py:45 0.002s
    Given An IfcAlignmentCant                                                                        # features/steps/givens/entities.py:12
[#2388=IfcAlignmentCant('0kwyzC0yb8ehcYFOVppIP9',#1,'C1',$,$,#2389,$,1.435)]

    Given A relationship IfcRelNests from IfcAlignmentCant to IfcAlignmentSegment and following that # features/steps/givens/relationships.py:13
[[#2435=IfcAlignmentSegment('33RHONOYr9mBgFO3_aTR9K',#1,'c8',$,$,#2436,#2680,#2440), #2441=IfcAlignmentSegment('0v2fCSQ8H1qejZRDkDUoz1',#1,'c9',$,$,#2442,#2691,#2446), #2393=IfcAlignmentSegment('0o7VCJvavDdA9xYwz5oBRK',#1,'c1',$,$,#2394,#2603,#2398), #2399=IfcAlignmentSegment('2pwzDsynP3ggq3OBYyg7wr',#1,'c2',$,$,#2400,#2614,#2404), #2405=IfcAlignmentSegment('2tLO9I_A1DbQwnSyol1_P8',#1,'c3',$,$,#2406,#2625,#2410), #2411=IfcAlignmentSegment('1CWfkLVFXCo8JV8zZdrLfo',#1,'c4',$,$,#2412,#2636,#2416), #2417=IfcAlignmentSegment('08LKrvnvj079Xxjd3KbIy9',#1,'c5',$,$,#2418,#2647,#2422), #2423=IfcAlignmentSegment('1_39nu3q10EgiUF2uuPA6u',#1,'c6',$,$,#2424,#2658,#2428), #6001=IfcAlignmentSegment('1v2fCSQ8H1qejZRDkDUoz2',#1,'cA',$,$,#2442,#6101,#6011), #2429=IfcAlignmentSegment('2e6W3TPHX9jAKv_l1faiaw',#1,'c7',$,$,#2430,#2669,#2434)]]

    Given Its final element at depth 1                                                               # features/steps/givens/attributes.py:77
[#2429=IfcAlignmentSegment('2e6W3TPHX9jAKv_l1faiaw',#1,'c7',$,$,#2430,#2669,#2434)]

    Given Its attribute DesignParameters                                                             # features/steps/givens/attributes.py:62
[#2434=IfcAlignmentCantSegment($,$,740.169262678128,109.431749924283,3.E-2,3.E-2,-3.E-2,-3.E-2,.CONSTANTCANT.)] 

    Then The HorizontalLength of the final IfcAlignmentCantSegment must be 0                         # features/steps/thens/attributes.py:94 0.001s
      Assertion Failed: Errors occured:
      ["Feature=ALB015 Outcome=E00020 Severity=ERROR Expected={'value': 0.0} Observed={'value': 109.431749924283}"]

```

^ `#2429` is indeed the last segment of the list

What was the filtering by id intended to do?

aothms:
Btw @civilx64 the issue was essentially that we're not using `inst`. The step_impl function is being ran for every segment. And every time we yield the last instance. We should only yield that basically when context.instances[-1] equals inst. But even though you yield context.instances[-1] the decorator still traces this back to the original inst as the "activation instance", so that's why we don't get exactly duplicate rows in the report.

But there is another problem that, in case of multiple aligments, context.instances[-1] is only the last segment of the last alignment in the model. With proper amounts of magic, the new code from Geert dus use a proper nesting so that the segments are grouped per alignment and we can precisely take the last element per alignment, which is what the `depth 1` does.

```
@gherkin_ifc.step('Its final IfcAlignmentSegment')
def step_impl(context, inst):
    """
    Implement ALB015
    This is a separate function from ALS015 because ALB015 needs to yield a ValidationOutcome.
    """
    yield ValidationOutcome(instance_id=context.instances[-1], severity=OutcomeSeverity.PASSED)
```

aothms:
I added two commits:
- remove ... filtering by id
- remove 'final' from the then predicate, because it doesn't concern that step
- make order consistent in relationship traversal

features/steps/givens/attributes.py
civilx64: Filtering by id was being used to select the last segment of the list because in our test files the STEP id is always increasing as you traverse along the alignments.  As you noted above the id is not semantically significant and not guaranteed to always be in increasing order along the alignment.

`Its final element at depth 1` also aligns nicely with ALS015 which is the companion to ALB015.

civilx64:
> Btw @civilx64 the issue was essentially that we're not using `inst`. 

Thanks for pointing that out.  Hopefully I'll start getting this right in the future and you won't have to keep pointing it out in code reviews.

features/ALB015_Alignment-business-logic-zero-length-final-segment.feature
civilx64, line 1: increment to `@version2`

features/ALS015_Alignment-representation-zero-length-final-segment.feature
civilx64, line 1: increment to `@version2`

features/steps/givens/attributes.py
civilx64: @Ghesselink If we leave as-is without the enum please change "final" to be -1 for consistency with python list indexing.

civilx64 (changes requested):

features/steps/givens/attributes.py
Ghesselink: I thought it would be the easiest and best readable to go along with the suggestiion of @aothms 
See commit [first_and_final adjustmenet, remove filtering by](https://github.com/buildingSMART/ifc-gherkin-rules/pull/179/commits/80533ec882a87f0d85d314ef4ce5aa6ba6bd3dd5) 

Also here the filtering_by is removed as it's not used anymore.

Ghesselink:
> I added ~two~ n commits:
> 
> * remove ... filtering by id
> * remove 'final' from the then predicate, because it doesn't concern that step
> * make order consistent in relationship traversal
> 
> Let me know what you think

Great, thanks for looking. I'm somehow happy that Scott run into this as it was a good incentive to implement our work on the `'a relationship X with Y'` statement. 
It was running correctly sometimes locally, and sometimes not. Also suspected it had something to do with the random order of sets in the new gcert implementation (so the test pass when coincidently the correct instance was selected). . 

> remove ... filtering by id

Also removed in the step_impl [first_and_final adjustmenet, remove filtering by](https://github.com/buildingSMART/ifc-gherkin-rules/pull/179/commits/80533ec882a87f0d85d314ef4ce5aa6ba6bd3dd5)


For a separate discussion (let's merge this PR), as Scott mentioned ; 
> So vertical and cant segments have a StartDistAlong attribute that we can filter by so all good there.  

It might be beneficial to consider alternative criteria for determining the "first" and "last" elements (ideally in a single step). Currently, we rely on the order as specified in the composition or decomposition instances (like IfcRelNests). I'm not sure if this makes sense, but could this approach lead to issues?   For instance, when identifying the "first element of a road," one would naturally expect this to refer to the starting point along the road's direction. But is this always the case/does this sometimes go wrong (e.g. some other critierum is used)?

As noted, for the rest it looks good. Let's continue this (if necessary) somewhere else and merge the PR. Similar to the PR fixing part-composing-part

features/steps/givens/attributes.py
```diff
@@ -74,34 +76,16 @@ def step_impl(context, inst, attribute, condition, prefix):
...
-def step_impl(context, inst):
-    """
-    Implement ALB015
-    This is a separate function from ALS015 because ALB015 needs to yield a ValidationOutcome.
-    """
-    yield ValidationOutcome(instance_id=context.instances[-1], severity=OutcomeSeverity.PASSED)
 
+@gherkin_ifc.step('Its {ff:first_or_final} element')
+@gherkin_ifc.step('Its {ff:first_or_final} element at depth 1')
+def step_impl(context, inst, ff=0, filter=None):
```
aothms, line 83:
```suggestion
def step_impl(context, inst, ff : FirstOrFinal):
```

We can probably leave out the default value. `filter` is no longer  used. Maybe add a type hint.

aothms:
> It might be beneficial to consider alternative criteria for determining the "first" and "last" elements

Yeah in that regard I liked how you approached it with `min(key=...)`. But in this case the last segment is not really something that concerns end-users. It really is an IFC-specific thing because if we don't have such a segment we don't have the direction at that end point.

PR #192, opened 2024-05-12
onhold/pse001_modifications

Ghesselink:
- unwrangle gherkin statement matching inside implemetation code
- schema implicit for csv lookup

todo: 
- Unwrangle predefined-types


features/steps/givens/attributes.py
```diff
@@ -68,12 +68,16 @@ def step_impl(context, inst, attribute, tail="single"):
 @gherkin_ifc.step("Its {attribute} attribute {condition} with {prefix}")
 def step_impl(context, inst, attribute, condition, prefix):
     assert condition in ('starts', 'does not start')
+    attr = str(getattr(inst, attribute, ''))
     if condition == 'starts':
-        if hasattr(inst, attribute) and str(getattr(inst, attribute, '')).startswith(prefix):
+        if attr and attr.startswith(prefix):
+            context.activation_attr = attr
```
aothms, line 74: I don't think this works. Isn't this one global variable for  the entire model? Did you test with multiple psets in one file?

features/steps/thens/psets.py
```diff
@@ -0,0 +1,243 @@
...
+def step_impl(context, inst, table):
+    property_set_definitions = get_pset_definitions(context, table)
+
+    if inst not in property_set_definitions.keys():
+        yield ValidationOutcome(inst=inst, observed = {'value':inst}, severity=OutcomeSeverity.ERROR)
+
+
+@gherkin_ifc.step('It must be named according to the property set definitions table "{table}"')
+def step_impl(context, inst, table):
+    accepted_values = get_accepted_pset_values(context, table)
```
aothms, line 117: What is `inst` here? How can we rewrite this without a global state in context.activation_atrr?

aothms (changes requested):

features/steps/thens/psets.py
Ghesselink: The name of the each individual property within a property set. 
```
        Given its attribute HasProperties
        Given its attribute Name -> inst 
```

Alternatively we could go back in the stack tree. Probably this would be more reliable and easier to trace down in the future then setting global variables.

features/steps/givens/attributes.py
Ghesselink: I think you're right. It did work with 
Screenshot from 2024-05-12 17-23-09
But that's only because they have the same pset name. But when the names are different it's overwritten before the next statement is reached.

features/steps/thens/psets.py
Ghesselink: Or use the activation_inst that we determined when implementing the gcert work. E.g. in 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/4217e02b06a295c73ca2c93125a565fe0151bce1/features/steps/validation_handling.py#L245

We could add it to  
`step_results = list(filter(lambda x: x.severity in [OutcomeSeverity.ERROR, OutcomeSeverity.WARNING], list(fn(context, inst=inst, **kwargs))))`

So 
```
            context.current_activation_inst = activation_inst
            step_results = list(filter(lambda x: x.severity in [OutcomeSeverity.ERROR, OutcomeSeverity.WARNING], list(fn(context, inst=inst, **kwargs))))
            context.current_activation_inst = None
```
Gives access to the relevant pset within the `step_impl` func

PR #195, opened 2024-05-26
Sps007

Ghesselink:
Rule handling spatial composition concept template with VS statements. 
https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Connectivity/Spatial_Structure/Spatial_Containment/content.html

#edit; todo is still to handle `ContainedInStructure` component
Needs to be restructured differently

e.g. 
```
Scenario Outline: 
        Given an <entity> 
        Given A relationship IfcRelContainedInSpatialStructure from IfcSpatialElement to <entity> and following that
        Given ContainedInStructure = not empty
        Given its attribute ContainedInStructure

        Then it must be of length 1 at depth 1
        Then it must be of type IfcSpatialStructure

        Examples: 
              |  entity | 
              |  IfcElement| 
              |  IfcGrid   | 
              |  IfcAnnotation | 
```

Some notes: 
- A custom NoneType is added that can be differentiated from the python build-in used None type. This is to handle case in which we find don't find an attribute for an entity in IFC, but do want to sustain this information in the stack for evaluation in a later step (i.e. at `assert existence.)`.
- The spatial relationship entity seems to be a logical starting point as the activation instance as the constraints in this CT can be derived from it.
- In the statement` 'The current directional relationship must not contain multiple entities at depth 1' `:One challenge was to, in the context stack, differentiate between a length of a sequence that is the result of an attribute lookup at the entity instance level (e.g. `'RelatedObjects'`) as opposed to two entities having one related object each. For instance, in `validation_handling`, [IfcWall, IfcWall] gets unpacked into two successive IfcWall as `inst` in the impl function. However, [(IfcWall, IfcWall)I or [(IfcWall, IfcWall],None] as well. This can be handled, I think, with the `'at depth 1'` functionality. Then, relationships with multiple entities are then always of type `tuple` containing more than one `entity_instance`.
https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Connectivity/Spatial_Structure/Spatial_Containment/content.html


Ghesselink:
https://github.com/buildingSMART/ifc-gherkin-rules/pull/197

PR #196, opened 2024-05-26
OJT001 - object typing

civilx64:


Ghesselink (approved):
I've added a README for the testfiles. 
This can be done now by a single command
`python test/generate_markdown.py ojt001`

Also I've created some testfiles to test it locally and left them in the repository. More testfiles never harm ..

PR #197, opened 2024-05-30
SPS007-Spatial Containment

Ghesselink:
### Gherkin Rule for Handling the Concept Template 'Spatial Containment'

This PR introduces a Gherkin rule to handle the concept template [Spatial Containment](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Connectivity/Spatial_Structure/Spatial_Containment/content.html). .

- **All IfcAnnotation must be contained in a IfcSpatialElement**:
- **All IfcGrid must be contained in a IfcSpatialElement**:
- **All IfcElement subtypes must be contained in a IfcSpatialElement EXCEPT for**:
  -  The subtype is an aggregated part of an other element (e.g IfcElementAssembly)).
   - The subtype is an IfcOpeningElement
   - The subtype is an IfcFeatureElementSubtraction (subtypes of)

- **All other IFC entities must not be contained in a IfcSpatialElement**:

### Further Changes

- **Second part ALB004 deleted**:
    - The following part is redundant; (Given an IfcAlignment) `It must not be contained in IfcSpatialElement directly or indirectly`

- **Custom NoneType (`VsNone`) Implementation**:
  - Added a custom `NoneType`, i.e. `VsNone` to allow `None` in the context stack for validation (sometimes an attribute lookup must indeed be `None`).

- **Attribute Lookup**:
  - Allowed the Gherkin statement `Given entity is_a` as well as the 'not equal' variant for attribute lookup, i.e., `Given != is_a` or `Name != BeautifulName`.

- **Attribute Lookup for both Given as well as Then statements**:
  - Return an error validation outcome with the statement `{attribute} = {value}` in case of **Then**.

- **Relating/Related Entity Handling**:
  - Implemented the statement for `'its relating/related entity'` which can be used as an alternative for the inverse attribute lookup, such as `Given ContainedInStructure` or `Given Decomposes`.


features/SPS007_Spatial-containment.feature
```diff
@@ -0,0 +1,23 @@
+@implementer-agreement
+@SPS
+@version1
+@E00040
+
+Feature: SPS007 - Spatial Containment
+The rule verifies that correct spatial containment - IfcRelContainedInSpatialStructure
```
civilx64, line 7:
```suggestion
The rule verifies that spatial containment via IfcRelContainedInSpatialStructure is utilised in accordance with [Concept Template 4.1.5.13.2](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/concepts/Object_Connectivity/Spatial_Structure/Spatial_Containment/content.html)
```

features/SPS007_Spatial-containment.feature
```diff
@@ -0,0 +1,23 @@
...
+
+    Scenario: Subtypes of IfcElement are placed within the project spatial hierarchy
+        Given an IfcSpatialElement
+        Given A relationship IfcRelContainedInSpatialStructure from IfcSpatialElement to IfcElement and following that
+        Given is_a != IfcElementAssembly
+        Given ContainedInStructure = not empty
+        Given its attribute ContainedInStructure
+        Given its relating entity
```
civilx64, line 15:
```suggestion
        Given its attribute RelatingStructure
```

Can we use an existing step implementation?

features/SPS007_Spatial-containment.feature
```diff
@@ -0,0 +1,23 @@
...
+@version1
+@E00040
+
+Feature: SPS007 - Spatial Containment
+The rule verifies that correct spatial containment - IfcRelContainedInSpatialStructure
+
+    Scenario: Subtypes of IfcElement are placed within the project spatial hierarchy
+        Given an IfcSpatialElement
+        Given A relationship IfcRelContainedInSpatialStructure from IfcSpatialElement to IfcElement and following that
+        Given is_a != IfcElementAssembly
```
civilx64, line 12:
```suggestion
        Given Its Type != IfcElementAssembly
```

`is_a` is familiar for users of `ifcopenshell` but using `Its Type` as suggested - or `Its Entity Type` would be better.

features/steps/givens/attributes.py
```diff
@@ -11,16 +12,40 @@
...
+register_type(include_or_exclude_subtypes=TypeBuilder.make_enum({"including subtypes": SubTypeHandling.INCLUDE, "excluding subtypes": SubTypeHandling.EXCLUDE }))
+register_type(first_or_final=TypeBuilder.make_enum({"first": FirstOrFinal.FIRST, "final": FirstOrFinal.FINAL }))
+register_type(equal_or_not_equal=TypeBuilder.make_enum({
+    "=": ComparisonOperator.EQUAL,
+    "!=": ComparisonOperator.NOT_EQUAL
+}))
+
+def check_entity_type(inst, entity_type, handling: SubTypeHandling):
+    handling_functions = {
+        SubTypeHandling.INCLUDE: lambda inst, entity_type: inst.is_a(entity_type),
```
civilx64, line 33: Add a comment above L32 explaining that `INCLUDE` will evaluate to True if inst is a subtype of `entity_type` while the second function for `EXCLUDE` will evaluate to True only for an exact type match.  The code by itself is not obvious enough to be self-documenting.

features/steps/givens/attributes.py
```diff
@@ -29,8 +54,24 @@ def step_impl(context, inst, attribute, value):
             value = set(map(ast.literal_eval, map(str.strip, value.split(' or '))))
             pred = misc.reverse_operands(operator.contains)
 
-    if hasattr(inst, attribute) and pred(getattr(inst, attribute), value):
+    entity_is_applicable = False
+    if attribute == 'is_a':
```
civilx64, line 58:
```suggestion
    if attribute == 'Type':
```

features/SPS007_Spatial-containment.feature
```diff
@@ -0,0 +1,23 @@
...
+        Given ContainedInStructure = not empty
+        Given its attribute ContainedInStructure
+        Given its relating entity
+
+        Then The current directional relationship must not contain multiple entities at depth 1
+
```
civilx64, line 18: I don't understand how this relates to the concept template.  It appears to be validating the [1:1] cardinality of `IfcRelContainedInSpatialStructure.RelatingStructure` which would be flagged in the VS schema check that runs before the normative rules, no?

features/steps/utils/vs_none.py
```diff
@@ -0,0 +1,56 @@
...
+Usage:
+    from custom_none import VSNone
+
+    print(VSNone)            # VSNone
+    print(VSNone == None)    # True
+    print(VSNone is None)    # False
+    print(bool(VSNone))      # False
+"""
+
+class VSNoneType:
```
civilx64, line 21: We should use a more descriptive and specific name for this custom type.  I recommend `NullAttribute` given that it refers to a blank attribute on an entity instance.

civilx64 (changes requested):

features/SPS007_Spatial-containment.feature
aothms: I would also prefer to not use `!=`  but `is not`.

features/SPS007_Spatial-containment.feature
Ghesselink: That's certainly an option. 
The intention was to create an implementation that is agnostic about the specific attribute used for lookup and to return only the 'relating' side of a relationship entity. Using the same tables as the statements _'A relationship X from/to Y from/to Z'_
In other words, regardless of the type of aggregate relationship entity—such as IfcRelAggregates or IfcRelContainedInSpatialStructure—it always retrieves the 'relating' side of that element. Or, the related side if you phrase it as 
`its related entities`

It also never harms to have multiple options to phrase gherkin I suppose .. 
What do you think?

features/SPS007_Spatial-containment.feature
Ghesselink: Implemented https://github.com/buildingSMART/ifc-gherkin-rules/pull/197/commits/db3e95c81359ad81abbd55232c0213e4a60d6400 

Both` !=` and `is not `are now accepted in the step_impl, but we are using `is not`

is_a is changed to allow` Its Type` or `Its Entity Type` in the gherkin statement for a type lookup of an entity. 
We're currently using` its type`

` Given Its type is not IfcElementAssembly`

features/steps/givens/attributes.py
Ghesselink: Ad described above, implemented https://github.com/buildingSMART/ifc-gherkin-rules/pull/197/commits/db3e95c81359ad81abbd55232c0213e4a60d6400

features/steps/utils/vs_none.py
Ghesselink: That's indeed more descriptive https://github.com/buildingSMART/ifc-gherkin-rules/pull/197/commits/b635532c83191a5dc2d04f65af96648728b7f7e9

features/SPS007_Spatial-containment.feature
civilx64: Looks good; thanks

civilx64 (approved):

PR #199, opened 2024-06-03
Use inverses instead of file-wide type query. Use djange bulk_create() for outcomes

aothms:


aothms:
Before and after times on two models I investigated are:

`python3 manage.py shell`

```
import datetime
from apps.ifc_validation_models.models import ValidationRequest, ValidationTask, ValidationOutcome
for t in ValidationTask.objects.order_by('id').all():
  if t.type == 'NORMATIVE_IA' and t.started and t.ended and (t.ended - t.started) > datetime.timedelta(minutes=45):
    print(t.id, t.ended - t.started, t.request.file_name)
```

```
2740 0:52:50.441424 redacted
11933 3:56:40.085567 redacted
```

```
2859 0:19:11.402237 redacted
2866 0:26:36.108956 redacted
```

civilx64 (approved):

PR #200, opened 2024-06-03
mat001 - material set

Ghesselink:
Testing whether material layer has a correct name
https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Association/Material_Association/Material_Set/content.html#General-Usage


features/MAT001_Material-set.feature
```diff
@@ -0,0 +1,97 @@
...
+@version1
+@E00020
+Feature: MAT001 - Material Set
+  The rule verifies that correct names are used for material sets.
+
+  Scenario Outline: Verify material names for entities
+    Given an <entity>
+    Given its attribute HasAssociations
+    Given its attribute RelatingMaterial
+    Given Each relating material layer
```
aothms, line 12:
```suggestion
    Given Each related material component
```

I'm struggling to find the correct word, but layer is too specific. Because it also involves constituents or profiles.

features/MAT001_Material-set.feature
```diff
@@ -0,0 +1,97 @@
+@implementer-agreement
+@MAT
+@version1
+@E00020
+Feature: MAT001 - Material Set
+  The rule verifies that correct names are used for material sets.
```
aothms, line 6:
```suggestion
  The rule verifies that correct names are used for material sets. In this rule, material sets are the implicit generalization of IfcMaterialConstituentSet, IfcMaterialLayerSet, IfcMaterialProfileSet.
```

Material Set is not a known thing in the schema, and we kind of haphazardly came up with this generalized concept for 4.3 when Dion and I were on a deadline. So I think it's helpful to clarify a bit what this means.

features/steps/givens/relationships.py
```diff
@@ -61,7 +62,20 @@ def make_aggregate(val):
     if not instances and required:
         yield ValidationOutcome(instance_id=inst, severity=OutcomeSeverity.ERROR)
 
-    
+
+@gherkin_ifc.step("Each relating material layer")
+def step_impl(context, inst):
+    if inst.is_a("IfcMaterialLayerSet"):
```
aothms, line 68: You should also take into account the _Usage brothers of these material aggregates.

https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcMaterialLayerSetUsage.htm

https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcMaterialProfileSetUsage.htm

aothms (changes requested):

aothms:
This can at most be an industry practices rule. And I wonder how this affects non-english exchanges.

evandroAlfieri:
@Ghesselink as discussed, please put his on hold

PR #208, opened 2024-06-10
fix/ivs7-ifcopenshell-versioning

Ghesselink:
> being IfcOpenShell a dependency this is a reminder to make sure the VS uses the correct version…once updated (don’t think it’s an automated process, I’m sure we need to do something)

The automated check is currently placed at the beginning of the protocol errors, which means it runs for every local test. A few alternatives:

-  Completely manual: No automated check. Instead, manually run the script periodically (e.g., by executing `python3 update_ifcopenshell_versioning.py`). This approach avoids unnecessary API calls but we'll have to remember to execute this once in a while.

- Inside the Protocol Error: The automated check runs within the protocol error handling. However, this can be inefficient as it executes for multiple scenarios, quickly reaching the API rate limit.

- Within generate_markdown.py: This check would run each time a new rule is created. The downside is that it might not be tested regularly when we're not making new rules.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install behave pytest tabulate pyparsing sqlalchemy numpy pydantic pydot sqlalchemy_utils django python-dotenv deprecated pandas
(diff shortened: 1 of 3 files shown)


civilx64:
Changing to a new version of IFCOS should be a manual step, not automated.  We want to stick with a known, good commit and not always follow the latest / bleeding edge.

Closing this PR since the version is being manually updated in #220.

PR #217, opened 2024-06-14
BLT001-Correct use of operation type attributes for doors (IVS-9)

Ghesselink:
Informal specification

> IfcDoor.UserDefinedOperationType shall only be provided if the value of IfcDoor.OperationType is set to USERDEFINED

Additionaly: 

> The OperationType shall only be used, if no type object [IfcDoorType](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcDoorType.htm) is assigned, providing its own [IfcDoorType](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcDoorType.htm).OperationType.

Documentation reference
[6.1.3.16 IfcDoor - IFC 4.3.2 Documentation ](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcDoor.htm#6.1.3.16.3-Attributes)see the NOTE of the 12th attribute


features/BLT001_Correct-use-of-operation-type-attributes-for-doors.feature
```diff
@@ -0,0 +1,28 @@
+@implementer-agreement
+@BLT
+@version1
+@E00020
+
+Feature: BLT001 - Correct use of operation type attributes for doors
+
+  The rule verifies that attribuet UserDefinedOperationType is provided only when
```
civilx64, line 8:
```suggestion
  The rule verifies that attribute UserDefinedOperationType is provided only when 
```

test/files/blt001/fail-blt001-scenario02-door_type_own_operation_type.ifc
```diff
@@ -0,0 +1,32 @@
...
+#12=IFCDIMENSIONALEXPONENTS(0,0,0,0,0,0,0);
+#13=IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.);
+#14=IFCSIUNIT(*,.AREAUNIT.,$,.SQUARE_METRE.);
+#15=IFCSIUNIT(*,.VOLUMEUNIT.,$,.CUBIC_METRE.);
+#16=IFCSIUNIT(*,.PLANEANGLEUNIT.,$,.RADIAN.);
+#17=IFCMEASUREWITHUNIT(IFCPLANEANGLEMEASURE(0.017453292519943295),#16);
+#18=IFCCONVERSIONBASEDUNIT(#12,.PLANEANGLEUNIT.,'DEGREE',#17);
+#19=IFCUNITASSIGNMENT((#13,#14,#15,#18));
+#20=IFCPROJECT('2meTAdIsTAwOKSutMNpxgA',#5,'',$,$,$,$,(#11),#19);
+#21=IFCDOORTYPE('0AvhKGa3DC4O3_85l41YQd',$,$,$,$,$,$,$,$,$,$,$,$);
```
civilx64, line 28:
```suggestion
#21=IFCDOORTYPE('0AvhKGa3DC4O3_85l41YQd',$,$,$,$,$,$,$,$,$,.SLIDING_TO_RIGHT.,$,$);
```

Technically this is correct as a fail since there is `IfcRelDefinesByType` but let's add a type to `IfcDoorType` to make it "fully" wrong.

features/BLT001_Correct-use-of-operation-type-attributes-for-doors.feature
```diff
@@ -0,0 +1,28 @@
...
+  the value of the attribute OperationType is set to USERDEFINED.
+
+
+Background:
+  Given an IfcDoor
+
+
+Scenario: Correct values for OperationType and UserDefinedOperationType
+
+  Given UserDefinedOperationType = not empty
```
civilx64, line 18:
```suggestion
  Given UserDefinedOperationType is not empty
```

features/BLT001_Correct-use-of-operation-type-attributes-for-doors.feature
```diff
@@ -0,0 +1,28 @@
...
+  Given IsTypedBy = empty
```
civilx64, line 19:
```suggestion
  Given IsTypedBy is empty
```

For consistency

civilx64 (changes requested):

features/BLT001_Correct-use-of-operation-type-attributes-for-doors.feature
Ghesselink: I did this initially. However there is a caveat with the following lines 
```
    pred = operator.eq
    if value == 'empty':
        value = ()
    elif value == 'not empty':
        value = ()
```
When we say 'is not empty', `value` will be equal to '`empty` and not` 'not empty'` (the not part will is then parsed in the `comparison_to_op` part).

civilx64: Ok, sounds like there is a good reason to leave as is so we can drop my suggestion.

Ghesselink: I agree with you, it's more if you can intuitively write it down as` is not empty`

Tried to solve it, but there is an error later in the code as well. I don't find an easy standardized solution that covers all the cases that we have in our rules. It's probably not preferable at the moment to spend too much time on what is actually more a technical depth than writing rules.
In the meantime .. I left a note in case someone trips over this issue.

civilx64 (approved):

aothms:
@Ghesselink 

Sorry a bit late to the party. 

Two comments:

- The first scenario I think should also be applicable to IfcDoorType. I would do that as part of the same feature (which likely means removing the background).
- This same behaviour is present in IfcWindow.PartitioningType. Better start copy pasting while our memory is still fresh. [NEW RULES!!!! 🎉 ]

Issue #129, opened 2024-01-14
ALS011 - Alignment segment entity type consistency


civilx64:
Ref buildingSMART/IFC4.x-IF#137

- Same type must be used for all segments
- IfcSegmentedReferenceCurve/IfcGradientCurve requires IfcCurveSegment

PR #218, opened 2024-06-15
Implement ALS011 (IVS-10)

civilx64:
See discussion in buildingSMART/ifc-gherkin-rules#129.  For alignment representation curves:

- the same entity type must be used for all segments
- `IfcSegmentedReferenceCurve` and `IfcGradientCurve` must use `IfcCurveSegment` for all segments


Ghesselink:
It looks good :-) 
I saw the implementation in attributes.py was getting quite similar and noticed that we can probably make it easier for ourselves and reuse older statements. The missing link seems to be the Given statement 
`Its entity type`

For instance; 

```
The entity type of the Segments must be the same
```
Can also be written as 
```
Given its entity type 
Given its attribute Segments 
Then the values must be identical at depth 1
```

In this case, we're reusing the last two statements from previous rules. The same goes for
```
The type of attribute Segments must be IfcCurveSegments
```

```
Given its attribute Segments 
Given its entity type
Then the value must be "IfcCurveSegments"
```
Here, the first and the last statements already exist. 

I experimented a bit  in https://github.com/buildingSMART/ifc-gherkin-rules/pull/219
What do you think? 
Maybe it will also help with other alignment rules and relieve some pressure off the 
`@gherkin_ifc.step('The {attribute_or_type} of the {segment_type} must be {required_value}')`
implementation
(This implementation remained untouched in the experiment)

civilx64:
Thanks for this!  I considered reaching out once I had the first draft of a feature file as I suspected there might be a way to re-use existing functionality.  Let's go with your approach from #219.  I will close this PR.

PR #219, opened 2024-06-16
ALS011 - Alignment segment entity type consistency (IVS-10)

Ghesselink:
Alternative for ALS011
re-using already existing gherkin statements with an addition of `its entity type `gherkin statement. 
See  https://github.com/buildingSMART/ifc-gherkin-rules/pull/218

ping @civilx64 

Notes:

- Only new step implementation that is added is for the statement  `its entity type`
- The implementation of` the values must be identical` can be simplified because the previous issue with `GRF001` (None's were not counted) has been resolved due to the use of our custom error type.
- The error message is perhaps not so clear. We could probably fine-tune the `observed` column to actually display the values instead of '_Not Identical_'.  However, with `GRF001 `we probably have to refrain from this as the list of `values` is long.
image


civilx64 (commented):
> Maybe it will also help with other alignment rules and relieve some pressure off the
@gherkin_ifc.step('The {attribute_or_type} of the {segment_type} must be {required_value}')
implementation
(This implementation remained untouched in the experiment)

I couldn't tell if my edits to '/features/steps/thens/attributes.py` from #218 had been reverted or not.  Please do so prior to merging if necessary.

civilx64 (approved):

civilx64:
@Ghesselink another thought while working through the next rule... perhaps we should revise

`its entity type`

to

`its type`

to allow for things like `IfcLengthMeasure` which is an IFC type but not an entity instance.  Thoughts?

aothms (approved):

civilx64:
> For that reason I would not remove the word entity, so that we keep constantly disambiguating which type we are talking about. But if you can think of another word that more generally captures entities as well as defined types, I think that would be a good idea.

One idea that comes to mind is `class`.  That is a word that I see used in other industry discussions (and of course points to the C in IFC).

By "defined types" do you mean things like IfcLengthMeasure or IfcPileType?  Or possibly both?

In the meantime, agreed with keeping `its entity type`.

aothms:
> One idea that comes to mind is `class`. That is a word that I see used in other industry discussions (and of course points to the C in IFC).

`class` or `class name` sounds pretty good to me. Simple and obvious, but clear.

Issue #141, opened 2024-01-30
ALA010 - Informal proposition for IfcCurveSegment


evandroAlfieri:
Add a check for the Informal Proposition of [IfcCurveSegment](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcCurveSegment.htm), also reported below.

> For the foreseeable future, until implementer agreements have resulted in a defined parametric space for all possible types in attribute ParentCurve, the values for attributes SegmentStart and SegmentStart shall be of type IfcLengthMeasure.

See snippets below. The first three segments are ok, the fourth one (# 2481) shall trigger the error. 

```
/* GEOMETRY OF HORIZONTAL PROFILE (composite curve) */
#2449 = IFCCOMPOSITECURVE((#2450, #2461, #2471, #2481), .U.);
/* SHAPE REPRESENTATION OF ALL SEGMENTS OF THE HORIZONTAL PROFILE (making the composite curve) */
#2450 = IFCCURVESEGMENT(.CONTSAMEGRADIENTSAMECURVATURE., #2453, IFCLENGTHMEASURE(0.), IFCLENGTHMEASURE(387.723276296965), #2454);
#2451 = IFCCARTESIANPOINT((452270.188250964, 4539403.94736217));
#2452 = IFCDIRECTION((9.39398720423267E-1, 3.42826550995001E-1));
#2453 = IFCAXIS2PLACEMENT2D(#2451, #2452);
#2454 = IFCLINE(#2455, #2456);
#2455 = IFCCARTESIANPOINT((0., 0.));
#2456 = IFCVECTOR(#2457, 1.);
#2457 = IFCDIRECTION((9.39398720423267E-1, 3.42826550995001E-1));
#2458 = IFCPRODUCTDEFINITIONSHAPE($, $, (#2460));
#2459 = IFCGEOMETRICREPRESENTATIONSUBCONTEXT('Axis', 'Model', *, *, *, *, #13, $, .MODEL_VIEW., $);
#2460 = IFCSHAPEREPRESENTATION(#2459, 'Axis', 'Segment', (#2450));
#2461 = IFCCURVESEGMENT(.CONTSAMEGRADIENTSAMECURVATURE., #2464, IFCLENGTHMEASURE(0.), IFCLENGTHMEASURE(39.9999999999925), #2465);
#2462 = IFCCARTESIANPOINT((452634.4150006, 4539536.86919573));
[...]

PR #220, opened 2024-06-17
Implement ALS012 (IVS-11)

civilx64:
Per #141 check the alignment segments in IfcCompositeCurve, IfcGradientCurve, and IfcSegmentedReferenceCurve for the use of IfcLengthMeasure type for attributes SegmentStart and SegmentLength.


features/ALS012_Alignment-segment-start-and-length-attribute-types.feature
```diff
@@ -0,0 +1,25 @@
...
+
+Scenario Outline: Correct entity type used for SegmentStart and Segment Length
+  Given an <entity>
+  Given its attribute Segments
+
+  Then The type of attribute <attribute> must be IfcLengthMeasure
+
+  Examples:
+    | entity                      | attribute     |
+    | IfcCompositeCurve           | SegmentStart  |
```
aothms, line 20: I think this works because in "The type of attribute ..." impl we do `getattr(i, attribute, None)` so if the attribute does not exist it doesn't raise an exception and the `None`s are later filtered out.

But! It probably is necessary though to implement a bit more formally in the gherkin logic that this only (a) applies to alignment representations or (b) only involves segments of type IfcCurveSegment and not IfcCompositeCurveSegment (the most confusing naming distinction ever!! still upset about this 😠 ). Because what would happen otherwise is that even a non-suspecting IFC2X3 file can get green checkmarks on alignment related rules.

aothms (changes requested):

Ghesselink (commented):
> Let's see in dev how the link in the description is rendered. I remember @Ghesselink having an issue with that

@evandroAlfieri 
Should be fine. I initially thought it had something to do with the link, but it had to do with the parsing of `'Scenario Outline:'` instead. 
We were looking for either` Scenario:` or `Background:` to parse the feature file in the validation platform. In the case of Scenario Outline:, one of these two is never reached and therefore the whole page was rendered along. But this issue is solved now.

features/ALS012_Alignment-segment-start-and-length-attribute-types.feature
civilx64: I added a few givens to ensure green checks only for Segments that are part of an alignment representation in a 4.3 model.  I also added `without subtypes` to address @Ghesselink 's comments.

aothms:
```
  Given An IfcAlignment
  Given Its attribute Representation
  Given Its attribute Representations
  Given an <entity> without subtypes
```

Thanks, but this only partially solves the problem. Because with a new `Given an {entity}` there is no correspondence with the lines above. I haven't checked how it would work in practice, but I guess it would just overwrite the context, which still means that IfcCompositeCurves not part of an IfcAlignment would be subsequently processed (as long as there is at least a single IfcAlignment in the model, not even sure about that last part).

What I think we could do is use the ifcopenshell.file.traverse() method.

```
  Given An IfcAlignment
  Given Its attribute Representation
  Given Its attribute Representations
  Given All refererenced instances
  Given The instances with class name <entity>
```

`All refererenced instances` would do file.traverse() on the previous inst in the stack. `The instances with class name <entity>` would filter those based on entity name. That way we can basically arbitrarily jump in the graph to a location of interest. Or - what we're trying to solve here - maintain that there is a relationship between the first Given and the instance of type {entity} we're investigating.

aothms (changes requested):

civilx64:
> What I think we could do is use the ifcopenshell.file.traverse() method.

Latest commit incorporates this into a step implementation.  I also added a second passing test file that contains an additional composite curve with segment start and end defined by `IfcParameterValue` used for solid extrusion of a column.  This second composite curve is correctly not flagged as an error since it is not part of an `IfcAlignment`.

features/ALS012_Alignment-segment-start-and-length-attribute-types.feature
civilx64: > What I think we could do is use the ifcopenshell.file.traverse() method.

Latest commit incorporates this into a step implementation.  I also added a second passing test file that contains an additional composite curve with segment start and end defined by `IfcParameterValue` used for solid extrusion of a column.  This second composite curve is correctly not flagged as an error since it is not part of an `IfcAlignment`.

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install behave pytest tabulate pyparsing sqlalchemy numpy pydantic pydot sqlalchemy_utils django python-dotenv deprecated pandas
-        wget -O /tmp/ifcopenshell_python.zip https://s3.amazonaws.com/ifcopenshell-builds/ifcopenshell-python-`python3 -c 'import sys;print("".join(map(str, sys.version_info[0:2])))'`-v0.7.0-f7c03db-linux64.zip
+        wget -O /tmp/ifcopenshell_python.zip https://s3.amazonaws.com/ifcopenshell-builds/ifcopenshell-python-`python3 -c 'import sys;print("".join(map(str, sys.version_info[0:2])))'`-v0.8.0-90ae709-linux64.zip
```
aothms, line 31: Just a note that we really need to be careful deploying this, making sure everything is well tested.

features/steps/givens/entities.py
```diff
@@ -58,4 +58,10 @@ def step_impl(context, inst, relationship_direction):
...
         attr_value = NullAttribute
-    yield ValidationOutcome(instance_id = attr_value, severity = OutcomeSeverity.PASSED)
\ No newline at end of file
+    yield ValidationOutcome(instance_id = attr_value, severity = OutcomeSeverity.PASSED)
+
+
+@gherkin_ifc.step("All referenced instances")
+def step_impl(context, inst):
+    instances = context.model.traverse(inst)
+    yield ValidationOutcome(instance_id=instances, severity=OutcomeSeverity.PASSED)
```
aothms, line 67: @Ghesselink In other occasions I see:

```
for inst in instances:
        yield ValidationOutcome(instance_id = inst, severity = OutcomeSeverity.PASSED)
```

Does it matter?

features/steps/givens/entities.py
```diff
@@ -58,4 +58,10 @@ def step_impl(context, inst, relationship_direction):
...
     if attr_value is None:
         attr_value = NullAttribute
-    yield ValidationOutcome(instance_id = attr_value, severity = OutcomeSeverity.PASSED)
\ No newline at end of file
+    yield ValidationOutcome(instance_id = attr_value, severity = OutcomeSeverity.PASSED)
+
+
+@gherkin_ifc.step("All referenced instances")
+def step_impl(context, inst):
+    instances = context.model.traverse(inst)
```
aothms, line 66:
```suggestion
    # Note that this includes `inst` as the first element in this list
    instances = context.model.traverse(inst)
```

Just one word of caution

aothms:
Super wonderful. Ready to merge, accept for just this one clarification I'd like from @Ghesselink

features/steps/givens/entities.py
Ghesselink: Yes, unfortunately. For reasons that are perhaps a bit complex.
In the occasions that you mention, the way we handle the outputted results from the step_impl is.
https://github.com/buildingSMART/ifc-gherkin-rules/blob/fa4b6e2f6335147fabf50813491a2fda795af80f/features/steps/validation_handling.py#L95
Note that we move the entire output list to the context stack. We do this solely to determine the initial set of activation instances. If we don't, i.e., if a statement executes a function with one of the activation instances as an argument (as is the case for any function following `Given an IfcEntity`), we move only the first element of the list to the context stack. 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/fa4b6e2f6335147fabf50813491a2fda795af80f/features/steps/validation_handling.py#L61

Note the `[0]`. This works in most cases (in fact, in all cases so until now) because the output list has a length of 1, containing a single yielded entity instance (or value) resulting from step_impl. I admit it's not ideal, but it could handle the complexity of irregular nested lists. 
In case we do the same for the current statement (`for inst in instances; yield ValidationOutcome`), this [0] indicates that we will take the first element of context.model.traverse(inst). This includes inst as the first element in this list, meaning that context.instances will remain exactly the same.

aothms: Clear. Thanks for clarifying, but indeed something to look out for.

aothms (approved):

features/ALS012_Alignment-segment-start-and-length-attribute-types.feature
aothms: > Another note; aren't IfcSegmentedReferenceCurve and IfcGradientCurve subtypes of IfcCompositeCurve?

Excellent comment btw, but it is addressed in the new structure because `Its Entity Type` does not take into account inheritance because it takes the entity types as strings and pushes these to the next stack layer.

https://github.com/buildingSMART/ifc-gherkin-rules/blob/579e66b94b5bb7270c6dfa60b5c7a305b92dc9bc/features/steps/givens/entities.py#L52-L57

PR #224, opened 2024-06-18
Fix ALA003 so that cant layout segment .LINEARTRANSITION. corresponds to representation by IfcClothoid (IVS-17)

### LINKED ISSUE #213: ALA003 - Maybe wrong with Cant
In an attempt to update the file from https://github.com/buildingSMART/IFC4.x-IF/issues/151 to pass the latest bSI Validation Checks, we noticed it fails on the ALA003 rule claiming that since the Semantic part of the Cant is using LinearTransition, the Geometry should use Lines and not Clothoids as per in the file.

This may be incorrect, leaving the floor to @peterrdf for the technicalities and pinging @evandroAlfieri and @aothms as it is not possible to set assignees.

Also, attaching the current version of the file for testing.

[ACCA_sleepers-linear-placement-cant-implicit.zip](https://github.com/user-attachments/files/15816340/ACCA_sleepers-linear-placement-cant-implicit.zip)

civilx64:
Fixes #213


evandroAlfieri (approved):
Lgtm

aothms (approved):

civilx64:
Add cross ref to buildingSMART/IFC4.x-IF/issues/145

PR #225, opened 2024-06-21
Fix/IVS-40 OJT001 change str attribute not found to empty tuple

Ghesselink:
After looking at one file, IfcMemberType had an attribute ‘USERDEFINED’ which implies that the predefinedType of IfcMember must be empty. This is the case, however we get an ‘attribute not found’, which is logically not the same as an empty value ()

 `attribute_value = getattr(inst, attribute, 'Attribute not found')`
Which will give a false negative with 
`elif not pred(attribute_value, value)`

Simple solution is therefore to simply change these lines  to 

` attribute_value = getattr(inst, attribute, ())`


features/steps/thens/attributes.py
```diff
@@ -78,8 +85,10 @@ def step_impl(context, inst, attribute, value):
         # nothing was activated by the Given criteria
         yield ValidationOutcome(inst=inst, severity=OutcomeSeverity.EXECUTED)
     elif not pred(attribute_value, value):
-        yield ValidationOutcome(inst=inst, expected= None if not value else value, observed=misc.recursive_unpack_value(attribute_value), severity=OutcomeSeverity.ERROR)
-
+        observed = misc.recursive_unpack_value(attribute_value)
+        if display_entity:
+            observed = f"{inst.is_a()}(#{inst.id()}); {attribute} = {observed}"
```
aothms, line 90: I'm not sure if I'm in favour of this.

We have `inst` already in `inst=inst`. I think we should make the formatting in the table more intelligent rather than adding this logic in every step impl function. Can this just be removed for now and solved in the UI later?

Also is it still necessary to do recursive_unpack_value()?

aothms (changes requested):

features/steps/thens/attributes.py
civilx64: My $0.02... the issue is that the results persisted to the DB are similar enough that they get collapsed into a single result by the de-duplication code in the front end.  I agree with this approach to add logic in `gherkin_rules` to make DB rows more descriptive.  The additional `and display entity instance` would not needed to be added to every impl function, but only used in this current unique case.

The other alternative is to adjust the feature file to have IfcObject as the activation instance instead of IfcTypeObject.  The graph traversal gets more complicated in that case but maybe it is a reasonable tradeoff to avoid any gymnastics in step implementation and UI front end.

features/OJT001_Object-predefined-type.feature
civilx64, line 1: need to bump version number

civilx64 (commented):
See note about version number in feature file.

features/steps/thens/attributes.py
civilx64: Update: adjusting the feature file does not solve the issue.  It does avoid de-duplication but `Expected` and `Observed` are still confusing.

Using this revision to the 3rd scenario...

```gherkin
Scenario: Validating the proper use of PredefinedType for an Object typed by IfcTypeObject
  Given An IfcObject with subtypes
  Given PredefinedType = not empty
  Given IsTypedBy = not empty
  Given its attribute IsTypedBy
  Given its attribute RelatingType
  Then The value of attribute PredefinedType must be NOTDEFINED
````

...gives this as output:

image

Which is correctly flagging the four failing entities but the messaging is still confusing.  These entities fail because each `IfcFooting` has `PredefinedType = .NOTDEFINED.` (not empty) and the `IfcFootingType` has `PredefinedType = .PILE_CAP.` 
For this model to be valid, `PredefinedType` should be empty for each `IfcFooting`.

The model would also technically be valid with `PredefinedType = .NOTDEFINED.` for the `IfcFootingType` and `PredefinedType = .PILE_CAP.` for each `IfcFooting` but would not be considered best practice as the type now has to be specified at each occurrence.  Which means that the current version of the gherkin feature file is superior to this revision above.

> We have inst already in inst=inst. I think we should make the formatting in the table more intelligent rather than adding this logic in every step impl function. Can this just be removed for now and solved in the UI later?

The issue is that this validation check passes or fails based upon the logic of the IfcTypeObject as activation instance, but we need to display the associated IfcObjects connected via the IfcRelDefinesByType.  Therefore I'm in favor of the current commit providing flexibility to dictate what is reported in the UI via the step implementation.

aothms: > The issue is that this validation check passes or fails based upon the logic of the IfcTypeObject as activation instance, but we need to display the associated IfcObjects connected via the IfcRelDefinesByType. Therefore I'm in favor of the current commit that provides flexibility to dictate what is reported in the UI via the step implementation.

Thanks for clarifying, but then still I'd opt for a more general solution:

Why don't we here in the decorator, look up whether we have `"and display entity instance"` in the `step.name` and if so don't use `activation_inst`, but just `inst` as the main instance in the report?

https://github.com/buildingSMART/ifc-gherkin-rules/blob/b7e2a68c605f9017028665cb00a770ff38a02a3b/features/steps/validation_handling.py#L148

aothms: Because ultimately I'd really like to keep the step impl functions conceptually as much as possible as pure function functions returning {True False} without too many side effects and have the presentations logic in the "decorator".

civilx64: Good point.  I'll take a shot and see if we can close this PR out.

civilx64: Got it... example for [fail-ojt001-scenario03-typed_via_relation_and_at_occurrence.ifc](https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/test/files/ojt001/fail-ojt001-scenario03-typed_via_relation_and_at_occurrence.ifc)

image

I was close with the last commit but had `inst` and `activation_inst` backwards.

civilx64 (approved):

PR #226, opened 2024-06-24
GEM051 - Presence of geometric context (IVS-43)

Ghesselink:
> I completely agree that the bareminimum should be that there should be:
> 
>  1. A minimum of at least one contex


aothms (changes requested):
The problem with the rule as it is written is that even when the context can exist in the model it might still not be associated to the project.

I would change it like:

```
Given an IfcProject
Then RepresentationContexts shall not be empty
Then type of attribute RepresentationContexts  shall be IfcGeometricRepresentationContext
```

The complexity here is that in IFC4 and higher we should not use IfcProject, but IfcContext (the supertype), so we need multiple scenarios with a schema wide applicability.

Ghesselink:
@aothms 

I was already wondering if there would be a rule that simple in IFC.
Schema-wide applicability are added in https://github.com/buildingSMART/ifc-gherkin-rules/pull/226/commits/d727319a4afad6c630fa9c28625379f2817dc43e

> regarding shall be IfcGeometricRepresentationContext it can also be its subtype so inheritance should be consider, just not its supertype IfcRepresentationContext which is the default type of the attribute.

I think this is covered by 
`Its entity type is 'IfcGeometricRepresentationContext including subtypes'`

IfcRepresentationContext here is False, IfcGeometricRepresentationSubContext is True. Added another testfile to check this 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/226/commits/aaf71fe7d758b8c1bac93c2889fd9a77d10eb6cd

features/steps/thens/existance.py
```diff
@@ -47,4 +47,26 @@ def step_impl(context, inst):
...
+    for item in lst:
+        if isinstance(item, (tuple, list)):
+            flattened_list.extend(recursive_flatten(item))
+        else:
+            flattened_list.append(item)
+    return flattened_list
+
+def get_previous_step_before_assertion(context):
+    for i, step in enumerate(context.scenario.steps):
+        if step.name == 'assert existence' and i > 0:
```
aothms, line 63: Oh this is nice. Well done, just small comment that I would not hardcode the step name since anyway we have it.

```suggestion
        if step.name.lower() == context.step.name.lower() and i > 0:
```

civilx64: incorporated into `27347b`

civilx64 (commented):
test file names needed a little housekeeping so I hopped in to do that and to commit the review suggestion.  Should be good to merge now.

civilx64 (approved):

PR #227, opened 2024-06-24
IVS-44 GEM052 Correct geometric subcontexts

Ghesselink:
> A minimum of at least one subcontext per context. The context
>  identifier should be one of the agreed values:
>          [
>             "Annotation",
>             "Axis",
>             "Box",
>             "FootPrint",
>             "Reference",
>             "Body",
>             "Clearance",
>             "CoG",
>             "Profile",
>             "SurveyPoints",
>             "Lighting",
>         ]

Further added a clearer error message in the case of 'not empty'. 
`Instead of expected = (), observed = ()`
we're displaying 
`expected = 'not empty', observed = 'empty'`


features/GEM052_Correct-geometric-subcontexts.feature
```diff
@@ -0,0 +1,20 @@
...
+    
+    Given an IfcGeometricRepresentationContext without subtypes
+    
+    Then HasSubContexts = not empty
+
+
+Scenario: Constraints on context type
+    Given An IfcGeometricRepresentationSubContext
+
+    Then ContextIdentifier = 'Annotation' or 'Axis' or 'Box' or 'Footprint' or 'Reference' or 'Body' or 'Clearance' or 'CoG' or 'Profile' or 'SurveyPoints' or 'Lightning'
```
civilx64, line 19:
```suggestion
    Then ContextIdentifier = 'Annotation' or 'Axis' or 'Box' or 'Footprint' or 'Reference' or 'Body' or 'Clearance' or 'CoG' or 'Profile' or 'Lighting'
```

I don't see `SurveyPoints` in the 4.3 docs.  Was it valid for an earlier schema version?  Per [CT 4.1.7.1.2.3](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Geometric_Representation/Annotation_Geometry/Set_Of_Survey_Points/content.html) it would fall under `Annotation`.

features/GEM052_Correct-geometric-subcontexts.feature
```diff
@@ -0,0 +1,20 @@
+@implementer-agreement
+@GEM
+@version1
+@E00050
+Feature: GEM052 - Correct geometric subcontexts
+The rule verifies that there is a minimum of at least one subcontext per context. The context
```
civilx64, line 6:
```suggestion
The rule verifies that there is a minimum of at least one subcontext per context. 
Reference: https://github.com/buildingSMART/Sample-Test-Files/issues/137.
The context
```

A reference to this discussion would be helpful.  As you noted in the JIRA ticket for GEM050 there is nothing documented in the spec for this.

Which begs the question - shouldn't the requirement for subcontext be a best practice warning and not a validation fail?

features/GEM052_Correct-geometric-subcontexts.feature
```diff
@@ -0,0 +1,22 @@
...
+    
+    Given an IfcGeometricRepresentationContext without subtypes
+    
+    Then HasSubContexts = not empty
+
+
+Scenario: Constraints on context type
+    Given An IfcGeometricRepresentationSubContext
+
+    Then ContextIdentifier = 'Annotation' or 'Axis' or 'Box' or 'Footprint' or 'Reference' or 'Body' or 'Clearance' or 'CoG' or 'Profile' or 'Lighting'
```
aothms, line 21: I think we should use the CSVs we already have https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/features/resources/valid_RepresentationIdentifier_IFC4.3.csv Not sure what that means for 2X3 for which we don't have a CSV. I guess leave it unconstrained for 2X3.

aothms (changes requested):

civilx64 (approved):

PR #229, opened 2024-06-28
Initial commit of self-intersection check - IVS-45

aothms:
This turned out quite a bit more elaborate than expected, but have a first look and let me know where you have questions.


.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         echo $CONDA/bin >> $GITHUB_PATH
     - name: Install dependencies
       run: |
-        pip install behave pytest tabulate pyparsing sqlalchemy numpy pydantic pydot sqlalchemy_utils django python-dotenv deprecated pandas
+        pip install behave pytest tabulate pyparsing sqlalchemy numpy pydantic pydot sqlalchemy_utils django python-dotenv deprecated pandas rtree
```
aothms, line 30: Reminder to self to update dependencies also on val service

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
...
         wget -O /tmp/ifcopenshell_python.zip https://s3.amazonaws.com/ifcopenshell-builds/ifcopenshell-python-`python3 -c 'import sys;print("".join(map(str, sys.version_info[0:2])))'`-v0.8.0-90ae709-linux64.zip
```
aothms, line 31: Probably doesn't work with this version of v0.8.0. Started a new build.

civilx64: LMK commit hash / link when ready please so that I can do an in-depth review & step through the implementation code.

test/files/gem010/images.png
civilx64, line 1: This is great but would be even better with a title and "pass / fail" indicator for each sketch.

test/files/gem010/plot.py
```diff
@@ -0,0 +1,21 @@
...
+    st = ifcopenshell.geom.settings(DIMENSIONALITY=ifcopenshell.ifcopenshell_wrapper.CURVES)
+
+shp = ifcopenshell.geom.create_shape(st, p)
+print(ifcopenshell.get_log())
+vs = numpy.array(shp.verts).reshape((-1, 3))
+ed = numpy.array(shp.edges).reshape((-1, 2))[:,0]
+ed = numpy.concatenate((ed, [ed[0]]))
+cs = vs[ed][:,0:2]
+plt.plot(*cs.T)
+try:
```
civilx64, line 18: Here's where you could add plt.Title per earlier comment

civilx64 (commented):
I don't see a *.feature file.  Did I miss it or is this a general template that will result in a series of GEM rules?

aothms:
> I don't see a *.feature file. Did I miss it or is this a general template that will result in a series of GEM rules?

@civilx64  1000x apologies. Added it.

aothms:
Note that the feature file and test files still need to be broken up per what we conclude.

aothms:
@civilx64 I added a new joke called 'full_stack_rule' which means that in addition to the `inst` parameter you get a `path` (list) of all previous givens that trace back `inst` to the first given statement. This is necessary for the tesselated faces because a tesselated face is just a list of integer indices, and the vertex coordinate positions are defined one level higher.

https://github.com/buildingSMART/ifc-gherkin-rules/pull/229/commits/e64c55104f7478e81038e5862ff3812d93a409ae#diff-2c838ec4f24dc4e43b898ddaa7de4de8f6cc972099850db514753312625e105a

features/GEM010_ArbitraryProfileDef-intersections.feature
```diff
@@ -0,0 +1,45 @@
+@informal-proposition
+@GEM
+@version1
+@E00050
+Feature: GEM010 - ArbitraryProfileDef intersections
+The rule verifies that IfCArbitraryClosedProfileDefs and IfcArbitraryProfileDefWithVoids do
```
civilx64, line 6:
```suggestion
The rule verifies that IfcArbitraryClosedProfileDefs and IfcArbitraryProfileDefWithVoids do
```

features/GEM010_ArbitraryProfileDef-intersections.feature
```diff
@@ -0,0 +1,45 @@
...
+    Then There must be no self-intersections
+
+  Scenario: IfcArbitraryProfileDefWithVoids
+
+    Given An IfcArbitraryClosedProfileDef
+    Given Its attribute InnerCurves
+
+    Then There must be no self-intersections
+
+  Scenario: Polyhedral IfcFace
```
civilx64, line 23: It feels like this would be a natural place to split into a separate *.feature file.

features/steps/thens/geometry.py
```diff
@@ -36,79 +38,112 @@ def step_impl(context, inst):
...
+    for edges in bounds:
+        for i, edge in enumerate(edges):
+            if isinstance(edge, numpy.ndarray):
+                ps = edge
             else:
-                qs = numpy.array([
-                    edges[j].start.coords,
-                    edges[j].end.coords
+                if edge.basis:
+                    raise NotImplementedError()
```
civilx64, line 100: I seem to remember from earlier discussions that raising exceptions was problematic in terms of properly tracking the validation outcome and severity.

features/steps/validation_handling.py
```diff
@@ -207,12 +225,15 @@ def wrapped_step(func):
 
 def execute_step(fn):
     is_global_rule = False
+    is_full_stack_rule = False
     while hasattr(fn, '__wrapped__'): # unwrap the function if it is wrapped by a decorator in casse of catching multiple string platterns
```
civilx64, line 229:
```suggestion
    while hasattr(fn, '__wrapped__'): # unwrap the function if it is wrapped by a decorator in case of catching multiple string platterns
```

civilx64 (commented):
Everything here seems logical and reasonable.  I recommend moving forward with breakout into individual rules.  I believe this will become GEM010 and GEM011.

PR #231, opened 2024-07-04
GEM001 - Consider bound orientation and loop reuse (IVS-52)

aothms:


test/files/gem001/generate.py
```diff
@@ -0,0 +1,60 @@
...
+    [1.0, 1.0, -1.0], 
+    [1.0, 1.0, 1.0], 
+    [-1.0, 0.5, -0.5], 
+    [-1.0, -0.5, -0.5], 
+    [-1.0, -0.5, 0.5], 
+    [-1.0, 0.5, 0.5]]
+
+vs = numpy.array(vs)
+
+# Forgive me, these are 1-based to comply with the express-based indices of tesselated items,
```
civilx64, line 21: I forgive you

test/files/gem001/generate.py
```diff
@@ -0,0 +1,60 @@
...
+                    f.createIfcFaceOuterBound(*create_bound_args(outer)),
+                    *[f.createIfcFaceBound(*create_bound_args(ib)) for ib in inner]
+                ])
+            case [outer]:
+                yield f.createIfcFace([
+                    f.createIfcFaceOuterBound(*create_bound_args(outer))
+                ])
+
+f.createIfcClosedShell(list(yield_faces()))
+f.write('pass-bound-orientation.ifc')
```
civilx64, line 60:
```suggestion
f.write('pass-gem001-bound_orientation.ifc')
```

test/files/gem001/pass-bound-orientation.ifc
```diff
@@ -0,0 +1,60 @@
+ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION(('ViewDefinition [CoordinationView]'),'2;1');
+FILE_NAME('','2024-07-04T11:23:15',(),(),'IfcOpenShell v0.7.0-f7c03db75','IfcOpenShell v0.7.0-f7c03db75','');
```
civilx64, line 4: Probably should update your local environment to 0.8.0, no?

features/steps/utils/geometry.py
civilx64, line 1: The diff view is a little confusing, possibly because I'm reviewing this part of the codebase for the first time.  It looks like some of this has been around for some time but remains commented out.  If so can we just delete it?

civilx64 (changes requested):
# General comments in addition to those on specific files / lines

## Required changes:
- feature file version needs to be bumped
- need to rename the new passing model

## Optional, nice to have:
- need additional fail files - at least one per scenario
  - scenario number should be added to the file name

test/files/gem001/generate.py
Ghesselink: This was actually the cause of the failing CI/CD. Seems like the protocol checks are working, at least :-) 
Changed it in https://github.com/buildingSMART/ifc-gherkin-rules/pull/231/commits/d52e8b2b1a12bf99c699ec92dc73aacac7e4d590

Ghesselink:
I took the liberty to commit a couple of additional changes; 

- [Include the rule-code in the name of the testfile](https://github.com/buildingSMART/ifc-gherkin-rules/pull/231/commits/d52e8b2b1a12bf99c699ec92dc73aacac7e4d590) (this was causing a error for the protocol check and, as a consequence, a failing CI/CD)
- [Split the scenario in the feature file in scenario outlines](https://github.com/buildingSMART/ifc-gherkin-rules/pull/231/commits/2b0621e78fd861847e2e965700f18d1f676a1784). While creating a test file, I noticed that if a `Then` step in a scenario fails, the subsequent `Then` condition isn't validated. If a file contains errors in both edges and oriented edges, both issues need to be reported.

Changed
```
Given an Entity
Then Condition 1 
Then Condition 2 -> does not run in case [1] yields an error
```

To

```
Given an IfcEntity
Then <stmt>
Examples:
    | stmt | 
    | condition 1| 
    | condition 2| -> always runs
```

Which is the same as 
```
Given an IfcEntity
Then Condition 1

Given an IfcEntity
Given Condition 2 -> Always runs, even when Condition 1 yields an error
```

aothms:
Nice find, but not sure if I'm very happy with the approach of writing entire statements in the outline examples.

I would rather propose

```
Scenario outline
  Given <entity>
  Then     Then Every edge must be referenced exactly ...
Examples:
 Ifc...
 Ifc..


Scenario outline
  Given <entity>
  Then     Every oriented edge must be referenced exactly ...
Examples:
 Ifc...
 Ifc..
```

features/steps/utils/geometry.py
```diff
@@ -12,31 +12,49 @@ def get_edges(file, inst, sequence_type=frozenset, oriented=False):
...
+                for ed in bnd.Bound.EdgeList:
+                    # @todo take into account edge geometry
+                    # edge_geom = ed[2].EdgeGeometry.get_info(recursive=True, include_identifier=False)
+                    
+                    coords = [
+                        ed.EdgeElement.EdgeStart.VertexGeometry.Coordinates,
+                        ed.EdgeElement.EdgeEnd.VertexGeometry.Coordinates,
+                    ]
+                    
+                    # @todo verify:
```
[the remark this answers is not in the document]
aothms: Not before deploying. Best seen as a note for future other.

test/files/gem001/pass-bound-orientation.ifc
aothms: haha, I'm even slow updating my own software

aothms:
@civilx64 manditory changes have been made (one by me, one by geert)

civilx64 (approved):
A few comments to close this out:

1. This PR should be revised to merge onto `devlopment`, not `main`.
2. Unit test file should be `pass-gem001-bound_orientation.ifc` with an underscore as final separator.
3. README in the unit test folder should be updated to include the additional test file.

aothms:
Thanks @civilx64

PR #237, opened 2024-07-16
GEM051 & 052; NR to IPB

Ghesselink:
There is no source somewhere in the documentation, therefore the rule is initially a best practice


evandroAlfieri (approved):
Agree

PR #238, opened 2024-07-17
IVS-43 - Fix/update error message GEM051

Ghesselink:
We're performing attribute lookups for both Given as well as Then statements.  `Observed` was calculated after a matching expected entity type and the entity type found for the value. This is however not sufficient if we check for mismatching entity types. In this case, we want to display the mismatching entity type and therefore have to save this before the actual matching (as everything after the actual matching will not be reached; it is incorrect). 

This will display the false entity type for GEM051
image

### DIFF
--- a/features/steps/validation_handling.py
+++ b/features/steps/validation_handling.py
@@ -133,6 +133,9 @@ def handle_then(context, fn, **kwargs):
                 return
             top_level_index = current_path[0] if current_path else None
             activation_inst = inst if not current_path or activation_instances[top_level_index] is None else activation_instances[top_level_index]
+
+            if "GEM051" in context.feature.name and context.is_global_rule:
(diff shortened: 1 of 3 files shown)


features/steps/validation_handling.py
```diff
@@ -133,6 +133,9 @@ def apply_then_operation(fn, inst, context, current_path, depth=0, **kwargs):
                 return
             top_level_index = current_path[0] if current_path else None
             activation_inst = inst if not current_path or activation_instances[top_level_index] is None else activation_instances[top_level_index]
+
```
civilx64, line 136:
```suggestion
            #TODO: refactor into a more general solution that works for all rules
```

civilx64 (approved):

PR #239, opened 2024-07-18
IVS-59 wip/activation of functional parts

Ghesselink:
A ‘dummy’ rule with a ‘000’ number can be used to solely determine whether the rule is activated or not, and therefore turn the scorecard to either grey or green.

Added two new, simple rules and a step implementation to test this.

A challenge would be to activate based on two entities being both present within a file, e.g.
`IfcDistributionPort AND IfcRelConnectsPorts` (POR)


test/files/cls000/pass-cls000-IfcClassificationReference_present.ifc
civilx64, line 1: I don't see a classification reference in this file.  Which means the unit test should fail because it does not emit a passing result.  However the unit tests are looking solely at whether or not an error is generated.

Similarly the web UI is gray, not green, for this test file:
image

test/files/spa000/pass-spa000-activated_space_present.ifc
civilx64, line 1: This file does not activate the rule:
image

features/resources/functional_parts.csv
civilx64, line 1: Add CLS

civilx64 (changes requested):
Generally speaking the concept seems sound.  You will need to test against a local version of the Validation Service as the gherkin unit tests alone are not sufficient to confirm that a rule is being activated.

features/SPA000_Spaces-information.feature
```diff
@@ -0,0 +1,19 @@
+@implementer-agreement
+@SPA
+@version1
+@E00020
+
+Feature: SPA000 - Spaces Information
+    The rule verifies that there is an ability to model spaces, such as rooms, hallways, clearance zones, and circulation areas.
```
evandroAlfieri, line 7:
```suggestion
    The rule verifies the presence of IFC entities used to model spaces, such as rooms, hallways, clearance zones, and circulation areas.
```

evandroAlfieri (changes requested):

features/steps/thens/existance.py
```diff
@@ -69,4 +69,10 @@ def step_impl(context, inst):
...
         expected = get_previous_step_before_assertion(context)
-        yield ValidationOutcome(instance_id=inst, expected = expected, observed='Nonexistent', severity = OutcomeSeverity.ERROR)
\ No newline at end of file
+        yield ValidationOutcome(instance_id=inst, expected = expected, observed='Nonexistent', severity = OutcomeSeverity.ERROR)
+
+@gherkin_ifc.step("The IFC model contains information on {functional_part_description}")
+def step_impl(context, inst, functional_part_description):
+    # This rule always passes and is used solely to trigger activation of the the rule if an instance linked to the functional part is present
+    if False:
+        yield ValidationOutcome(inst=inst, severity=OutcomeSeverity.ERROR)
```
aothms, line 78:
```suggestion
    pass
```

I would just write it like this. (Otherwise for example you trigger code quality warnings about constant boolean expressions and unreachable code).

test/files/cls000/pass-cls000-IfcClassificationReference_present.ifc
aothms: Maybe I'm mistaken but I think this is the intention. This kind of rule by definition can only differentiate between Executed and Passed. Gray vs Green.

features/CLS000_Classification-association.feature
```diff
@@ -0,0 +1,15 @@
...
+@version1
+@E00020
+
+Feature: CLS000 - Classification Association
+    The rule verifies that there is an ability to classify elements, materials, and systems according to various classification systems,
+    such as the UNIFORMAT or Omniclass classification systems.
+
+    Scenario: Check for activation of Classification Association
+
+    Given an IfcClassificationReference
```
aothms, line 12:
```suggestion
    Given an IfcObjectDefinition
    Given its attribute HasAssociations
    Given its entity type is IfcRelAssociatesClassification
    Given its attribute RelatingClassification
    Given its entity type is IfcClassificationReference
```

I would write it like this mimicking the concept template and making sure that the ClassificationReference is meaningfully connected to elements in the model. Note that some of the entity type checks are probably redundant, because the attribute already strictly mandates that type.

features/Functional-parts.md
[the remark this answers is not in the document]
Ghesselink: It's already added a while back (see line 17) 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/IVS-59-Activation-functional-parts/features/Functional-parts.md

As well as in 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/d608eecbed91d2afc0ae40ab584af12d9885adbb/features/resources/functional_parts.csv#L11

features/CLS000_Classification-association.feature
```diff
@@ -0,0 +1,15 @@
+@implementer-agreement
+@CLS
+@version1
+@E00020
+
+Feature: CLS000 - Classification Association
+    The rule verifies that there is an ability to classify elements, materials, and systems according to various classification systems,
```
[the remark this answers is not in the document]
Ghesselink: Ok, added some extra leniency in the protocol checks for the feature description (it was enforcing that this description starts with `'The rule verifies **that**'`
https://github.com/buildingSMART/ifc-gherkin-rules/pull/239/commits/7b398c611216f107fb734afd68ac0a56df1c39ec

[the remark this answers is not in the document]
Ghesselink: Added 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/239/commits/6a131331ecc3f0dbf0c805ab21111896bac7f5e0

features/SPA000_Spaces-information.feature
Ghesselink: Added 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/239/commits/f2eace6613838c21a7629b17e7285d40c2ff53d8

features/steps/thens/existance.py
Ghesselink: That's what I implemented initially, but I ran into some errors in the decorator. We're iterating over yielded results, however when we only include 'pass' it seems that behave returns a None, causing an error (as yielding over None doesn't work). 
image

test/files/cls000/pass-cls000-IfcClassificationReference_present.ifc
aothms: Shouldn't we change that? Instead of "not failed" check for "passed" in the unit tests?

features/resources/functional_parts.csv
Ghesselink: Is already added a while ago 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/f2eace6613838c21a7629b17e7285d40c2ff53d8/features/resources/functional_parts.csv#L11

test/files/cls000/pass-cls000-IfcClassificationReference_present.ifc
Ghesselink: I think so, yes. Right now we're using a string representation of the model instance to parse some info to tabulate, probably we should use that information for testing instead of the pass/fail from behave itself. That part is, I think, a bit outdated but we're heavily relying on it with the unit tests .. 
I'll have a look

civilx64: > Maybe I'm mistaken but I think this is the intention. This kind of rule by definition can only differentiate between Executed and Passed. Gray vs Green.

Agree.  The issue is that this file is supposed to be a passing example that turns the rule green as Passed.

Ghesselink: Ah, I see the confusion .. Changed the testfile to pass instead https://github.com/buildingSMART/ifc-gherkin-rules/pull/239/commits/8f4d6e99c82698bc423f313235e6657b66b063e8
Highlighting the point to include this in the unit tests.

test/files/spa000/pass-spa000-activated_space_present.ifc
Ghesselink: That's weird .. Is the same true for the pass file with IfcSpatialZone on your local machine?

Ghesselink:
Superceded by smaller pull requests

PR #240, opened 2024-07-22
ivs43-hotfix/GEM051 error message

Ghesselink:
image

### DIFF
--- a/features/steps/validation_handling.py
+++ b/features/steps/validation_handling.py
@@ -285,6 +285,8 @@ def expected_behave_output(context: Context, data: Any, is_observed : bool = Fal
     """
     When serializing values in the observed field we never produce OneOf
     """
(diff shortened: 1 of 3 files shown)


features/steps/validation_handling.py
```diff
@@ -285,6 +285,8 @@ def expected_behave_output(context: Context, data: Any, is_observed : bool = Fal
     """
     When serializing values in the observed field we never produce OneOf
     """
+    if "GEM051" in context.feature.name and context.is_global_rule and not is_observed:
```
aothms, line 288: I don't think this is a scalable way to fix this as it creates a really tight and hard to discover coupling between rule code and handling.

I would say we better concatenate the step names or something in the impl function

```
expected = "Assert existance of " + get_previous_step_before_assertion(context)
```

PR #243, opened 2024-07-24
ALS016 Industry best practice for horizontal alignment geometric continuity (IVS-20)

civilx64:


features/ALS016_Alignment-horizontal-segment-geometric-continuity.feature
```diff
@@ -0,0 +1,29 @@
...
+Background:
+
+  Given A model with Schema "IFC4.3"
+  Given An IfcAlignment
+  Given Its attribute Representation
+  Given Its attribute Representations
+  Given RepresentationType = 'Curve2D'
+  Given All referenced instances
+  Given Its Entity Type is 'IfcCompositeCurve'
+  Given Its attribute Segments grouped pairwise
```
aothms, line 21:
```suggestion
  Given Its attribute Segments 
  Given The values grouped pairwise
```

I think I would have preferred this broken down into two statements, but we can revisit that when we have multiple grouping operations or grouping operations on in a different context instead of attribute retrieval.

The alternative is to supply a utility function used in the implementation of the then function to lookup the next sibling in execution context, but I think this approach in the given statements is very nice.

features/steps/thens/alignment.py
```diff
@@ -352,3 +359,67 @@ def step_impl(context, inst, activation_phrase):
...
+    position_transition_codes = ["CONTINUOUS", "CONTSAMEGRADIENT", "CONTSAMEGRADIENTSAMECURVATURE"]
+    tangency_transition_codes = ["CONTSAMEGRADIENT", "CONTSAMEGRADIENTSAMECURVATURE"]
+
+    length_unit_scale_factor = ifcopenshell.util.unit.calculate_unit_scale(
+        ifc_file=context.model,
+        unit_type="LENGTHUNIT"
+    )
+    for pair in inst:
+        pair_continuity_was_calculated = False
+        previous, current = pair
```
aothms, line 379:
```suggestion
    for previous, current in inst:
        pair_continuity_was_calculated = False
```

features/steps/thens/alignment.py
```diff
@@ -352,3 +359,67 @@ def step_impl(context, inst, activation_phrase):
...
+        precision = ifc.get_precision_from_contexts(entity_contexts)
+        precision_str = str(precision)
+        parts = precision_str.lower().split("e")
+        mantissa = parts[0]
+        sig_figs = len(mantissa)
+        if "." in mantissa:
+            sig_figs -= 1
+
+        exponent_str = parts[-1]
+        display_sig_figs = int(abs(float(exponent_str)) + sig_figs + 1)  # add one more digit for rounding confirmation
```
aothms, line 390: Would this also work?

```suggestion
      from math import log10, ceil
      display_sig_figs = int(ceil(log10(precision))) + 1
```

```
>>> -log10(0.01)
2.0
```

features/steps/thens/alignment.py
```diff
@@ -352,3 +359,67 @@ def step_impl(context, inst, activation_phrase):
...
+                    observed = geometry.alignment_segment_positional_difference(
+                        length_unit_scale_factor,
+                        previous,
+                        current
+                    )
+                    pair_continuity_was_calculated = True
+
+            case "tangency":
+                if current.Transition in tangency_transition_codes:
+                    expected = math.atan2(precision, float(current.SegmentLength.wrappedValue) * 1000)
```
aothms, line 405: Why `float()`. SegmentLength is already a float type I assume?

features/steps/thens/alignment.py
```diff
@@ -352,3 +359,67 @@ def step_impl(context, inst, activation_phrase):
...
+                    observed = geometry.alignment_segment_angular_difference(
+                        length_unit_scale_factor,
+                        previous,
+                        current
+                    )
+                    pair_continuity_was_calculated = True
```
aothms, line 411: Instead of capturing whether something was calculated which is error prone and easily forgotten I would rather explicitly handle the default case.

```suggestion
            case _: return
```

features/steps/thens/alignment.py
aothms, line 405: I think this is a good way of converting the linear tolerance to a an angular one, but 

1 I don't understand the factor 1000.
2 I think this kind of calculation deserves to be noted down in the scenario text.

features/steps/thens/alignment.py
```diff
@@ -352,3 +359,67 @@ def step_impl(context, inst, activation_phrase):
...
+            observed_msg = "{:.{width}e}".format(observed, width=display_sig_figs)
+
+            # add a suffix so that we keep the number formatting
+            # otherwise the values will be coerced back to floats
+            expected_msg += f" (max deviation in {continuity_type.lower()})"
+            observed_msg += f" (calculated deviation in {continuity_type.lower()})"
+            yield ValidationOutcome(
+                inst=current,
+                expected=expected_msg,
+                observed=observed_msg,
```
aothms, line 424: I know this is annoying, but I'd really urge us to do formatting in the frontend to allow for internationalization and querying the db.

e.g `{'value': observed, 'ndigits': ..., 'context': ...}`

features/steps/utils/geometry.py
```diff
@@ -77,3 +85,41 @@ def is_closed(context, instance):
     precision = get_precision_from_contexts(entity_contexts)
     points_coordinates = get_points(instance)
     return math.dist(points_coordinates[0], points_coordinates[-1]) < precision
+
+
+def evaluate_segment(segment: ifcopenshell.entity_instance, dist_along: float) -> np.ndarray:
+    s = ifcos_geom.settings()
+    pwf = wrapper.map_shape(s, segment.wrapped_data)
+
+    prev_trans_matrix = pwf.evaluate(dist_along)
```
aothms, line 94: How sure are we that this does not do the "interpolation" / "smoothing" with next segment. (The one that messes up multi-threading.)

Do we have some exact hand calculated solutions to verify?

features/steps/utils/geometry.py
```diff
@@ -77,3 +85,41 @@ def is_closed(context, instance):
...
+def alignment_segment_positional_difference(length_unit_scale_factor, previous_segment, segment_to_analyze):
+
+    u = abs(float(previous_segment.SegmentLength.wrappedValue)) * length_unit_scale_factor
+    prev_end_transform = evaluate_segment(segment=previous_segment, dist_along=u)
+
+    pX = prev_end_transform[3][0] / length_unit_scale_factor
+    pY = prev_end_transform[3][1] / length_unit_scale_factor
+    preceding_end = (pX, pY)
+
+    current_start = segment_to_analyze.Placement.Location.Coordinates
```
aothms, line 107: 1 How sure are we Placement.Location is 2D. I assume it would mess up our call to dist() if we pass a 2d and a 3d vector.
2 How sure are we Placement is not Linear?
3 Did you consider simply calling evaluate(0)?

aothms (changes requested):
Awesome work @civilx64

features/ALS016_Alignment-horizontal-segment-geometric-continuity.feature
civilx64: I will go ahead and make the change now as I anticipate `pairwise` being re-used very soon for things like confirming same number of vertices on cross-sections being used for swept geometry.

features/steps/thens/alignment.py
civilx64: Good question!  In earlier experiments with the IFCOS v0.8.0 wrapper I thought I saw this coming back as a string.  Perhaps my inner GPT hallucinated it.

features/steps/thens/alignment.py
civilx64: Yes but there is an additional layer of logic regarding the `Transition` code of the segment and whether it is to be interpreted as continuous or not.  But point taken; I will revise to stack conditions with `if` statements and remove the pattern matching.

features/steps/utils/geometry.py
civilx64: There is no "interpolation" / "smoothing" between segments.  That was discussed early in the C++ implementation but it turned out to be difficult and problematic, therefore was dropped.

features/steps/utils/geometry.py
civilx64: > 1 How sure are we Placement.Location is 2D. I assume it would mess up our call to dist() if we pass a 2d and a 3d vector. 

Good point, I have made this explicit.

> 2 How sure are we Placement is not Linear?

Not sure I understand.  Do you mean that `Placement` might be of type `IfcAxis2PlacementLinear`?  And I assume that this would be for things like an alignment defined by IfcOffsetCurveByDistances?

> 3 Did you consider simply calling evaluate(0)?

Yes, in fact I initially had the logic set up that way but prefer it as-is as it has the side-effect of providing and double-check on imperial unit conversions.  It was a happy accident that I started developed with an imperial example, otherwise I might have completely overlooked this.

features/steps/thens/alignment.py
civilx64: > 1 I don't understand the factor 1000.

That remained from some debugging and shouldn't be there.

> 2 I think this kind of calculation deserves to be noted down in the scenario text / description.

Agree; it has been added to the top of the feature file.

features/steps/thens/alignment.py
civilx64: Good point.  I have revised this.

features/steps/thens/alignment.py
civilx64: Yes. I have revised accordingly.

features/steps/thens/alignment.py
civilx64: > 1 I don't understand the factor 1000.

That remained from some debugging and shouldn't be there.  I've removed it, which requires a small tweak to a test file to ensure that it continues to fail.

features/ALS016_Alignment-horizontal-segment-geometric-continuity.feature
```diff
@@ -0,0 +1,32 @@
...
+
+  Given A model with Schema "IFC4.3"
+  Given An IfcAlignment
+  Given Its attribute Representation
+  Given Its attribute Representations
+  Given RepresentationType = 'Curve2D'
+  Given All referenced instances
+  Given Its Entity Type is 'IfcCompositeCurve'
+  Given Its attribute Segments 
+  Given The values grouped pairwise
```
Ghesselink, line 24:
```suggestion
  Given The values grouped pairwise at depth 1
```
I've ran this locally and it resolved the failing CI/CD. 
Without `at depth 1` we'd be executing`itertools.pairwise` for each segment individually instead of grouped together.
image

features/steps/givens/values.py
```diff
@@ -1,7 +1,17 @@
...
 @gherkin_ifc.step("Its values excluding {excluding}")
-def step_impl(context, inst, excluding = None):
-    yield ValidationOutcome(instance_id=inst.get_info(recursive=True, include_identifier=False, ignore=excluding), severity=OutcomeSeverity.PASSED)
\ No newline at end of file
+def step_impl(context, inst, excluding=None):
+    yield ValidationOutcome(instance_id=inst.get_info(recursive=True, include_identifier=False, ignore=excluding),
+                            severity=OutcomeSeverity.PASSED)
+
+
+@gherkin_ifc.step("The values grouped pairwise")
```
Ghesselink, line 14:
```suggestion
@gherkin_ifc.step("The values grouped pairwise at depth 1")
```

features/steps/thens/alignment.py
```diff
@@ -352,3 +359,61 @@ def step_impl(context, inst, activation_phrase):
...
+                inst=current,
+                expected={
+                    "expected": expected,
+                    "num_digits": display_sig_figs,
+                    "context": context,
+                },
+                observed={
+                    "observed": observed,
+                    "num_digits": display_sig_figs,
+                    "context": context,
```
Ghesselink, line 417: Do we want to pass the whole context over? 
In the backend, it's not needed as we have this available in the decorator. In the frontend, it's perhaps not so useful information 
image

test/files/als016/fail-als016-scenario01-imperial_not_continuous.ifc
Ghesselink, line 1: Can we make the filenames for testing a bit more distinctive? They are scenario01 and scenari01, but have the same description.

Ghesselink (commented):
Looks good :) 
It seems that the failing CI/CD can be resolved by adding `at depth 1` to the pairwise grouping statement. At least locally, this works. Before we were grouping`itertools.pairwise` for each segment individually instead of all the segments at once.

features/steps/thens/alignment.py
Ghesselink: Ah, sorry, I now understand the purpose after seeing Thomas' comment. I think there is a confusion between two types of contexts:
- The behave Context class that we're using in Gherkin to transfer, for example, stack attributes.
- The context for parsing in frontend. 

For the latter, please refer to the PRs from a little while ago regarding[ adding context in Gherkin](https://github.com/buildingSMART/ifc-gherkin-rules/pull/185/files#diff-eddbd16c15972064cf608a2e27ef9ccc873a806fa85084208c1d59e1028846f6) and [parsing on the frontend](https://github.com/buildingSMART/validate/pull/65/files). 
We'd probably need to do something similar here

features/steps/utils/geometry.py
aothms: >  Do you mean that Placement might be of type IfcAxis2PlacementLinear?

yes, this is allowed by the schema, or am I hallucinating this time?

features/ALS016_Alignment-horizontal-segment-geometric-continuity.feature
aothms: Cool. Thanks for taking the time. I just realized I forgot another point related to the complete lack of separation of concerns in the IFC schema.

What if we have an IfcAlignment that uses IfcCompositeCurveSegment as opposed to IfcCurveSegment (lord, do I hate this naming distinction). I.e the pre 4.3 composite curve with regular step segments. We at least need to make sure this is gracefully handled, preferably explicitly.

I think Alignment with pre-4.3  composite curve should be allowed as wel also allow polyline.

Maybe something we can do something with `Given its entity type is IfcCurveSegment`. But my early morning brain cannot predict how that would work with pairwise.

That would also mean `ALS011 - Alignment segment entity type consistency` is kind of a precursory check for this rule.

Thoughts?

features/steps/utils/geometry.py
civilx64: It appears to be allowed by the schema.  I am not aware of any examples in the wild that are using this.

features/ALS016_Alignment-horizontal-segment-geometric-continuity.feature
civilx64: Agree.  I added `Its Entity Type is 'IfcCurveSegment'`

features/steps/utils/geometry.py
civilx64: To resolve this part of the conversation - I recommend that we leave the rule as-is and see if any implementations actually utilize `IfcAxis2PlacementLinear` before we spend effort handling that potential wrinkle.

In other words, not letting the perfect be the enemy of the good :smiley:

features/ALS016_Alignment-horizontal-segment-geometric-continuity.feature
civilx64: Agree, added `at depth 1`

test/files/als016/fail-als016-scenario01-imperial_not_continuous.ifc
civilx64: Done, added `_break_in_position` and `_break_in_direction_angle` for scenario01 and scenario02, respectively.

features/steps/utils/geometry.py
aothms: > In other words, not letting the perfect be the enemy of the good 😃

But the problem is that when it occurs it's not gracefully/explicitly handled. Even if we don't "support" this configuration for checking the rule, I don't think that we should let the check crash and burn on an AttributeError.

Because even if we later add another rule that says this Placement should be local, than still we cannot raise exceptions here, because it would prevent the other rule results from showing up.

I would say:

- We do an if-statement here just to prevent the exception, but doesn't need to be handled here
- We ask for feedback in the IF whether we can constrain the schema here with another normative rule. These kind of "imprecise" attributes can be real interoperability killers.

aothms (approved):

features/steps/thens/alignment.py
civilx64: Done, see https://github.com/buildingSMART/validate/pull/99

features/steps/utils/geometry.py
civilx64: So in summary you are correct, of course, given your wide understanding of potential land mines in the schema.  The best solution is to use `evaluate(u=0.0)` to determine the start position and direction angle of the segment being analyzed.  Which was one of your first questions in review of this PR.  So you get a double gold star.

PR #244, opened 2024-07-25
feature/IVS-65 functional part commin step impl

Ghesselink:
See comments at https://github.com/buildingSMART/ifc-gherkin-rules/pull/239#discussion_r1685865257 

This task is relatively small, but it's probably best to avoid adding it to every new rule created to activate the functional part. This way, if we need to make changes to this code in the future, we won't have to adjust every affected PR.

Further changes; 
- We can be more lenient in the protocol check regarding the feature description starting with '`The rule verifies that`'. Rather, `'the rule verifies'` is sufficient and gives us a bit more room. See https://github.com/buildingSMART/ifc-gherkin-rules/pull/244/commits/ef4cbc2b3b71b5621daf13ef0c7e25a01bf09667
- Fixed a small error, but with big consequences. The distinction between passed and activated was not reached in the handling script because of wrong indentation. Fixed in https://github.com/buildingSMART/ifc-gherkin-rules/pull/244/commits/73ab399d24ca4fecbb07cef693c5317cea760c80

### DIFF
--- a/features/steps/thens/existance.py
+++ b/features/steps/thens/existance.py
@@ -69,7 +69,7 @@ def step_impl(context, inst):
 
     if not any(recursive_flatten(inst)):
         expected = get_previous_step_before_assertion(context)
-        yield ValidationOutcome(instance_id=inst, expected = expected, observed='Nonexistent', severity = OutcomeSeverity.ERROR)
+        yield ValidationOutcome(instance_id=inst, expected=expected, observed='Nonexistent', severity=OutcomeSeverity.ERROR)
 
 
 @gherkin_ifc.step("The IFC model contains information on {functional_part_description}")
--- a/features/steps/thens/existance.py
+++ b/features/steps/thens/existance.py
@@ -70,3 +70,9 @@ def step_impl(context, inst):
     if not any(recursive_flatten(inst)):
         expected = get_previous_step_before_assertion(context)
         yield ValidationOutcome(instance_id=inst, expected = expected, observed='Nonexistent', severity = OutcomeSeverity.ERROR)
+
+@gherkin_ifc.step("The IFC model contains information on {functional_part_description}")
+def step_impl(context, inst, functional_part_description):
+   # This rule is designed to always pass and is used solely to trigger the activation of the rule 
+    # if an instance linked to the functional part is present.
+    return []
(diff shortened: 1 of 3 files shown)


aothms (approved):

features/steps/thens/existance.py
```diff
@@ -69,4 +69,11 @@ def step_impl(context, inst):
 
     if not any(recursive_flatten(inst)):
         expected = get_previous_step_before_assertion(context)
-        yield ValidationOutcome(instance_id=inst, expected = expected, observed='Nonexistent', severity = OutcomeSeverity.ERROR)
\ No newline at end of file
+        yield ValidationOutcome(instance_id=inst, expected = expected, observed='Nonexistent', severity = OutcomeSeverity.ERROR)
```
civilx64, line 72:
```suggestion
        yield ValidationOutcome(instance_id=inst, expected=expected, observed='Nonexistent', severity=OutcomeSeverity.ERROR)
```
Linting cleanup

features/steps/validation_handling.py
```diff
@@ -164,19 +164,19 @@ def apply_then_operation(fn, inst, context, current_path, depth=0, **kwargs):
...
-                        feature_version=misc.define_feature_version(context),
-                        severity=OutcomeSeverity.PASSED,
-                        instance_id = safe_method_call(activation_inst, 'id', None),
-                        validation_task_id=context.validation_task_id
-                    )
-                    context.gherkin_outcomes.append(validation_outcome)
+            if not step_results:
+
+                validation_outcome = ValidationOutcome(
+                    outcome_code=ValidationOutcomeCode.PASSED,  # todo @gh "Rule passed" # deactivated until code table is added to django model
```
civilx64, line 170: Is this `todo` still applicable?  To me it looks like all of the outcomes are present in the django model - ref https://github.com/buildingSMART/ifc-validation-data-model/blob/b29efda228a1b019e21b9761fbedbdba0a40f81e/models.py#L979

Ghesselink: No, not applicable anymore .. 
I actually also don't see any passed results in our database for normative rules or warnings. Which makes sense; the code never gets reached. The reason that the blocks are turned grey is then because 'executed' flags a file as valid.

PR #248, opened 2024-07-28
IVS69/POR000 Network Connections

Ghesselink:
From IVS-59; 

Also refer to; https://github.com/buildingSMART/ifc-gherkin-rules/pull/239

Passing testfiles, i.e. activated. Activated test files are those where the rule is triggered, reaching the step implementation without producing any errors. Therefore, it has the status` OutcomeSeverity.PASSED ` as the highest severity stored in the db.
image

Not activated file; 
image


features/POR000_Network-connections.feature
```diff
@@ -0,0 +1,20 @@
+@implementer-agreement
+@POR
+@version1
+@E00020
+
+Feature: POR000 - Network Connections
+    The rule verifies the ability to define ports (as means for an element to connect to other elements) and the relationship that is made between two ports.
```
evandroAlfieri, line 7:
```suggestion
    The rule verifies the presence of IFC entities used to define ports (as means for an element to connect to other elements) and the relationship that is made between two ports.
```

evandroAlfieri (changes requested):

evandroAlfieri (approved):

aothms:
I have some high level comments:

- The functional part and scenario name are called "network connections", but this is about a specific type of network called Distribution Systems in IFC. E.g the Railway people might be confused into thinking that network connections is about railway networks. Indeed, everybody has their own networks, but the same applies to systems of course. Naming things is hard. Especially because of the 4.3 push for non-building specific naming, so we can't use terms like "building services" or "building systems" anymore. I would try to include something like "utility system" or "distribution system" in the name.

- This is the port to port connection. But that doesn't make it a network yet. It only becomes a network when these ports are also nested into distribution elements and with elements having multiple ports a network can be constructed. What is missing to have a proper network is described in concept template "4.1.4.4.4 Port Nesting".

  I think both have to be present in the model before we can say the tool supports "distribution system networks".

PR #249, opened 2024-07-28
Fix ContextIdentifier rule to use correct values for context identifier not representation types

Moult:
I didn't actually test / run this as I don't have a dev setup.


evandroAlfieri:
Thanks! Upon a quick investigation we actually found a related issue in GEM051. Working on fixing both.
They'll be part of next release (end of August)

Ghesselink:
Superseded by https://github.com/buildingSMART/ifc-gherkin-rules/pull/251

PR #250, opened 2024-07-29
IVS70/MAT000 Materials

Ghesselink:
Activated
image

Not-activated
image
image


features/MAT000_Materials.feature
```diff
@@ -0,0 +1,20 @@
+@implementer-agreement
+@MAT
+@version1
+@E00020
+
+Feature: MAT000 - Materials
+    The rule verifies the ability to define materials assigned to elements.
```
evandroAlfieri, line 7:
```suggestion
    The rule verifies the presence of IFC entities used to define materials assigned to elements.
```

evandroAlfieri (changes requested):

features/MAT000_Materials.feature
```diff
@@ -0,0 +1,20 @@
...
+    https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Association/Material_Association/Material_Single/content.html
+
+
+    Scenario: Check for activation of Materials
+
+    Given an IfcObjectDefinition
+    Given its attribute HasAssociations
+    Given its entity type is 'IfcRelAssociatesMaterial'
+    Given its attribute RelatingMaterial
+    Given its entity type is 'IfcMaterialDefinition' including subtypes
```
aothms, line 17: IfcMaterialDefinition is added in IFC4 so we can't reference it like this on a file with 2x3 schema. Good news is that this last line is redundant. I would just remove the bottom two given statements.

aothms (changes requested):

features/MAT000_Materials.feature
Ghesselink: What if we make it schema-specific? For IFC4.3 and IFC4, we can keep it as-is, but for IFC2X3, we remove the bottom two given statements. See also ; https://github.com/buildingSMART/ifc-gherkin-rules/pull/250/commits/20399c36819d81bad2be0db3b41dab424a93d0aa

```
    Scenario: Check for activation of Materials - IFC4 & IFC4.3

        Given A model with Schema "IFC4.3" or "IFC4"
        Given an IfcObjectDefinition
        Given its attribute HasAssociations
        Given its entity type is 'IfcRelAssociatesMaterial'
        Given its attribute RelatingMaterial
        Given its entity type is 'IfcMaterialDefinition' including subtypes

        Then The IFC model contains information on materials assigned to elements


    Scenario: Check for activation of Materials - IFC2X3

        Given A model with Schema "IFC2X3"
        Given an IfcObjectDefinition
        Given its attribute HasAssociations
        Given its entity type is 'IfcRelAssociatesMaterial'

        Then The IFC model contains information on materials assigned to elements
```
cc @evandroAlfieri

evandroAlfieri: Fine for me, no objections. If we go that direction, why not making it explicit for 2x3 by adding?
```
        [...]
        Given its attribute RelatingMaterial
        Given its entity type is 'IfcMaterial'

        Then The IFC model contains information on materials assigned to elements
```

Ghesselink: The tests won't be completely identical in that case. Correct me if I'm wrong, but if we'd like to define it explicitly like you mention and be as close as possible to the CT (but for IFC2X3 instead) we should take `IfcMaterialSelect including subtypes` instead. 
This includes IfcMaterial, but also other subtypes 
image
image
https://standards.buildingsmart.org/IFC/RELEASE/IFC2x3/TC1/HTML/ifcproductextension/lexical/ifcrelassociatesmaterial.htm

evandroAlfieri: You're right. We should write all the possible 5 entities in the select. Or leave this out as you suggested initially. Your call, as said, no preferences.

Ghesselink: In that case I'll leave it as suggested. 
I did a quick test; changing the entity that is linked through the RelatingMaterial attribute to anything else than a subtype of IfcMaterialSet will cause a schema error in the validation platform.

PR #252, opened 2024-07-29
IVS-73/Update functional part in gherkin docs

Ghesselink:


evandroAlfieri (approved):

features/resources/functional_parts.csv
evandroAlfieri, line 1: Not sure what we use this list for, maybe internal checks that a rule has a code that uses one of these values. In any case, ALB is missing

evandroAlfieri (changes requested):

features/resources/functional_parts.csv
Ghesselink: > maybe internal checks that a rule has a code that uses one of these values
 
 This indeed. It's an easy maintainable list to validate (in the protocol checks) whether the rule-code used in the feature file is valid. For instance, typo's will then lead to a failing CI/CD. 
 If we'd use the markdown file it's (a) a bit more time-consuming to write a script and (b) every time the format of that markdown changes the script also has to be modified (this would've been the case now). 
 
 > ALB is missing
 
 It's [here (line 4)](https://github.com/buildingSMART/ifc-gherkin-rules/blob/d10fa69f37f6b9d5fa83bf1db394428dbfebaf09/features/resources/functional_parts.csv#L4) :-)

Ghesselink: To demonstrate, I[ deleted the ALB lines ](https://github.com/buildingSMART/ifc-gherkin-rules/pull/252/commits/f7d504e81331720d4949fecba31096c1d12b726c)and it resulted in a failing CI/CD. 
image

image


(and, afterwards, reverse the change) 
image

evandroAlfieri: Can you do a final check. Numbers don't add up. I see 53 tags, but functional parts are 27+9+3+3(4 if you count alignment having both ALA and ALB)=43. It's 10 more

Ghesselink: Yes, the following are not in the markdown
- IFC
- IBP
- SYS 

These three actually have rules (all-tough perhaps no `IBP` soon)

These come from the previous markdown, but are unused. See also 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/main/features/Functional-parts.md
- MDS- (Model setup)
- ALG - Alignment (Geometry)
- STN - (Stationing)
- PRP - (Properties)

evandroAlfieri: The last 4 can go. No rules will have those codes, as no functional parts exist with such tag.
- IFC, can keep it for naming rules. But no tag can exist
- IBP, can be removed. We have another task to update the one and only rule using this code. After that, no IBP rules nor tags shall exist
- SYS, can be removed too. The only rule using it is disabled. And if we'll ever activate it, we should rename it and tag it properly.

Bottom line, you can remove 6 out of 7. Keep only IFC

PR #255, opened 2024-08-06
IVS-77/ASM000: Composed elements

Ghesselink:
Instances are only populated in the activated file; i.e. no then step is actually being executed on the others for ASM000. 

https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Composition/Aggregation/Element_Composition/content.html


features/ASM000_Assemblies-information.feature
```diff
@@ -0,0 +1,21 @@
+@implementer-agreement
+@ASM
+@version1
+@E00020
+
+Feature: ASM000 - Assemblies information
+    The rule verifies the presence of elements composed of / constructed by other elements.
```
evandroAlfieri, line 7:
```suggestion
    The rule verifies the presence of IFC entities used to model elements composed of / constructed by other elements.
```
It's the same structure, over and over and over again.
Take this sentence "The rule verifies the presence of IFC entities used to" and append it before the verb of each description - in this case "...model elements composed of..."

evandroAlfieri (changes requested):

features/ASM000_Assemblies-information.feature
```diff
@@ -0,0 +1,21 @@
...
+
+Feature: ASM000 - Assemblies information
+    The rule verifies the presence of IFC entities used to model elements composed of / constructed by other elements.
+    For example, a roof might be assembled from a series of prefabricated truss components.
+    https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Composition/Aggregation/Element_Composition/content.html
+
+
+    Scenario: Check for activation
+
+    Given an IfcElement
```
evandroAlfieri, line 14:
```suggestion
    Given an IfcElementAssembly
```
I would make this rule specific to detect presence of element assemblies

features/ASM000_Assemblies-information.feature
```diff
@@ -0,0 +1,21 @@
+@implementer-agreement
+@ASM
+@version1
+@E00020
+
+Feature: ASM000 - Assemblies information
```
aothms, line 6:
```suggestion
Feature: ASM000 - Composed elements
```

I like this. This is essentially what you've written in the final then step.

features/ASM000_Assemblies-information.feature
```diff
@@ -0,0 +1,21 @@
...
+    The rule verifies the presence of IFC entities used to model elements composed of / constructed by other elements.
+    For example, a roof might be assembled from a series of prefabricated truss components.
+    https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Composition/Aggregation/Element_Composition/content.html
+
+
+    Scenario: Check for activation
+
+    Given an IfcElementAssembly
```
aothms, line 14: We have to take a second here to think about what this functional part means.

- Does it mean the usage of IfcElementAssembly
- Or does it mean the usage of the IfcRelAggregates relationship, which can equally well be applied to e.g a wall or a stair.

(I would actually say the 2nd one, because IfcElementAssembly is a bit the proxy element of compositions.

Note from the docs:

> The [IfcElementAssembly](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcElementAssembly.htm) is a general purpose entity that is required to be decomposed. **Also other subtypes of [IfcElement](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcElement.htm) can be decomposed**

(emphasis mine)

aothms (commented):
CC @evandroAlfieri

features/ASM000_Assemblies-information.feature
evandroAlfieri: Fair enough. If what this FP entails is the ability of applications to define assemblies (ie, elements made of other elements) then I'd say it's ok to have it generic (ie, Given an IfcElement or its subtypes)

features/ASM000_Assemblies-information.feature
Ghesselink: Changed it in https://github.com/buildingSMART/ifc-gherkin-rules/pull/255/commits/6056105fa6493832381c41a50b742a8c0c54448e

Also at the feature filename, the PR name and the Jira ticket. 
It's also closer to the CT name (Element Composition)

features/ASM000_Assemblies-information.feature
Ghesselink: [Reversed it to IfcElement](https://github.com/buildingSMART/ifc-gherkin-rules/pull/255/commits/6056105fa6493832381c41a50b742a8c0c54448e)

PR #256, opened 2024-08-07
IVS-79/LAY000 - Presentation Layer Assignment

Ghesselink:
A couple of remarks; 

- The schema does not allow for any other entity type than `IfcPresentationLayerAssignment` for the attribute `LayerAssignments` (of `IfcShapeRepresentation`). It seems it's not possible to assign any other entity than mentioned above due to the assigning taking place by the AssignedItems (of IfcPresLayerAssign) attribute. Therefore, I think it would be redundant to add this explicitly in a gherkin steps; solely `Its attribute LayerAssignments `suffices.
image
image

- The not-activated file has the following list in context.instances that is getting handled by the then step. 
image
- The activated file has the following instances in the stack for the last then step. Index 5 and 7 (both IfcPresentationLayerAssignment) activate the rule.
image


evandroAlfieri (approved):
great work

features/LAY000_Layers-information.feature
```diff
@@ -0,0 +1,19 @@
...
+
+
+    Scenario: Check for activation
+
+    Given an IfcProduct
+    Given its attribute Representation
+    Given its attribute Representations
+    Given its attribute LayerAssignments
+
+    Then The IFC model contains information on geometry layers
```
aothms, line 19: geometry layers != layers information

Shouldn't this be consistent?

features/LAY000_Layers-information.feature
```diff
@@ -0,0 +1,19 @@
+@implementer-agreement
+@LAY
+@version1
+@E00020
+
+Feature: LAY000 - Layers Information
```
aothms, line 6: Better english would be layer, but I would also really add the word Presentation, so that it's not to be confused with the physical material layers for compound structures such as walls and slab (IfcMaterialLayerSet). Lastly since again it doesn't really contain information *about* these layers, I would call it assignment. And then accidentally I have arrived at the verbatim entity name.

```suggestion
Feature: LAY000 - Presentation Layer Assignment
```

features/LAY000_Layers-information.feature
```diff
@@ -0,0 +1,19 @@
...
+    The rule verifies the presence of IFC entities used to assign layers (also known as, CAD layer) to collection of elements. 
+    This is used mainly for grouping and visibility control, and in general to organise geometry into groups that may be shown or hidden.
+    https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Geometry_Layer/content.html
+
+
+    Scenario: Check for activation
+
+    Given an IfcProduct
+    Given its attribute Representation
+    Given its attribute Representations
```
aothms, line 16: There is one additional consideration, that's a bit in detail, but actually you can have layer assignment on the *representation* but also on the *items*. This is missing in the concept template, but is rendered in the documentation. 

ifcpresentationlayerassignment-fig1

I would actually add this, because semantically it makes more sense for me to assign the representation *item* than to assign the representation itself.

```
    Scenario: Layer assignment to representation

    Given an IfcProduct
    Given its attribute Representation
    Given its attribute Representations
    Given its attribute LayerAssignments

    Scenario: Layer assignment to representation items

    Given an IfcProduct
    Given its attribute Representation
    Given its attribute Representations
    Given its attribute Items
    Given its attribute LayerAssignments
```

features/LAY000_Layers-information.feature
```diff
@@ -0,0 +1,19 @@
...
+    Given its attribute LayerAssignments
```
aothms, line 17: Is the check that this is not empty in the `The IFC model contains information on` implementation? That's a bit implicit.

Should there not be something like:

   Given its attribute LayerAssignments is not empty

or separated into two steps?

aothms (changes requested):
I must say I find this functional part not something I would advise end users to make decisions on.

features/LAY000_Layers-information.feature
Ghesselink: [Changed it ](https://github.com/buildingSMART/ifc-gherkin-rules/pull/256/commits/9f838a0518dfea291bdec4fb3477ce8ae05665af)to 
`The IFC model contains information on the selected functional part`
See https://github.com/buildingSMART/ifc-gherkin-rules/pull/257#discussion_r1712626433

features/LAY000_Layers-information.feature
Ghesselink: [Modified it](https://github.com/buildingSMART/ifc-gherkin-rules/pull/256/commits/1487b2ecafdb2edea99b9ca5f5d4d69269222745)
We'll also have to modify the feature filename in that case.

features/LAY000_Layers-information.feature
Ghesselink: The two statements are similar and, for activation, they function the same. The difference lies in which entity is selected for the next step. For these 000 rules, we don't evaluate the instances but simply check if they exist in the model. If the `LayerAssignments` attribute in an IFC file isn't empty, the rule will activate regardless of which statement we use.

Currently, the last instances in the stack (if any) are of type IfcPresentationLayerAssignment. With the 'is not empty' statement, they will be of type IfcShapeRepresentation (filtered out by the current statement from the previous Given statement).

Ghesselink:
Outdated https://github.com/buildingSMART/ifc-gherkin-rules/pull/281

Issue #124, opened 2023-12-20
Possible false positives for CT 4.1.7.4.2 Product Linear Placement


RickBrice:
If I understand [CT 4.1.7.4.2 Product Linear Placement](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/concepts/Product_Shape/Product_Placement/Product_Linear_Placement/content.html) correctly, IfcLinearPlacement.PlacementRelTo is the same IfcLocalPlacement as used to place the IfcLinearPositioningElement.

At the end of the attached file, three linear placements are defined.
`#279` IfcLinearPlacement.PlacementRelTo = `#22` which is the same placement for `#248` IfcAlignment.ObjectPlacement - This placement should be correct

`#287` IfcLinearPlacement.PlacementRelTo is a unique IfcLocalPlacement with the same values as `#22` but is a different placement - This seems to violate the CT but passes the validation service checks.

`#295` IfcLinearPlacement.PlacementRelTo is also a unique IfcLocalPlacement but with different values than `#22` - This seems to violate the CT as well but passes the validation service checks.

[FHWA_Bridge_Geometry_Alignment_Example.ifc.txt](https://github.com/buildingSMART/validate/files/13733158/FHWA_Bridge_Geometry_Alignment_Example.ifc.txt)

aothms:
I think the original issue is definitely valid and actionable. I've tagged and moved it to the rule repo so that it can be picked up.

civilx64:
@evandroAlfieri I propose we rename this to LIP001 since it corresponds to the Functional Part for Linear Placement.

evandroAlfieri:
Agree @civilx64
And feel free to propose a more concise title ;) You can integrate the current one in the description

aothms:
It's the same rule for LinearPlacement as well as GridPlacement. I agree that from a vendor / end-user perspective these should be separately positioned under `GDP` and `LIP` (I would probably name them `GPL` and `LPL` so that's a bit more clear that they are both `PL`acement related - but forgive me my obsessive side tracks).

This basically means that same thing we observed in "gcert". We end up copy pasting a lot, but that's fine. With the feature file / implementation divide and "plain-text-first" nature we kind of allow our selves to copy paste feature files as it's a bit like a configuration / orchestration file.

evandroAlfieri:
Good that we can assign multiple tags. So, 
- let's use the generic one for the rule's code (i.e, `OJP` for object placement)
[...]

PR #257, opened 2024-08-07
IVS-80/GRD000 - Grid Information

Ghesselink:
This one is a bit more elaboration and includes changes in the python implementation

General/improvements not limited to the current rule:
- Added a return to entity statement used to navigate within the concept templates in gherkin by returning to a previous layer in the context stack. This statement was previously used for an [unmerged pull request](https://github.com/buildingSMART/ifc-gherkin-rules/pull/103/files#diff-ad0dd7437200c694263bb5dcbbc1dd7ee980d2264bc1e19587e6c905a8ed8848), and updated to work with the decorator. One caveats is that I'm only confident in using this statement for activation purposes, not for normative rules due to the risk validating instances that are filtered out by previous given statement (as we go back to the point in the context stack where they were not filtered out yet). It would also be nice to simply go back X amount of statements, or [repeat previous steps](https://github.com/buildingSMART/ifc-gherkin-rules/pull/103/files#diff-ad0dd7437200c694263bb5dcbbc1dd7ee980d2264bc1e19587e6c905a8ed8848). 
- The activation instances issue of GEM051 is now handled generically. We used activation_instances[top_level_index] twice which was redundant. I've tested that GEM051 has the same activation instances as before.
- Some cleanup was done to keep the logic of including/excluding subtypes in our statements in one location (we're re-using it for GRD000). 

GRD000 specific: 
- Again, ensured that non-activated files have no instances when the "then" step is reached, and vice versa; activated files has context.instances populated at the same point.. 
- Added a file that tests the return statement by being correct except for the absence of IfcGrid (which is in the middle of the sequence of statements)
- Three files created with IfcOpenShell, one from the IFC4.3x sample files repo.


features/GRD000_Grid-information.feature
```diff
@@ -0,0 +1,26 @@
...
+Feature: GRD000 - Grid Information
+    The rule verifies the presence of IFC entities used to define a design grid to be used as reference for object placement.
+    https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Placement/Product_Grid_Placement/content.html
+
+
+    Scenario: Check for activation
+
+    Given an IfcGridPlacement
+    Given its attribute PlacesObject
+    Given its entity type is 'IfcProduct' including subtypes
```
aothms, line 16: This is governed by the schema. 

    PlacesObject: SET [0:?] OF IfcProduct FOR ObjectPlacement

But what is not governed is that there is at least one Product being placed by the Placement, which I think is a good requirement to add here as an activation criterion.

features/GRD000_Grid-information.feature
```diff
@@ -0,0 +1,26 @@
...
+    Given return to IfcGridPlacement
+    Given its attribute PlacementRelTo
+    Given its attribute PlacesObject
+    Given its entity type is 'IfcGrid'
```
aothms, line 20: These three lines should actually not be a given statement, but a then statement.

    Given an IfcGridPlacement Then PlacementRelTo/PlacesObject **must be** IfcGrid

features/GRD000_Grid-information.feature
```diff
@@ -0,0 +1,26 @@
...
+    https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Placement/Product_Grid_Placement/content.html
+
+
+    Scenario: Check for activation
+
+    Given an IfcGridPlacement
+    Given its attribute PlacesObject
+    Given its entity type is 'IfcProduct' including subtypes
+    Given return to IfcGridPlacement
+    Given its attribute PlacementRelTo
```
aothms, line 18: PlacementRelTo does not exist on schemas pre IFC4.3. I think this must be reflected in the scenario.

features/GRD000_Grid-information.feature
```diff
@@ -0,0 +1,26 @@
...
+    Given its attribute PlacesObject
+    Given its entity type is 'IfcGrid' 
+    Given return to IfcGridPlacement
+    Given its attribute PlacementLocation
+    Given IntersectingAxes = not empty
```
aothms, line 23: This is given by the schema:

    IntersectingAxes: LIST [2:2] OF UNIQUE IfcGridAxis

features/GRD000_Grid-information.feature
```diff
@@ -0,0 +1,26 @@
...
+
+    Then The IFC model contains information on quantities of elements
```
aothms, line 25: Copy pasta

aothms: Speaking of which, maybe this is too error prone and we can rather easily replace this with:

```suggestion
    Then The IFC model contains information on the selected functional part
```

features/GRD000_Grid-information.feature
```diff
@@ -0,0 +1,26 @@
...
+
```
aothms, line 26: So all in all, you could consider rewriting this scenario as:

```gherkin
Given an IfcProduct
And its attribute ObjectPlacement
And its entity type is IfcGridPlacement

Then The IFC model contains information on ...
```

features/steps/steps/return_to.py
```diff
@@ -0,0 +1,50 @@
...
+            correct_entity = False
+            if isinstance(input, ifcopenshell.entity_instance):
+                correct_entity = check_entity_type(input, entity, context.include_or_exclude_subtypes)
+            context.include_layer = is_bool or correct_entity
+        layer = layer.get('instances')
+        misc.map_state(layer, check_inclusion_criteria)
+        return layer if context.include_layer else None
+
+    #ensure the stack does not get pupulated when nothing was yielded in the last step
+    if (lambda f: f(f))(lambda f: lambda data: bool(data) and (not isinstance(data, (list, tuple)) or any(f(f)(item) for item in data)))(context.instances):
```
aothms, line 47: Again this is rocket science to me. Can we clear this up first in the other PR

features/steps/validation_handling.py
```diff
@@ -203,6 +200,7 @@ def should_apply(items, depth):
 
     # evokes behave error
     generate_error_message(context, [gherkin_outcome for gherkin_outcome in context.gherkin_outcomes if gherkin_outcome.severity in [OutcomeSeverity.WARNING, OutcomeSeverity.ERROR]])
+    pass
```
aothms, line 203: Why this?

test/files/grd000/generate.py
```diff
@@ -0,0 +1,92 @@
...
+    GlobalId = ifcopenshell.guid.new(),
+    ObjectPlacement = grid_placement,
+)
+
+grid.ObjectPlacement = placement
+
+grid_placement.PlacementLocation = f.createIfcVirtualGridIntersection(
+    [grid_axis_1, grid_axis_2], 
+    (0.0, 0.0, 0.0)
+    )
```
aothms, line 66: This violates the informal propositions here: https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcVirtualGridIntersection.htm

It's also a really creative grid.

I would define:

```
  U axes:
    (0, 0) -> (0, 2)
    (1, 0) -> (1, 2)
    (2, 0) -> (2, 2)
  V axes:
    (0, 0) -> (2, 0)
    (0, 1) -> (2, 1)
    (0, 2) -> (2, 2)
  W:
    <empty>
```

And then you intersect e.g UAxes[1] with VAxes[1], but they need to be instances from the grid itself.

Also you don't actually reference this variable, but redefined sth on L 777

Also the polyline grid axes are best 2d instead of 3d.

aothms (changes requested):

features/steps/validation_handling.py
Ghesselink: A debug statement I forgot to remove, sorry 
https://github.com/buildingSMART/ifc-gherkin-rules/commit/4e3ce4409cc08ba0e69ba9f6a9b672f28979fc4b

features/steps/steps/return_to.py
Ghesselink: Let's bring the rocket science back to building information science then :rocket:  
See  https://github.com/buildingSMART/ifc-gherkin-rules/pull/258#discussion_r1712635216

features/GRD000_Grid-information.feature
Ghesselink: But would this be close enough to the [relating concept template](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Placement/Product_Grid_Placement/content.html)? 
Would it be nice if we could state that some software supports a functional part based on these templates?

features/GRD000_Grid-information.feature
Ghesselink: Undercooked pasta, in this case .. 

It's indeed very easy to miss, and maybe not such an important part. We have the description in the feature file to provide context after all.

features/GRD000_Grid-information.feature
Ghesselink: Yes. I was wondering how close we should stay to the relating concept template.

features/GRD000_Grid-information.feature
Ghesselink: This makes it a normative rule, and not a rule to check for activation. 
Which is part of the reason why this PR is de-activated. It would probably be better to create two new normative rules and then skip the 000-rule. Based on [this issue](https://github.com/buildingSMART/ifc-gherkin-rules/issues/124)

features/GRD000_Grid-information.feature
```diff
@@ -0,0 +1,26 @@
+@implementer-agreement
+@GRD
+@POS
+@version1
+@E00020
+
+Feature: GRD000 - Grid Information
```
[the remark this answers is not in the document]
evandroAlfieri: Sorry, I thought moving the pr back to draft was enough to save you from reviewing it. It's about this rule that we (Geert and I) wanted to talk with you on Monday

features/GRD000_Grid-information.feature
aothms: Well, it goes back to the age old question that we have to second guess the intention behind the template, and this case I'd say it's twofold:

- [GRD000] Describe grid placement functionality: `IfcProduct ----ObjectPlacement---> GridPlacement`
- [GRD001-a] Constrain placement of grid *"a grid placement should be defined relatively to a grid"*: `GridPlacement ----PlacementRelTo----> IfcObjectPlacement ---PlacesObject---> IfcGrid`

And then a third component that is important, but is not in the template:

- [GRD001-b] *"The grid to which the placement is relative should contain the axes that are used for the intersection"*

afbeelding

aothms: Let's phrase it differently, as soon as the model contains:

`IfcProduct ----ObjectPlacement---> GridPlacement`

The file contains a "grid placement" according to the functional part.

But if it's not according to the 2 further specifications I just highlighted, it's not that the model doesn't contain it, it's that the usage is invalid. So yes, these are additional normative parts and we need to create this separation.

aothms: @evandroAlfieri @Ghesselink Annother rule for the backlog is this informal proposition from the docs:

> IntersectingAxes[1] and IntersectingAxes[2] shall not be part of the same row of grid axes, i.e. both shall not be within the same set of [IfcGrid](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcGrid.htm).UAxes or [IfcGrid](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcGrid.htm).VAxes of the corresponding [IfcGrid](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcGrid.htm).

https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcVirtualGridIntersection.htm

PR #258, opened 2024-08-08
IVS-63 - Activation testing updates

Ghesselink:
- In the Gherkin statement for activating a rule, we always return an error instead of a pass.
- In the CI/CD process, we check for errors only when the filename pertains to activation and includes the term "activated." We expect such files to reach the end of their Gherkin statements, as they contain an activating instance or sequence as defined in the concept template. We do not expect non-activating test files to reach this point.
- Ensure that a ValidationOutcome with EXECUTED status is only achieved and passed over to the database when context.instances include meaningful entities or values. Since the structure cannot be generalized, this function needs to be recursive.

Open questions; 

Should we change the following severities? This would probably be more aligned with our terminology
-  `EXECUTED` to `ACTIVATED`'
- '`NOT_APPLICABLE`' to '`NOT_ACTIVATED`

How do we handle PASSED? IIRC, we did not want to populate the database with each passing instance.


features/steps/validation_handling.py
```diff
@@ -116,16 +116,20 @@ def handle_then(context, fn, **kwargs):
...
-        expected=None,
-        feature=context.feature.name,
-        feature_version=misc.define_feature_version(context),
-        severity=OutcomeSeverity.EXECUTED,
-        validation_task_id=context.validation_task_id
-    )
-    context.gherkin_outcomes.append(validation_outcome)
+    #ensure the rule is not activated when there are no instances
+    #in case there are no instances but the rule is applicable (e.g. SPS001), then the rule is still activated and will return either a pass or an error
+    if misc.do_try(lambda: (lambda f: f(f))(lambda f: lambda data: bool(data) and (not isinstance(data, (list, tuple)) or any(f(f)(item) for item in data)))(context.instances), context.applicable):
```
aothms, line 121: Sorry, but this really above my head.

`lambda: (lambda f: f(f))(lambda f: lambda data:` ... `f(f)(item)`

Can you explain what this does and then we try to come up with a simpler way to express this.

test/test_main.py
```diff
@@ -53,6 +53,8 @@ def test_invocation(filename):
 
     if base.startswith("fail-") and not any(description == 'Rule disabled' for description in [result[4] for result in gherkin_results]):
         assert len(results) > 0
+    elif bool(re.search(r'000-activated', base)):
```
aothms, line 56: Alternatively can we just let the files begin with `fail-` `pass-` `activated-`, instead of `pass-por000-activated`. I don't think we expect `fail-...-activated-` cases, or do we?

aothms (changes requested):

test/test_main.py
evandroAlfieri: See @civilx64 ? We definitely need that piece of documentation 😂. 
@aothms I agree with the principle, but I think we don't need "activated" unit tests files that much. But rather the "not-activated" ones. More on that from Scott's doc next week

features/steps/validation_handling.py
Ghesselink: It's a recursive lambda function to check whether the list contains meaningful entities. See also[ Fixed-point combinator](https://en.wikipedia.org/wiki/Fixed-point_combinator) (also includes an example in C++).  I wrote it down simpler before and was wondering if it was possible to write recursive functions without actually defining one (so with lambda instead). Turns out it is. But I agree, it's awful to read ..
The simpler version would be; 
```
def contains_truthy(data):
    if not data:
        return False
    if isinstance(data, (list, tuple)):
        return any(contains_truthy(item) for item in data)
    return True

result = misc.do_try(lambda: contains_truthy(context.instances), context.applicable)
```
The `misc.do_try` part is in case we don't have any context.instances instantiated. In this case, we look for applicability (e.g. correct schema). As a concrete example, (a part of) SPS001;
```
    Given A file with Schema "IFC2X3"

    Then There must be at most 1 instance(s) of IfcSite
```
We don't have any Given statement that selects instances from the model (e.g. Given an IfcSite). The activation criteria here is therefore not related to instances but merely to the schema identifier/MVD. We discussed last Thursday that this could be desirable, but I suppose it's not set in stone. The alternative would be to re-write rules so that there are always activating instances. For this rule, it would mean that 0 instances of IfcSite would mean 'not-activated' and more than 1 IfcSite would be an error. In Gherkin, it will look something like; 
```
Given a correct Schema & MVD 
Given An IfcSite (or 'All instances of IfcSite', similar as GRF001) 
Then the length must be 1 at depth 1
```
Another approach would be to create a '000' rule for each functional part. This way, we clearly define what we mean when we say the software supports a feature. Then, we can use this rule to determine whether a block should be colored grey or green in case the validation passes.

aothms: Very nice, didn't spot the y-combinator immediately, should have, but it's not very pythonic because we're not actually doing lambda calculus. Isn't it the similar to e.g `any(utils.recursively_flatten(context.instances))`? Which I would say is a bit more direct way of writing this.

https://github.com/buildingSMART/ifc-gherkin-rules/blob/f7dcaac762beb0c613d46f29fe94ca8aa778f9dd/features/steps/thens/existance.py#L52

---

> The misc.do_try part is in case we don't have any context.instances instantiated

I would prefer if we not catch arbitrary exceptions for known situations. `getattr(context.instances, [])` would work equally well in that case (?) and doesn't result in silencing arbitrary exceptions that we might actually want to know about.

---

To repeat I like what you did here, but since this involves the formalization of our certification logic I prefer we write it in a way that a typical MBA can almost comprehend this. Lambda calc and y combinators is really a bit too much I think.

---

Final note. In python you can actually have recursive lambdas by just naming them.

```
>>> f = lambda n: 0 if n == 1 else 1 if n == 2 else f(n-2) + f(n-1)
>>>
>>> f(10)
34
```

test/test_main.py
Ghesselink: Yes, in that case we treat 'activation' similar as 'fail'. There is a behave error triggered, but this is only to indicate that the last then statement has been reached. The only prequisite for this that we limit this 'dummy-rule statement' only for activation(000)-rules. This was the plan anyhow. 

I've created a small script that renames files to start with pass-, or activated- instead of activated after the rule-code. So 
`'pass-aaa000-activated_something.ifc'` to `'activated-aaa000-something.ifc' `and 
`'pass-aaa000-not_activated_something_else` to  `'pass-aaa000-something_else'`
. We don’t expect any fail-...-activated cases, but even if they occur, it won’t be an issue (they'll just be flagged as fail files in the CI/CD). I'll share the script here so we can use it for other activation rules as well (the ones that currently have an open PR). As in the example above, it removes the 'activation/not-activation' part after the rule code in the filename, and if the rule is activated (as indicated by the current text after the rule code), change 'pass' to 'activated'.

```
for file in pass-???000-*; do
  if [[ "$file" == *-activated_* ]]; then
    new_file=$(echo "$file" | sed -e 's/^pass-/activated-/' -e 's/-activated_/-/')
    mv "$file" "$new_file"
  elif [[ "$file" == *-not_activated* ]]; then
    new_file=$(echo "$file" | sed -e 's/-not_activated//')
    mv "$file" "$new_file"
  fi
done
```

> but I think we don't need "activated" unit tests files that much. But rather the "not-activated" one

It allows us to check whether a block will turn grey (not-activated) or green (activated) on the platform without having to manually upload files in UAT. I think it's quite important to include in the CI/CD.

features/steps/validation_handling.py
Ghesselink: Thanks for the explanations :-) 

> Isn't it the similar to e.g any(utils.recursively_flatten(context.instances))

Yes. I think I was a bit too enthusiastic about the lambda recursive lambdas and overlooked the more simple solution. 

> getattr(context.instances, []) would work equally well in that case (?) 

I think it could be even simpler. In a couple of lines above, we already define `instance` as
```
instances = getattr(context, 'instances', None) or (context.model.by_type(kwargs.get('entity')) if 'entity' in kwargs else [])
```
Knowing this, we can simply do; 
```
is_activated = any(misc.recursive_flatten(instances)) if instances else context.applicable
    if is_activated:
       # pass to db
```
Which is a lot more comprehensible for a typical MBA. 

---

> Final note. In python you can actually have recursive lambdas by just naming them.

Ah, I thought that lambda functions could not reference themselves. I was also looking to immediately invoke the function. Not really with a y combinator, but an outer lambda function. 
```
>>> (lambda f, n: f(f, n))(lambda f, n: 0 if n == 1 else 1 if n == 2 else f(f, n-2) + f(f, n-1), 10)
34
```

test/test_main.py
evandroAlfieri: This is not what we agreed last Thursday. Activated doesn't correspond to any colour. It's an internal state, users don't care. I would expect test files for not-activated (grey), but why test files for activated? Fail and Pass files are, by definition, activated before they turn one of the two. What am I missing?

Ghesselink: Activation refers to a state indeed where we can color the card either green or red, indicating that the final step of the validation has been reached and there is a (passing or failing) outcome. However, with the current 000 rule, we don't have any failing files. That's why I mentioned that 'activated' corresponds to green, but this is only true within the context of the current rule
Including these test files in the CI/CD allows us to verify whether a file passes because the validation was successfully executed and completed (reaching the final statement) or if the validation stopped because the activation criteria were not met.

evandroAlfieri: Precisely. So for the 000 series, I'd expect only not-activated and pass files. Or not?

Ghesselink: Currently, we're naming the files 'activated' and 'pass', where 'pass' actually means 'pass because it is not activated'. 
We could reverse it to make this clearer. In that case, with the 'pass' file we do expect a behave error (because of the dummy 'Then' statement; we're not using this statement anywhere else), but then in the CI/CD we'd have to differentiate between 'pass in normative rules' and 'pass with 000-series'. Something like (untested)
```
if base.startswith('pass'): 
    if '000' in base: #should be better pattern matching
          assert len(results) > 0
    elif: 
         assert len(results) == 0
elif base.startswith("fail-") and not any(description == 'Rule disabled' for description in [result[4] for result in     gherkin_results]):
        assert len(results) > 0
elif base.startswith('not-activated')
        assert len(results) == 0
   
```
Note that the scope of these files is purely testing in gherkin, not for end-users. However, I'd agree that we should probably try to keep the used terminology consistent as much as possible because to avoid this kind of confusion.

features/steps/validation_handling.py
```diff
@@ -163,20 +181,21 @@ def apply_then_operation(fn, inst, context, current_path, depth=0, **kwargs):
...
-                    feature=context.feature.name,
-                    feature_version=misc.define_feature_version(context),
-                    severity=OutcomeSeverity.PASSED,
-                    instance_id = safe_method_call(activation_inst, 'id', None),
-                    validation_task_id=context.validation_task_id
-                )
-                context.gherkin_outcomes.append(validation_outcome)
+            
+            # Currently, we should not inject passed outcomes for each individual instance to the databse
+            # if not step_results:
```
aothms, line 186: But this is not for individual instances right, there is no loop? This was for the file-wide notification that for a certain feature at least one then step passed. So that in the UI/DB we can take the highest severity,

test/test_main.py
```diff
@@ -44,17 +55,29 @@ def test_invocation(filename):
...
-    if gherkin_results:
+    def print_tabulate(gherkin_results):
         print(tabulate.tabulate(
             [[c or '' for c in r] for r in gherkin_results],
             maxcolwidths=[30] * len(gherkin_results[0]),
             tablefmt="simple_grid"
         ))
+    
+    if gherkin_results and not base.startswith("pass-"):
+        print_tabulate(gherkin_results)
```
aothms, line 66: I wouldn't do this `... and not base.startswith("pass-")`. If we're developing new features and something that we thinks should pass raises an error, it's nice to see that in the console output.

features/steps/validation_handling.py
```diff
@@ -116,16 +117,33 @@ def handle_then(context, fn, **kwargs):
...
+            validation_outcome = ValidationOutcome(
+                outcome_code=ValidationOutcomeCode.EXECUTED,  # "Executed", but not no error/pass/warning #deactivated for now
+                observed="passed",
+                expected=None,
+                feature=context.feature.name,
+                feature_version=misc.define_feature_version(context),
+                severity=OutcomeSeverity.EXECUTED,
+                validation_task_id=context.validation_task_id
+            )
+        context.gherkin_outcomes.append(validation_outcome)
```
aothms, line 145: I think I understand what you're trying to do, but it's rather hacky for two reasons:

- the combination of `OutcomeSeverity.ERROR` and `ValidationOutcomeCode.PASSED`
- the difference in behaviour between production and local testing

This is bound to confuse people in the future, us included.

I don't think it needs to be this way. We just need to figure out how we get information from the gherkin_outcomes into the json log.

I've stumbled upon something called `context.attach` that seems to be a way to propagate some information all the way the json log, but I think we could also write our own JsonFormatter subclass that picks information from the outcomes up.

https://github.com/behave/behave/commit/29b6d471c7d73b894753f85fcf161fa7b5d58105#diff-23cf37bbf1511c05d382150305e9f11f0e2a6bfcfcfd7de211d588c1e54f7585R136

aothms (changes requested):

test/test_main.py
Ghesselink: [We still do](https://github.com/buildingSMART/ifc-gherkin-rules/blob/a787445de8fa18c26a57fa60343a1360edb2775b/test/test_main.py#L80) .In case of a pass testfile, the gherkin_result list is modified to describe that the found error is actually a passed result and print it out afterwards with tabulate.
For example, here we see that a single pass file activates the rule through three scenarios
image

aothms: Thanks for clarifying. I think we should figure out another way though to pass the activation state than through raising an exception. That probably will affect this bit as well.

features/steps/validation_handling.py
Ghesselink: No, I think we agreed that adding an entry for every outcome for each instance would make the database too heavy. Currently, we only use '`passed`' in Gherkin between given statements.
Therefore, the out-commented code can probably be deleted.

`context.instances = list(map(attrgetter('instance_id'), filter(lambda res: res.severity == OutcomeSeverity.PASSED, insts)))`

We could define a rule as _passed_ if it is activated, unless a `fail` outcome is found. In other words, if the highest severity is '`executed`,' we consider the file to have passed the rule.

It might be even clearer to use 'activated' between steps, which would also allow us to pinpoint the last step that was executed. In that case, we'd can explicitly state the passing outcome with the '`passed`' severity instead (and ensure that error has a higher severity as passed).

Ghesselink:
Outdated https://github.com/buildingSMART/ifc-gherkin-rules/pull/267

PR #262, opened 2024-08-15
IVS-144/TAS001 Polygonal face boundary no self intersections

aothms:


aothms:
@evandroAlfieri 

``` 
TAS	Tessellated (i.e. meshes)
 ^       ^
``` 

is this intended?

civilx64:
replaced by #298

civilx64:
> is this intended?

TAS fits well with the long A sound in the second syllable. Maybe it was not originally intended that way but I see no need to revise.

PR #264, opened 2024-08-16
IVS-84/Fix - PSE001 stricter prefix matching

Ghesselink:
The rule has been made stricter to ensure that incorrect property set names now trigger the rule and produce an error. Previously, this activation was not case-sensitive, leading to false positives with the following examples:
- `pset_Wallcommon`
- `PSET_Wallcommon`

For the property set `Pset Wallcommon`, the rule still activates and results in an error as intended.

Additionally:
- An option has been added to check for literal values, allowing a rule to be activated or validated with the same statement. This does not apply to the current rule, where we want activation to be broad while ensuring validation is case-sensitive.
- Several test files have been added to verify these changes.

### DIFF
--- a/features/steps/givens/attributes.py
+++ b/features/steps/givens/attributes.py
@@ -21,6 +21,12 @@ class SubTypeHandling (Enum):
     INCLUDE = auto()
     EXCLUDE = auto()
 
+class StartOrNotStart (Enum):
+    STARTS = auto()
+    NOT_START = auto()
+    STARTS_LITERALLY = auto()
+    NOT_START_LITERALLY = auto()
+
 register_type(include_or_exclude_subtypes=TypeBuilder.make_enum({"including subtypes": SubTypeHandling.INCLUDE, "excluding subtypes": SubTypeHandling.EXCLUDE }))
 register_type(first_or_final=TypeBuilder.make_enum({"first": FirstOrFinal.FIRST, "final": FirstOrFinal.FINAL }))
 register_type(equal_or_not_equal=TypeBuilder.make_enum({
@@ -29,6 +35,11 @@ register_type(equal_or_not_equal=TypeBuilder.make_enum({
     "is not": ComparisonOperator.NOT_EQUAL,
     "is": ComparisonOperator.EQUAL,
 }))
+
+register_type(start_or_not_start=TypeBuilder.make_enum({"starts": StartOrNotStart.STARTS,
+                                                        "does not start": StartOrNotStart.NOT_START,
+                                                        "starts literally" : StartOrNotStart.STARTS_LITERALLY,
+                                                        "does not start literally" : StartOrNotStart.NOT_START_LITERALLY}))
 
 def check_entity_type(inst: ifcopenshell.entity_instance, entity_type: str, handling: SubTypeHandling) -> bool:
     """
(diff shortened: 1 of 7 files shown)


civilx64 (commented):
image

Everything is working as anticipated but the Expected message does not provide any clue that the failure is caused by `PSET_` at the beginning of the `Name` attribute instead of `Pset`. Is there a way to inject that in the `Expected` field?

features/steps/givens/attributes.py
```diff
@@ -21,6 +21,12 @@ class SubTypeHandling (Enum):
     INCLUDE = auto()
     EXCLUDE = auto()
 
+class StartOrNotStart (Enum):
+    STARTS = auto()
+    NOT_START = auto()
+    STARTS_LITERALLY = auto()
+    NOT_START_LITERALLY = auto()
```
aothms, line 28: I'm thinking out loud here, but one thing we could maybe do is to define the enum with a decorator that does the TypeBuilder and register_type. If we use the string labels as the enum values, the enum is fully self contained and a reusable decorator should be enough.

```suggestion
@register_enum_type
class StartOrNotStart (Enum):
    STARTS = 'starts'
    NOT_START = 'does not start'
    STARTS_LITERALLY = 'starts literally'
    NOT_START_LITERALLY = 'does not start literally'
```

features/steps/givens/attributes.py
```diff
@@ -135,14 +146,23 @@ def step_impl(context, file_or_model, field, values):
 def step_impl(context, inst, attribute, tail="single"):
     yield ValidationOutcome(instance_id=getattr(inst, attribute, None), severity = OutcomeSeverity.PASSED)
 
-@gherkin_ifc.step("Its {attribute} attribute {condition} with {prefix}")
+@gherkin_ifc.step("Its {attribute} attribute {condition:start_or_not_start} with {prefix}")
+@gherkin_ifc.step("Its {attribute} attribute {condition:start_or_not_start} with {literal} {prefix}")
```
aothms, line 150: The phrase with `{literal}` does not appear to be used. Was that an earlier experiment?

aothms:
> Everything is working as anticipated but the Expected message does not provide any clue that the failure is caused by PSET_ at the beginning of the Name attribute instead of Pset. Is there a way to inject that in the Expected field?

Agreed.

- I also find the feature text maybe not as explicit as could be. Instead of with or without `literally` we could also consider `in a case sensitive manner` / `insensiitive` . To make it a bit more explicit. Maybe that already helps a bit pointing the user to the PSET/Pset distinction.
- It's also a little bit 'strange' that we're going to be trying to match `"PSET something"` to `"{Pset_WallCommon", ...}` now because we already know that every we're not going to find anything if it's not `Pset_...` verbatimly.

These things combined, I could also imagine we write a new scenario specifically around this:

```
Given its Name attribute starts with "pset" in a case insensitive manner
Then its Name attribute starts with "Pset_" in a case sensitive manner
```

features/steps/givens/attributes.py
Ghesselink: Yes, indeed. There was some discussion on how strict we should be in our matching, I did some experiments with pattern matching (e.g. `Pset_Name`' vs `Pset Name` vs `pset Name`). I put it on hold until a conclusion was made. Let's try to get to one this afternoon.

aothms: Perfect, add it to the agenda so that we're all prepared :)

Ghesselink:
Sounds good, as you suggested before 

```
 Pset_* -> error
[pP][sS][eE][tT]* -> warning
```

We would be able to make sensitive part for errors and the insensitive for warnings, maybe with the tags. 

So for example

```
@E00020 #Error
Scenario : Case sensitive validation for errors 
Given An IfcPropertySet
Given its Name attribute starts with "pset" in a case insensitive manner
Given its Name attribute starts with "Pset_" in a case sensitive manner
Then The IfcPropertySet Name attribute value must use predefined values according to the "pset_definitions" table #e.g.

@W00030 #Warning
Scenario: Case insensitive validation for warnings
Given an IfcPropertySet
Given its Name attribute starts with "pset" in a case insensitive manner
Given its Name attribute does not start with "Pset_" in a case sensitive manner #don't trigger both error and warning
Then The IfcPropertySet Name attribute value must use predefined values according to the "pset_definitions" table #e.g.
```

To-do's would then be 
- Structure the feature file differently (it's not possible to use two Backgrounds in the same feature file). 
- Ensure scenario-tags scenario overwrite feature-wide tags (i.e. @industry-practice / @implementer-agreement) on frontend. 
- While we're add it, give the python implementation a good cleanup. Possibly in different functions or statements; it's a bit difficult to read.

PR #267, opened 2024-08-21
IVS-63 - Activation wrap up

Ghesselink:
Rebase of https://github.com/buildingSMART/ifc-gherkin-rules/pulls
Including changes of QTY000 in new format (na/pass/fail)


main.py
```diff
@@ -143,9 +144,9 @@ def run(filename, rule_type=RuleType.ALL, with_console_output=False, execution_m
                                 if match:
                                     instance_id = '#' + match.group(1)
                                     results = re.sub(pattern, '', str(results))
-                                else:
-                                    instance_id = "-"
-                                yield f"{feature_name}/{scenario_name}.v{version}", f"{remote}/blob/{shas[0]}/{feature_file}", f"{step_name}", instance_id, results
+                                yield f"{feature_name}/{scenario_name}.v{version}", f"{remote}/blob/{shas[0]}/{feature_file}", f"{scenario_name}", f"{step_name}", instance_id, results
+                            elif item['activated']:
```
aothms, line 148: I'm a bit hesitant here on declaring victory at this point and yielding "Rule passed", because there can be more `then` steps in the scenario or more scenarios. Granted, in none of the 000-series we have multiple then steps or multiple scenarios, but I think the logic here needs to be match the more general case.

We can solve this in two ways:
- Don't immediately yield RulePassed when item['activated'], but do so at the end of the outer for loop when making sure that we really never got any `results`.
- Reverse the logic and yield `"Rule activated"` whenever we see `item['activated']` and adapt the logic in test_main.

test/test_main.py
```diff
@@ -34,27 +34,55 @@ def rule_disabled(code: str) -> bool:
...
-        assert len(results) == 0
+                [[c or '' for c in r] for r in gherkin_results],
+                headers = ["Rule", "Location", "Last step", "Scenario", "Instance", "Result"],
+                maxcolwidths=[30] * len(gherkin_results[0]),
+                tablefmt="simple_grid"
+            ))
+    elif rule_is_disabled:
+        print("Rule is disabled")
+    else: 
+        print("Rule not activated by test file")
```
aothms, line 76:
```suggestion
        # Errors and warnings come from exceptions raised in Python and are
        # propagated by the logger. Pass results are emitted in the run() loop
        # when inspecting the json log from behave, when there have been no
        # errors or warnings, but feature.activated is set to true. This flag is
        # managed by our custom JSONFormatter, based on whether a then
        # step is executed over a prior selected set of instances or applicable
        # state originating from given steps. Therefore when results without disabled messages
        # is empty, it means that the rule has not been activated. I.e given statements
        # did not result in an actionable set of instances at the time of the first then step.
        print("Rule not activated by test file")
```

aothms (changes requested):
Great work. My most important comments are regarding:

- `context.step` vs `behave.model.Step`.
- yielding "Rule passed" in the inner loop when there could be in theory more results coming in subsequent iterations of the loops.

enhanced_json_formatter.py
```diff
@@ -0,0 +1,33 @@
...
+    name = 'enhanced_json'
+
+    def feature(self, feature):
+        super().feature(feature) 
+        self.current_feature_data.setdefault("activated", False)
+
+    def result(self, step):
+        super().result(step)
+        if getattr(step, 'activating_feature', False):
+            self.current_feature_data["activated"] = True
```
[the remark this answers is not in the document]
Ghesselink: This probably more clear and consistent with the behavior on the platform indeed; to pass the ValidationOutcome directly to the JSON output and handle it within test_main.py, rather than triggering a Behave error. This means moving away  some more from Behave's built-in functionalities,  but maintains consistency with our own terminology and handling methods. Which is, I think, more important.

test/test_main.py
```diff
@@ -34,27 +34,55 @@ def rule_disabled(code: str) -> bool:
     return next((exclude_disabled_rules(path) for path in feature_paths if os.path.basename(path).lower().startswith(code)), False)
 
 
+def contains_outcome_code(data, code='P00010'):
```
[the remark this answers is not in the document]
Ghesselink: Yes, I used it initially to parse the behave outcome in test_main.py, but decided to go for another sotution and forgot about this part. Deleted it. 
Instead of these tricks, it would probably be better to add the VAlidationOutcome to the JSON output as you suggest in a later comment.

main.py
```diff
@@ -119,7 +119,7 @@ def run(filename, rule_type=RuleType.ALL, with_console_output=False, execution_m
                 version = len(shas)
                 check_disabled = 'disabled' in item['tags']
                 if check_disabled:
-                    yield f"{feature_name}.v{version}", f"{remote}/blob/{shas[0]}/{feature_file}", "Rule disabled", ("Rule disabled", "This rule has been disabled from checking"), "Rule disabled"
+                    yield f"{feature_name}.v{version}", f"{remote}/blob/{shas[0]}/{feature_file}", "Rule disabled", ("Rule disabled", "This rule has been disabled from checking"), "Rule disabled", "Rule disabled"
```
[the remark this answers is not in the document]
Ghesselink: > But did you double check everything works as intended?

Yes, in the validation service itself we don't depend on these indeed. This was the case before. Currently, he validation outcomes are passed directly into the database from the gherkin repo and queried from there in platform the platform frontend.

features/environment.py
```diff
@@ -4,6 +4,7 @@
 import os
 from rule_creation_protocol import protocol
 import json
+import orjson
```
aothms, line 7: Why 2 json libraries? Can we just settle on one? If we use `orjson` we need to make sure all the requirements.txt are updated.

features/environment.py
```diff
@@ -126,5 +127,8 @@ def get_or_create_instance_when_set(spf_id):
 
             ValidationOutcome.objects.bulk_create(outcomes_to_save)
 
-    else: # invoked via console
-        pass
+    else: # invoked via console or CI/CD pipeline
+        outcomes = [outcome.to_dict() for outcome in context.gherkin_outcomes]
+        outcomes_json_bytes = orjson.dumps(outcomes) #encoded to utf-8
```
aothms, line 132: In case you're looking for `b""` byte strings and that's the reason you picked `orjson`. You can also do:

```suggestion
        outcomes_json_bytes = json.dumps({}).encode('utf-8')
```

main.py
```diff
@@ -117,36 +118,24 @@ def run(filename, rule_type=RuleType.ALL, with_console_output=False, execution_m
...
                 shas = get_commits(cwd, feature_file)
                 version = len(shas)
-                check_disabled = 'disabled' in item['tags']
-                if check_disabled:
-                    yield f"{feature_name}.v{version}", f"{remote}/blob/{shas[0]}/{feature_file}", "Rule disabled", ("Rule disabled", "This rule has been disabled from checking"), "Rule disabled", "Rule disabled"
-
+                rule_is_disabled = 'disabled' in item['tags']
+                yield { # add feature information, mainly to check disabled feature
+                    'feature_name': f"{feature_name}.v{version}",
+                    'remote' : f"{remote}/blob/{shas[0]}/{feature_file}",
```
aothms, line 124: I don't think we depend on this anymore.

main.py
```diff
@@ -117,36 +118,24 @@ def run(filename, rule_type=RuleType.ALL, with_console_output=False, execution_m
...
-                                yield f"{feature_name}/{scenario_name}.v{version}", f"{remote}/blob/{shas[0]}/{feature_file}", f"{scenario_name}", f"{step_name}", instance_id, results
-                            elif item['activated']:
-                                yield f"{feature_name}/{scenario_name}.v{version}", f"{remote}/blob/{shas[0]}/{feature_file}", f"{scenario_name}" , f"{step_name}", instance_id, "Rule passed"
+                    scenario_validation_outcomes = json.loads(base64.b64decode(el.get('validation_outcomes', [{}])[0].get('data', '')).decode('utf-8')) if el.get('validation_outcomes') else []
+                    for validation_outcome in scenario_validation_outcomes:
+                        validation_outcome.update({ # add scenario information to display in terminal
+                            'scenario_name': el['name'],
+                            'step_names': [step['name'] for step in el['steps']]
+                        })
+                        yield validation_outcome
```
aothms, line 138: I think it's a little bit nicer to move this out of the loop.


```suggestion
                    scenario_info = {
                            'scenario_name': el['name'],
                            'step_names': [step['name'] for step in el['steps']]
                        }
                    for validation_outcome in scenario_validation_outcomes:
                        yield validation_outcome | scenario_info
```

(indentation needs to be fixed)

main.py
```diff
@@ -117,36 +118,24 @@ def run(filename, rule_type=RuleType.ALL, with_console_output=False, execution_m
...
-                        instance_id = "-"
-                        if step_status and step['step_type'] == 'then':
-                            try:
-                                results = step['result']['error_message']
-                            except KeyError:  # THEN not checked
-                                results = []
-                            except json.decoder.JSONDecodeError:  # THEN not checked
-                                results = []
-                            if results:
-                                pattern = r'ifc_instance_id=(\d+)'
```
aothms, line 141: It's great that we're getting rid of this. Great job!

test/utils.py
```diff
@@ -36,4 +37,36 @@ def collect_test_files(top_order_string = False, insert_args = False):
...
+
+    hyphen = Literal("-")
+
+    three_letters = Word(alphas, exact=3)
+    three_digits = Word(nums, exact=3)
+    letters_digits_combination = Combine(three_letters + three_digits)
+
+    remainder = Word(alphas + nums + "_.-")
+
+    filename_structure = (
```
aothms, line 64: Nicely done

aothms (approved):

main.py
aothms: The `'remote'` I mean

features/environment.py
Ghesselink: [Reversed this change already in a later commit](https://github.com/buildingSMART/ifc-gherkin-rules/commit/f1d9f9b534bae5531cb7e5c9f26f8cdc8040d242). Orjson is a bit faster as the output is already in bytes. But I don't think time is really an issue here and it indeed led to some more unnecessary work updating libraries.

features/environment.py
Ghesselink: That was indeed the reason and how it's resolved :)

main.py
Ghesselink: No, [deleted it ](https://github.com/buildingSMART/ifc-gherkin-rules/commit/101ec619eb8815e7f074479aaf4199ab1a8f3800)
The feature version is also unused at the moment. I left it in since I have a feeling we'll use this versioning more in the future.

main.py
Ghesselink: [That's nicer indeed, thanks](https://github.com/buildingSMART/ifc-gherkin-rules/commit/101ec619eb8815e7f074479aaf4199ab1a8f3800).

PR #271, opened 2024-09-10
IVS-117 outcome massaging

aothms:


features/environment.py
```diff
@@ -68,9 +67,14 @@ def get_validation_outcome_hash(obj):
...
     # Given steps may introduce an arbitrary amount of stackframes.
     # we need to clean them up before behave starts appending new ones.
+    old_outcomes = getattr(context, 'gherkin_outcomes', [])
     while context._stack[0].get('@layer') == 'attribute':
         context._pop()
+    # preserve the outcomes to be serialized to DB in after_feature()
+    context.gherkin_outcomes = old_outcomes
+
 
+def after_feature(context, feature):
```
aothms, line 77: This is the main change. We don't want N/A results + X results in the dashboard at the feature level, but we're taking max severity at the scenario level.

features/environment.py
```diff
@@ -81,20 +85,15 @@ def reduce_db_outcomes(feature_outcomes):
...
                 outcome_counts = Counter(outcome.severity for outcome in context.gherkin_outcomes)
-
                 for severity in [OutcomeSeverity.PASSED, OutcomeSeverity.EXECUTED, OutcomeSeverity.NOT_APPLICABLE]:
                     if outcome_counts[severity] > 0:
-                        for outcome in context.gherkin_outcomes:
-                            if outcome.severity == severity:
-                                return [outcome]
-
-            return []
+                        yield next(o for o in context.gherkin_outcomes if o.severity == severity)
```
civilx64, line 93:
```suggestion
                        yield next(outcome for outcome in context.gherkin_outcomes if outcome.severity == severity)
```
For clarity and to match with what we previously had

civilx64 (approved):

PR #272, opened 2024-09-10
Always emit N/A outcomes; they are not written to db when higher severity outcomes exist

aothms:
Not sure whether tests pass on this. Please have a look @Ghesselink but we shouldn't depend on contexr.applicable for setting the initial n/a state. This context.applicable is very likely to be true, only in case of mvd/schema related things it can be false. Whereas what we care about for the state in the DB is the selected instances at the time of the first then statement.


Ghesselink:
The tests pass, and the issue from ASM000 is also resolved. 
After checking locally, we always display all of the rules for each uploaded file. 

image

aothms:
Ok, the fact that we both independently wrote the same PR probably qualifies as a review ;)

PR #273, opened 2024-09-10
IVS-77 - Always include at least one Validation Outcome

Ghesselink:
We currently have two mechanisms to prevent Behave from executing subsequent steps:

However, the platform does not display a rule if:

- Empty List: If a list is empty, the following steps will always yield empty results.
- `context.applicable `Variable: This is used to skip steps that are irrelevant, such as when a schema doesn't apply.

This is not the desired behavior. Instead, we should always log a ValidationOutcome with a status of "NotApplicable." Since the platform searches for the maximum severity, adding this outcome will not affect other results negatively.
This ensures that all rules are displayed for every uploaded file.

### DIFF
--- a/features/steps/validation_handling.py
+++ b/features/steps/validation_handling.py
@@ -251,20 +251,18 @@ def execute_step(fn):
         Data is circulated using the 'behave-context' and is ultimately stored in the database, as 'ValidationOutcome' corresponds to a database column.
         """
 
-        if not getattr(context, 'applicable', True):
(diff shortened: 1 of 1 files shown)


aothms:
Ok, hurray we litterally just opened the same PR https://github.com/buildingSMART/ifc-gherkin-rules/pull/272 Good that we agree!

aothms:
Duplicate of #272

PR #281, opened 2024-09-15
IVS-79 - LAY000

Ghesselink:
Superceding https://github.com/buildingSMART/ifc-gherkin-rules/pull/256

Note; in some cases the attribute LayerAssignment is singular ('LayerAssignment' as opposed to the plural 'LayerAssignment`s`') as is the case in IfcExtrudedAreaSolid. This is handled by Scenario Outlines. 

Furthermore, since we have four scenarios (two outlines with two examples each). The two outlines are for layer assignments added to the [representation](https://github.com/buildingSMART/ifc-gherkin-rules/blob/19b6dc469f89dee05dff1b39b6e8a984b43ea460/features/LAY000_Presentation-layer-assignment.feature#L12) and [items](https://github.com/buildingSMART/ifc-gherkin-rules/blob/19b6dc469f89dee05dff1b39b6e8a984b43ea460/features/LAY000_Presentation-layer-assignment.feature#L27).

> There is one additional consideration, that's a bit in detail, but actually you can have layer assignment on the representation but also on the items. This is missing in the concept template, but is rendered in the documentation.

image

It also would be nice to quickly pinpoint the scenario activating the rule. Currently, we have to navigate and set pinpoints within the gherkin repository to retrace this. Embedding `validation_outcomes` to JSON in environment.py works, but everything is handled together in `after_feature` which the embeds the Outcome to the feature rather than the last step or scenario. Two ways of handling this: 
- The most obvious -> adding a last_step and scenario entry in the datamodel (e.g. to ValidationOutcome)
- Only in DEV -> Adding the current scenario in the decorator to the outcome that is appended to context.gherkin_outcomes. [Probably here](https://github.com/buildingSMART/ifc-gherkin-rules/blob/31df6539adce05ed9a5b4966ac3ee804b2895791/features/steps/validation_handling.py#L134)


aothms:
> Note; in some cases ...

But these cases are already different scenarios. 2nd scenario through items is singular (not only attribute name, also max cardinality of the set). 1st scenario on representation directly is plural.  So these different scenarios don't need to be outlines with examples.

afbeelding

afbeelding

aothms:
> adding a last_step and scenario entry in the datamodel (e.g. to ValidationOutcome)

can we solve this in a single number? Like the nth step from the file counting downwards.

Ghesselink:
> But these cases are already different scenarios. 2nd scenario through items is singular (not only attribute name, also max cardinality of the set). 1st scenario on representation directly is plural. So these different scenarios don't need to be outlines with examples.

I was unsure whether singular/plural was consistent, so thought just to be on the safe side. 
The [outlines are removed](https://github.com/buildingSMART/ifc-gherkin-rules/pull/281/commits/a010958c2f1a85cf4c6bf06b9e5f9d097f0860bc)from the feature, so there are only two separate scenarios.

aothms (approved):

Issue #279, opened 2024-09-13
IfcTextLiteralWithExtent failing check IFC102


jmirtsch:
There is a validation check that nominates IfcTextLiteral as deprecated.
https://github.com/buildingSMART/ifc-gherkin-rules/blob/31df6539adce05ed9a5b4966ac3ee804b2895791/features/IFC102_Absence-of-deprecated-entities.feature#L42

It seems that IfcTextLiteral was deprecated in IFC4 (https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcTextLiteral.htm), proposed to be made abstract.
IfcTextLiteralWithExtent was not deprecated.  https://standards.buildingsmart.org/IFC/RELEASE/IFC4/FINAL/HTML/link/ifctextliteralwithextent.htm

In IFC4.3, IfcTextLiteral is included in RV and ABV
https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcTextLiteral.htm
Although IfcTextLiteralWithExtent is not tagged.
https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcTextLiteralWithExtent.htm

We are using IfcTextLiteralWithExtent representation in export for 4.3.   But it triggers 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/31df6539adce05ed9a5b4966ac3ee804b2895791/features/IFC102_Absence-of-deprecated-entities.feature#L42

Can this rule please be adjusted to not flag non compliance for IfcTextLiteralWithExtent
The nomination of IfcTextLiteralWithExtent for mvds rather than the base class would seem appropriate.

Thanks,

Jon

aothms:
Thanks for bringing this up. I think there have been other cases in IFC4 implying: "deprecated for direct instantiation, intended to become abstract in the future".

I see two options:

- For me this is not standard usage of the term "deprecation". So an option is to remove this deprecation from the schema altogether.
- Alternatively we make the list checked as exact types, instead of type + subtypes. Most of these items are leaf entities anyway. This is probably the safest way forward.

@Ghesselink I see there is also selects in the list: a select can never be instantiated so can't be in this list, also this one is already deleted in 4.3

[...]

PR #282, opened 2024-09-16
IVS-131 - Issue #279 - IFC102 improvements & cleanup validation_handling

Ghesselink:
Addresses https://github.com/buildingSMART/ifc-gherkin-rules/issues/279

- Remove `IfcPresentationStyleSelect` -> is deleted in IFC4X3
- Implement the following.
> Alternatively we make the list checked as exact types, instead of type + subtypes. Most of these items are leaf entities anyway. This is probably the safest way forward.

We should probably also do this for attributes of entities (?). For this, there's a [new small decorator to register enumeration types](https://github.com/buildingSMART/ifc-gherkin-rules/blob/b0000b23f4418f3832691fda9e8f69462ec6b412/features/steps/validation_handling.py#L48). Using this is quite intuitive; simply [decorate the Enum-inherited class](https://github.com/buildingSMART/ifc-gherkin-rules/blob/b0000b23f4418f3832691fda9e8f69462ec6b412/features/steps/thens/existance.py#L9) with `@register_enum_type),`. There is probably no need to use `TypeBuilder` (correct me if I'm wrong)

Furthermore, a[ testfile is added ](https://github.com/buildingSMART/ifc-gherkin-rules/blob/IVS-131-Fix-279-PSE001-Cleanup_handling/test/files/ifc102/pass-ifc102-ifc4x3_deprecated_entity_IfcTextLiteralWithExtent.ifc)to ensure IfcTextLiteralWithExtent is passing.

- Small cleanup for validation_handling: 
  -  Remove unused functions
  -  Functions are in a more logical order now
  -  Add some documentation


aothms (changes requested):
You can remove IfcPresentationStyleSelect from all listings, because:
 - it's never of that type, because you can't instantiate a select
 - even if the is_a() check would check whether the inst is one of the entities possible to be selected by the select, it's not the behaviour we want, because even if the select is deprecated, the types it selects are not. Deprecating a select in the schema is basically meaningless...

```
>>> import ifcopenshell
>>> f = ifcopenshell.file()
>>> f.createIfcCurveStyle()
#1=IfcCurveStyle($,$,$,$,$)
>>> f[1].is_a('IfcPresentationStyleSelect')
False
```

Ghesselink:
Another nuance I was not aware of .. Thanks :) 
I'm assuming the same is true for [IfcStyleAssignmentSelect](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/schema/ifcpresentationappearanceresource/lexical/ifcstyleassignmentselect.htm), [they're both deleted from the deprecated list. ](https://github.com/buildingSMART/ifc-gherkin-rules/pull/282/commits/a5d30d962b8e7cd0f05aa789a116eb15149ea499)

We're now checking for exact types for deprecated entities. Just to be sure, I think this doesn't apply for the deprecated attributes of entities? 
To give an example, for `IfcRelAssigns`, the RelatedObjectsType attribute is deprecated. This will then also be the case for its subtypes; `IfcRelAssignsToActor`, `RelAssignsToGroup`, and so forth. The same applies to [IfcReinforcingElement](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcReinforcingElement.htm) -> IfcReinforcingBar, Mesh, Anchor, Conduit and IfcTendon,

aothms:
> To give an example, for IfcRelAssigns, the RelatedObjectsType attribute is deprecated. This will then also be the case for its subtypes; IfcRelAssignsToActor, RelAssignsToGroup, and so forth. The same applies to [IfcReinforcingElement](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcReinforcingElement.htm) -> IfcReinforcingBar, Mesh, Anchor, Conduit and IfcTendon,

Correct!

aothms (approved):

PR #284, opened 2024-09-22
IVS-134/ALB010 - Alignment Nesting Referents

Ghesselink:
+ if `*requirement* `is in Given statement and _context.instances_ is not empty, the rule should be activated (i.e. the step is treated as a Then step within a sequence of Given steps)


features/ALB010_Alignment-nesting-referents.feature
```diff
@@ -0,0 +1,17 @@
+@implementer-agreement
+@ALB
+@POS
+@RFT
+@version1
+@E00100
+Feature: ALB010 - Alignment Nesting Referents
+The rule verifies that Alignment nest IfcReferent's, such as stations or mileage points.
```
civilx64, line 8:
```suggestion
The rule verifies that each alignment nests at least one IfcReferent, such as stations or mileage points. 
```

features/ALB010_Alignment-nesting-referents.feature
```diff
@@ -0,0 +1,17 @@
+@implementer-agreement
+@ALB
+@POS
+@RFT
+@version1
+@E00100
+Feature: ALB010 - Alignment Nesting Referents
+The rule verifies that Alignment nest IfcReferent's, such as stations or mileage points. 
+These can be used as semantic entities holding information about locations along the alignment curve.
```
civilx64, line 9:
```suggestion
These can be used as semantic entities holding information about locations along the alignment. 
```

features/ALB010_Alignment-nesting-referents.feature
```diff
@@ -0,0 +1,17 @@
+@implementer-agreement
+@ALB
+@POS
+@RFT
+@version1
+@E00100
+Feature: ALB010 - Alignment Nesting Referents
+The rule verifies that Alignment nest IfcReferent's, such as stations or mileage points. 
+These can be used as semantic entities holding information about locations along the alignment curve. 
+This is done using the IfcRelNests relationship.
```
civilx64, line 10:
```suggestion
IfcReferent is associated to IfcAlignment via the IfcRelNests relationship.
```

features/steps/validation_handling.py
```diff
@@ -289,6 +282,25 @@ def should_apply(items, depth):
     # evokes behave error
     generate_error_message(context, [gherkin_outcome for gherkin_outcome in context.gherkin_outcomes if gherkin_outcome.severity in [OutcomeSeverity.WARNING, OutcomeSeverity.ERROR]])
 
+
```
civilx64, line 285: I don't understand the need for breaking out `activate_rule` as a separate function.  ALB010 and SPS007 are the only two rules that use `*required*` and both do it in a `Then` step.  What am I missing?

test/files/alb010/fail-alb010-alignments_not_nesting_referents.ifc
```diff
@@ -0,0 +1,3103 @@
+ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION (('ViewDefinition [Alignment-basedReferenceView]'), '2;1');
+FILE_NAME ('fail-ala003-scenario01-different_horizontal_segment_geometry_types.ifc', '2023-12-31T23:59:59', ('redacted'), ('redacted'), 'redacted', 'redacted', 'redacted');
```
civilx64, line 4:
```suggestion
FILE_NAME ('fail-alb010-alignments_not_nesting_referents.ifc', '2023-12-31T23:59:59', ('redacted'), ('redacted'), 'redacted', 'redacted', 'redacted');
```

test/files/alb010/pass-alb010-alignments_nesting_referents.ifc
```diff
@@ -0,0 +1,3104 @@
+ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION (('ViewDefinition [Alignment-basedReferenceView]'), '2;1');
+FILE_NAME ('fail-ala003-scenario01-different_horizontal_segment_geometry_types.ifc', '2023-12-31T23:59:59', ('redacted'), ('redacted'), 'redacted', 'redacted', 'redacted');
```
civilx64, line 4:
```suggestion
FILE_NAME ('pass-alb010-alignments_nesting_referents.ifc', '2023-12-31T23:59:59', ('redacted'), ('redacted'), 'redacted', 'redacted', 'redacted');
```

features/ALB010_Alignment-nesting-referents.feature
```diff
@@ -0,0 +1,17 @@
+@implementer-agreement
```
civilx64, line 1:
```suggestion
@industry-practice
```

This is a best practice, not a normative rule.

civilx64 (commented):
Testing on a local deployment of the validation service and after adjusting the feature file to `@industry-practice` I get the expected pass and warning results.

features/steps/validation_handling.py
```diff
@@ -174,6 +174,10 @@ def handle_given(context, fn, **kwargs):
     3) Filter the set of IfcAlignment based on a value ('Given attribute == X' -> [IfcAlignm, None, IfcAlignm])
     4) Set instances to a given attribute ('Given its attribute Representation') -> [IfcProdDefShape, IfcProdDefShape, IfcProdDefShape]
     """
+    if '*required*' in context.step.name and context.instances:
+        # in cases *required* is in the step name, the rule is activated, as the step is treated as a 'Then' step
```
aothms, line 178: > is treated as a 'Then' step

In this case it is a then step? Ah same comment by scott below....

features/steps/validation_handling.py
Ghesselink: In gcert, we used this statement in a sequence of Given steps. For example, [here](https://github.com/buildingsmart-community/gherkin-test-cases/blob/development/features/testcases/CHILLER01M.feature). 
```
        Given an <IfcEntity>
        Given Name = <Name>
        Given A *required* relationship IfcRelAggregates to <IfcEntity> from <RelatingObject> and following that
        Given Its attribute Name

        Then Assert existence
        Then The value must be "<RelatingName>"
```
Initially, I thought we'd need to use the step this way for ALB010 as well. However, during implementation, I realized that activation wouldn't work in this case because the rule is only triggered in the `Then` step, not in the `Given` step. That's why I split the function, as it's used in two different places in the code. Later, I realized that for this rule, it would be sufficient to place the step at the end of the sequence, making it a `Then` step.

I'm unsure whether to keep the function. The advantage is that it could be very if we encounter this situation again in the future, which isn't unlikely. The disadvantage, as noticed now, is that it could be confusing. 
Should we remove it? If we decide to keep it, I'll add a note summarizing this context.

cc for @aothms for asking the same question below

civilx64: [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

These changes should be reverted, then we can bring the function back when we actually need it.

Ghesselink: Fair point, learned something new and  [reverted](https://github.com/buildingSMART/ifc-gherkin-rules/pull/284/commits/855bce36003a4e10632b3ce19f5a8c234ef4e8b5) it.

civilx64 (approved):

PR #288, opened 2024-09-24
IVS-122 Testfiles Revision 02 wrong number (4) in guid

Ghesselink:
```
The base64 encoding process may differ from common base64 implementations. The following steps are used:

    The first byte is encoded in the first two characters
    The remaining bytes are encoded in groups of 3, taking up 4 characters

As a result, the first character must be either a 0, 1, 2, or 3.
```

A couple of testfiles started with '4'


civilx64 (approved):

aothms:
One step closer to a `Given an IfcRoot, Then the GlobalId attribute is a valid base64 encoded rfc4122` guid :)

PR #289, opened 2024-09-24
IVS-133/PJS003 - Globally Unique Identifiers

Ghesselink:
- Implement PJS003
- Rewrite implementation and add documentation to 'The attribute X must start with Y' step. It's easier to read now and hopefully more future-proof
- After the decorator, we can use step implementations for both Given and Then. Does the ordering of step implementations in the folders `given` `then` and `steps` still suffice, or should we organize the implementations differently?
- It would be useful to implement an option to: 
  - Execute a specific gherkin rule to an external testfile, e.g. `'python -m gherkin_validate --file testfile.ifc --rules pjs alb'`
  - Execute a specific rule to a whole map of folders, e.g.  `python -m gherkin_validate --folder testfile_folder --rules pjs alb`
  - Return the output in JSON?


features/steps/thens/attributes.py
```diff
@@ -1,4 +1,5 @@
 import operator
+import re
```
civilx64, line 2:
```suggestion
```

This is an unused import and should be removed.

test/files/pjs003/fail-pjs003-IFC4_3_ADD2_GuidTests.ifc
```diff
@@ -0,0 +1,42 @@
...
+#16= IFCSIUNIT(*,.LENGTHUNIT.,.MILLI.,.METRE.);
+
+
+/* SITES AGGREGATED TO PROJECT */
+#29=IFCRELAGGREGATES('2te7nJfPvJyvPEahDDrZW0',#2,'ProjectContainer',$,#1,(#29990,#30000,#30010,#30020,#30030,#30040));
+#31=IFCLOCALPLACEMENT($,#11);
+
+ 
+#29990=IFCSITE('14oxcl5LHMZQ6iCWJKpe5w',#2,'SITE_withCorrectGuid',$,$,#31,$,$,.COMPLEX.,$,$,$,$,$);
+#30000=IFCSITE(30000,#2,'SITE_withIncorrectGuid','GUID is not a string',$,#31,$,$,.COMPLEX.,$,$,$,$,$);
```
civilx64, line 33:
```suggestion
#30000=IFCSITE('30000',#2,'SITE_withIncorrectGuid','GUID is not a string',$,#31,$,$,.COMPLEX.,$,$,$,$,$); 
```
`GlobalId` needs to be a string to avoid schema errors.

test/files/pjs003/fail-pjs003-IFC4_3_ADD2_GuidTests.ifc
```diff
@@ -0,0 +1,42 @@
...
+#29=IFCRELAGGREGATES('2te7nJfPvJyvPEahDDrZW0',#2,'ProjectContainer',$,#1,(#29990,#30000,#30010,#30020,#30030,#30040));
+#31=IFCLOCALPLACEMENT($,#11);
+
+ 
+#29990=IFCSITE('14oxcl5LHMZQ6iCWJKpe5w',#2,'SITE_withCorrectGuid',$,$,#31,$,$,.COMPLEX.,$,$,$,$,$);
+#30000=IFCSITE(30000,#2,'SITE_withIncorrectGuid','GUID is not a string',$,#31,$,$,.COMPLEX.,$,$,$,$,$); 
+#30010=IFCSITE('14oxcl5',#2,'SITE_withIncorrectGuid','Invalid Length',$,#31,$,$,.ELEMENT.,$,$,$,$,$); 
+#30020=IFCSITE('1-+=cl5LHMZQ6iCWJKpe5w',#2,'SITE_withIncorrectGuid','Invalid Characters',$,#31,$,$,.PARTIAL.,$,$,$,$,$); 
+#30030=IFCSITE('b4oxcl5LHMZQ6iCWJKpe5w',#2,'SITE_withIncorrectGuid','Invalid Starting Character',$,#31,$,$,.COMPLEX.,$,$,$,$,$); 
+#30040=IFCSITE('14oxcl5LHMZQ6iCWJKpe5w',#2,'SITE_withIncorrectGuid','Duplicated GUID',$,#31,$,$,.COMPLEX.,$,$,$,$,$);
```
civilx64, line 37:
```suggestion
#30040=IFCSITE('04qxcl5LHMZQ7iCWJKpe5w',#2,'SITE_withIncorrectGuid','Duplicated GUID',$,#31,$,$,.COMPLEX.,$,$,$,$,$); 

```

`GlobalId` needs to be unique within a model to avoid schema errors.

civilx64 (commented):
Looks good!  There are a few changes required to one unit test file so that it passes the schema checks.  Ready for UAT after that is addressed.

features/PJS003_Globally-Unique-Identifiers.feature
```diff
@@ -0,0 +1,33 @@
...
+
+    Then <Constraint>
+
+    Examples:
+    | Constraint |
+    | The string length must be exactly "22" characters |
+    | The characters must be within the official encoding character set |
+
+
+  Scenario: Correct first character for IFC4X3_ADD2
```
aothms, line 28: This is not only for 4.3 add2. This has this way since IFC 1 ;)

Ghesselink: Ah, how I read it it was only for ifc4.3_add2. That makes the feature even easier to write; [all the steps are in the same scenario now](https://github.com/buildingSMART/ifc-gherkin-rules/pull/289/commits/5fcbc719ee492b85137b0458d0baf036a6bff1df).

features/PJS003_Globally-Unique-Identifiers.feature
```diff
@@ -0,0 +1,26 @@
+@implementer-agreement
+@PJS
+@version2
```
evandroAlfieri, line 3:
```suggestion
@version1
```
Right?

features/PJS003_Globally-Unique-Identifiers.feature
```diff
@@ -0,0 +1,33 @@
...
+@E00010
+Feature: PJS003 - Globally Unique Identifiers
+  The rule verifies that the GUID of each element adheres to the Global Unique Identifier format 
+  and ensures compliance with constraints that are not yet validated by other methods. 
+  Specifically, the characters must be within the official encoding character set
+  "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_$,"
+  and the resulting string must be exactly 22 characters in length. 
+  Additionally, starting from IFC4.3_ADD2, the first character must be either 0, 1, 2, or 3.
```
evandroAlfieri, line 11:
```suggestion
  , the resulting string must be exactly 22 characters in length, and the first character must be either 0, 1, 2, or 3.
```

test/files/pjs003/fail-pjs003-IFC4_3_ADD2_GuidTests.ifc
Ghesselink: True, this was the testfile that I initially used. Two of them were indeed triggered schema-errors already indeed, three were to be implemented by us.[ I've deleted the two lines](https://github.com/buildingSMART/ifc-gherkin-rules/pull/289/commits/c79b759ffe4363be3200e12a65b9b1e2830b2c7d)
Furthermore, the validation stopped when `inst` had an invalid type. We should probably implement something that prevents the validation from silently stopping for the rest of the instances in the sequence when one of them causes an error within the step implementation. For now, I've implemented a[ solution for the current step_imp](https://github.com/buildingSMART/ifc-gherkin-rules/pull/289/commits/c79b759ffe4363be3200e12a65b9b1e2830b2c7d).

test/files/pjs003/fail-pjs003-IFC4_3_ADD2_GuidTests.ifc
Ghesselink: See answer above

civilx64 (approved):

PR #291, opened 2024-09-25
IVS-135 Fix/Enforce tag hierarchy

Ghesselink:
- Add missing tags
- A protocol error is now triggered when a tag is missing (e.g. a `ALS` rule without a `GEM` tag)
- Protocol errors are embedded to JSON, causing a more straightforward error message

image


features/rule_creation_protocol/tags_hierarchy.py
```diff
@@ -0,0 +1,5 @@
+tags_hierarchy = {
```
aothms, line 1: Is it possible to move this to the data-model repo? We need to have access to this in the score card repo as well.

Ghesselink: Sure :) 
I've added [these to the data-model repo](https://github.com/buildingSMART/ifc-validation-data-model/pull/21/) .
I thought we could probably also use more info regarding the functional parts to colour the blocks (is also added in the same PR)

Ghesselink:
Superceded by https://github.com/buildingSMART/ifc-gherkin-rules/pull/291

PR #300, opened 2024-10-10
IVS-137 ALB012 Vertical Segment Radius of Curvature

### LINKED ISSUE #241: ALB012 IfcAlignmentVerticalSegment.RadiusOfCurvature
Not sure if you are intending to check IfcAlignmentVerticalSegment.RadiusOfCurvature since it is an optional parameter. I found a case where the VS doesn't recognize the RadiusOfCurve is not consistent with the curve parameters.

`#119=IFCALIGNMENTVERTICALSEGMENT($,$,365.76,487.68,36.8808,0.0175,-0.01,100.0,.PARABOLICARC.);`

Radius of curvature should be 487.68/(-0.01 - 0.0175) = -17733.81818.

I changed the value to 100.0 to see if the VS would detected it.

[FHWA_Bridge_Geometry_Alignment_Bridge_1.ifc.txt](https://github.com/user-attachments/files/16289800/FHWA_Bridge_Geometry_Alignment_Bridge_1.ifc.txt)

civilx64:
Fixes #241

Also adds an additional check per Note 3 on the `RadiusOfCurvature` attribute in the 
[IFC 4.3.2.0 schema documentation](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/lexical/IfcAlignmentVerticalSegment.htm).


features/ALB012_Alignment-vertical-segment-radius-of-curvature.feature
```diff
@@ -0,0 +1,20 @@
...
+    Given A relationship IfcRelNests from IfcAlignmentVertical to IfcAlignmentSegment and following that
+    Given Its attribute DesignParameters
+
+  Scenario: Validating the absence of curvature radius for constant gradient vertical segment
+    Given PredefinedType != 'ARC' or 'PARABOLICARC'
+    Then The value of attribute RadiusOfCurvature must be empty
+
+  Scenario: Validating the radius of curvature for parabolic segments
+    Given PredefinedType = 'PARABOLICARC'
+    Then RadiusOfCurvature value must be equal to the expression: HorizontalLength / ( EndGradient - StartGradient )
```
aothms, line 20: Not really consistent wording with:

> The value of attribute RadiusOfCurvature must be empty

```suggestion
    Then The value of attribute RadiusOfCurvature must be equal to the expression: HorizontalLength / ( EndGradient - StartGradient )
```

features/steps/thens/attributes.py
```diff
@@ -143,3 +143,169 @@ def step_impl(context, inst):
...
+                else:
+                    raise Exception('Expression attribute value is an IFC entity')
+            else:
+                raise Exception('Expression attribute not found')
+
+    # Evaluate the string expression using eval
+    try:
+        expression_value = eval(expression)
+    except Exception as e:
+        raise ValueError(f"Error evaluating expression: {e}")
```
aothms, line 200: I've asked this before but what's currently the behaviour of raising exceptions?

What happens for example with a `ZeroDivisionError`.

Regarding that specifically should there be some rules to prevent zeros in the denominator in this equation?

features/ALB012_Alignment-vertical-segment-radius-of-curvature.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  Background:
+    Given A model with Schema "IFC4.3"
+    Given An IfcAlignmentVertical
+    Given A relationship IfcRelNests from IfcAlignmentVertical to IfcAlignmentSegment and following that
+    Given Its attribute DesignParameters
+
+  Scenario: Validating the absence of curvature radius for constant gradient vertical segment
+    Given PredefinedType != 'ARC' or 'PARABOLICARC'
+    Then The value of attribute RadiusOfCurvature must be empty
```
aothms, line 16: The correspondence of `IfcAlignmentVertical <- nests -> Segment -> IfcAlignmentVerticalSegment` is assessed in ALB002. But we don't have rule dependencies. Are we sure that violations of ALB002 don't cause crashes in ALB012 causing the entire validation task to fail?

features/steps/thens/attributes.py
```diff
@@ -143,3 +143,169 @@ def step_impl(context, inst):
...
+                expected=f"A value {comparison_op} {expression_value}",
+                observed={attr_compared_value},
+                severity=OutcomeSeverity.ERROR,
+            )
+    except ValueError as e:
+        raise ValueError(f"Error during comparison: {e}")
+
+@gherkin_ifc.step('First instance {first_expression} value must be {comparison_op} the second instance {second_expression}')
+@gherkin_ifc.step('First instance {first_expression} value must be {comparison_op} the second instance {second_expression} at depth 1')
+def step_impl(context, inst, first_expression:str, comparison_op:str, second_expression:str):
```
aothms, line 227: I don't see a usage of this impl function

features/steps/utils/geometry.py
```diff
@@ -222,3 +222,30 @@ def alignment_segment_angular_difference(
...
+    The valid comparison operators are:
+        'equal to';
+        'not equal to';
+        'greater than';
+        'less than';
+        'greater than or equal to';
+        'less than or equal to'. 
+    '''
+    if comparison_operator == 'equal to':
+        return math.isclose(value_1, value_2, abs_tol=precision)
```
aothms, line 239: `rel_tol` has a default value. Should we set it to `0`? https://docs.python.org/3/library/math.html#math.isclose

features/steps/thens/attributes.py
civilx64: Correct, and thanks for the catch.  We originally had a number of additional scenarios and this one handles the check of `StartDistAlong` that I mentioned in Slack.  See also IVS-150 in JIRA.  I would like to leave it in for now so that it's there fore future rules.

features/ALB012_Alignment-vertical-segment-radius-of-curvature.feature
civilx64: ALB002 is currently disabled with an item in backlog to be re-activated.  I added a link to this comment to the ALB002 reactivation ticket IVS-75 so that I remember to verify independence of ALB002 and ALB012 in UAT.

features/ALB012_Alignment-vertical-segment-radius-of-curvature.feature
civilx64: It looks like this was done so that the implementation could stay separate from what was existing previously.  I have started work on combining the new 'expression' capabilities to the existing implementation.  ATM the new code path is working but at expense of now failing when processing the original implementations.

features/steps/thens/attributes.py
```diff
@@ -143,3 +143,169 @@ def step_impl(context, inst):
...
+        '*' : operator.mul,
+        '/' : operator.truediv,
+        '%' : operator.mod,
+        '**' : operator.pow,
+        'equal to' : operator.eq,
+        'not equal to' : operator.ne,
+        'greater than' : operator.gt,
+        'less than' : operator.gt,
+        'greater than or equal to' : operator.ge,
+        'less than or equal to' : operator.le,
```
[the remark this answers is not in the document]
civilx64: `equal to` is being matched for the second scenario so should be OK to leave as-is

civilx64:
Need to step away for a bit.  Will revisit soon to finalize.

features/ALB012_Alignment-vertical-segment-radius-of-curvature.feature
aothms: That appears even worse. Is this logic never checked?

https://github.com/buildingSMART/ifc-gherkin-rules/blob/31df6539adce05ed9a5b4966ac3ee804b2895791/features/ALB002_Alignment-layout.feature#L53-L64

I'm just worried about the AttributeErrors when DesignParameters is not of type IfcAlignmentVerticalSegment

civilx64: I agree, for some reason I thought it was IfcAlignmentVerticalSegment that nests IfcAlignmentVertical and that IfcAlignmentSegment was a supertype.

Do we have an implementation "Given attribute is of type {entity_type}?  I looked briefly earlier and didn't find anything.

Regardless of where this lands I will add additional fail files so that we are certain that the type check is handled correctly for both ALB002 and ALB012.

features/steps/utils/geometry.py
civilx64: From the docs:
>  rel_tol must be greater than zero.

This appears correct to me.

aothms: The code disagrees. I really prefer to be explicit about this. If we want a relative tolerance in addition to an absolute tolerance I prefer that we write it down in the code as an act of explicitly documenting our behaviour. I think only an absolute tolerance better matches the intent of the IFC spec (even if that's very limited).

```
>>> import math
>>> math.isclose(1.0, 1.0, rel_tol=0.)
True
>>> math.isclose(1.0, 1.0, rel_tol=-1.0)
ValueError: tolerances must be non-negative
```

features/steps/thens/attributes.py
civilx64: > IIRC, in case there is an error unrelated to the rule (e.g. a ZeroDivisionError as mentioned earlier), it's a bit annoying for the gherkin CI/CD but without consequences for subsequent rules.

It would be nice to have something in the decorator or validation_handling that could return an ERROR outcome whenever an exception is raised.  I started to replace all of the `raise` statements with `yield ValidationOutcome(... ValidationSeverity.ERROR)` but that felt like a non-pythonic brute force approach.

features/steps/utils/geometry.py
civilx64: I see.  Implemented in 515cc490c25ee619937491a4ba1ed0b76103f20c.

features/ALB012_Alignment-vertical-segment-radius-of-curvature.feature
civilx64: Coming back to the original question:

> Are we sure that violations of ALB002 don't cause crashes in ALB012 causing the entire validation task to fail?

Yes, we are sure now because of the additional Given statement in the feature's Background.

features/steps/thens/attributes.py
civilx64: > What happens for example with a ZeroDivisionError.

It is not caught in CI/CD.  There is nothing in place to handle the exception and therefore the process fails before it can yield a ValidationOutcome with severity ERROR.  Current pytest logic is still `0 errors = pass`, not `all results are ValidationOutcome(ValidationSeverity.PASSED)`.

image

features/steps/thens/attributes.py
```diff
@@ -60,37 +60,124 @@ def accumulate_errors(i):
...
-    elif not pred(attribute_value, value):
-        yield ValidationOutcome(
-            inst=inst,
-            expected=None if not value else value,
-            observed=misc.recursive_unpack_value(attribute_value),
-            severity=OutcomeSeverity.ERROR
-        )
+@gherkin_ifc.step('The value of attribute {attribute} must be {value_or_comparison_op}')
+@gherkin_ifc.step('The value of attribute {attribute} must be {value_or_comparison_op} {display_entity:display_entity}')
+@gherkin_ifc.step('The value of attribute {attribute} must be {value_or_comparison_op} the expression: {expression}')
```
fg55b, line 65: Do you think that using this with constant values directly in place of the expression would be confusing? For example, “The value of attribute <RadiusOfCurvature> must be equal to the expression: 0”?

I ask because, in ALB011, the StartRadiusOfCurvature and EndRadiusOfCurvature of IfcAlignmentHorizontalSegment need to be zero or non-zero according to its PredefinedType. This cannot be verified with the first expression because "The value of attribute <RadiusOfCurvature> must be 0.0" would compare the 0.0 float value of the IFC file with the 0.0 string value declared on the rule. It also does not consider the precision.

civilx64: > “The value of attribute must be equal to the expression: 0”

While maybe not ideal, this pattern would be fine IMHO.

aothms: I would first try to fit it into

https://github.com/buildingSMART/ifc-gherkin-rules/blob/development/features/steps/thens/attributes.py#L63

Could be as simple as:

```
>>> import ast
>>> required = '1.0'
>>> try: required = ast.literal_eval(required)
... except: pass
...
>>> type(required)
<class 'float'>
```

But maybe also take into account tolerances when you detect you're comparing floats

fg55b: > > “The value of attribute must be equal to the expression: 0”
> 
> While maybe not ideal, this pattern would be fine IMHO.

This would be particularly useful to evaluate if StartRadiusOfCurvature and EndRadiusOfCurvature of IfcAlignmentHorizontalSegment is not zero for nonlinear PredefinedTypes, as the "not equal to" operator can be used.


> I would first try to fit it into
> 
> https://github.com/buildingSMART/ifc-gherkin-rules/blob/development/features/steps/thens/attributes.py#L63
> 
> Could be as simple as:
> 
> ```
> >>> import ast
> >>> required = '1.0'
> >>> try: required = ast.literal_eval(required)
> ... except: pass
> ...
> >>> type(required)
> <class 'float'>
> ```
> 
> But maybe also take into account tolerances when you detect you're comparing floats

The comparison using tolerances, after evaluating the string, can be done using https://github.com/buildingSMART/ifc-gherkin-rules/blob/42959f2d1b1d4defcd6fa8ec592eddb4a4f63977/features/steps/utils/geometry.py#L227

However, the expression as is would only compare using the "equal to" operator, which is not applicable to this case.

features/steps/thens/attributes.py
```diff
@@ -60,37 +60,124 @@ def accumulate_errors(i):
...
+        '**' : operator.pow,
+        'equal to' : operator.eq,
+        'not equal to' : operator.ne,
+        'greater than' : operator.gt,
+        'less than' : operator.gt,
+        'greater than or equal to' : operator.ge,
+        'less than or equal to' : operator.le,
+    }
+
+    if expression is not None:
```
Ghesselink, line 96: Can probably be simply
```suggestion
    if expression:
```

features/steps/thens/attributes.py
```diff
@@ -60,37 +60,124 @@ def accumulate_errors(i):
...
+        '/' : operator.truediv,
+        '%' : operator.mod,
+        '**' : operator.pow,
+        'equal to' : operator.eq,
+        'not equal to' : operator.ne,
+        'greater than' : operator.gt,
+        'less than' : operator.gt,
+        'greater than or equal to' : operator.ge,
+        'less than or equal to' : operator.le,
+    }
```
Ghesselink, line 94: We're using operators at quite some points in the code now. What do you think of moving them to a dataclass and seperating them into arithmetic operators, comparison operators and the various interpretations of operators. 
Maybe not in the scope of this PR, but what about a ticket with these kind of cleanup/maintanance tasks?

civilx64:
> A few things to discuss in order to wrap up the PR:
> 
>     1. Raising and handling Exceptions versus ValidationOutcomes with ERROR severity
> 
>     2. Avoiding ZeroDivisionError.  One solution would be an addition additional scenario to enforce StartGradient = EndGradient for CONSTANTGRADIENT only.

Exception handling added in 6a3c82faa4f5f88871f93ece47fa74b9372cc53f.  Thanks @Ghesselink!

features/steps/thens/attributes.py
civilx64: Ticket is good.  This PR is turning into a marathon and I'd like to wrap up.

features/steps/thens/attributes.py
civilx64: Ref: [Zen of Python](https://en.wikipedia.org/wiki/Zen_of_Python) - 

> Explicit is better than implicit

I say let it stand.

features/steps/thens/attributes.py
civilx64: resolved in 6a3c82faa4f5f88871f93ece47fa74b9372cc53f

features/steps/utils/geometry.py
```diff
@@ -222,3 +222,31 @@ def alignment_segment_angular_difference(
...
+    elif comparison_operator == 'greater than':
+        return value_1 > value_2 and not math.isclose(value_1, value_2, rel_tol=0., abs_tol=precision)
+    elif comparison_operator == 'less than':
+        return value_1 < value_2 and not math.isclose(value_1, value_2, rel_tol=0., abs_tol=precision)
+    elif comparison_operator == 'greater than or equal to':
+        return value_1 > value_2 or math.isclose(value_1, value_2, rel_tol=0., abs_tol=precision)
+    elif comparison_operator == 'less than or equal to':
+        return value_1 < value_2 or math.isclose(value_1, value_2, rel_tol=0., abs_tol=precision)
+    else:
+        raise ValueError(f"Invalid comparison operator: {comparison_operator}")
```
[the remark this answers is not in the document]
civilx64: Actually worse IMO.  Replaced with pattern matching in f6fc490ea19d5296869596bde6fc1743d5234faf.

Ghesselink (approved):
There's [ a small side-track discussion/experiments](https://github.com/buildingSMART/ifc-gherkin-rules/commit/c382fe798c9db4c5b1b68dc8f04868db0662c46a#commitcomment-148641694). It's regarding how to access the `instance_id` in JSON and, finally, our tabulate table we use for visualizing the ci/cd. I think this should not block ALB012 for the next release. 
If that discussion will result in more commits, those can be moved to a different PR.

PR #301, opened 2024-10-14
IVS-138 improve ALS016 message

### LINKED ISSUE #285: ALS016 reporting message
The reporting message isn't so helpful for trying to troubleshoot.  It reports the observed discrepancy.
Would it be possible to also make it report the expected coordinate (and bearing for a tangency check) and the file defined coordinate (and bearing when appropriate)?

For an alignment with many segments, it's near impossible to troubleshoot.

civilx64:
fixes #285

### DIFF
--- a/features/steps/thens/alignment.py
+++ b/features/steps/thens/alignment.py
@@ -7,7 +7,7 @@ import ifcopenshell.entity_instance
 import ifcopenshell.util.unit
 
 from utils import ifc43x_alignment_validation as ifc43
-from utils import geometry
+from utils.geometry import AlignmentSegmentContinuityCalculation
 from utils import ifc
 from validation_handling import gherkin_ifc
 from . import ValidationOutcome, OutcomeSeverity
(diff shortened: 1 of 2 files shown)


aothms:
Looks good to me :)

I would probably return (instead of store; so the class becomes stateless) `positional_difference` and `directional_difference` as a dataclass and use `dataclasses.asdict()` instead of manually composing the dict. But at this point in time there is no clear advantage.

civilx64:
> I would probably return (instead of store; so the class becomes stateless) `positional_difference` and `directional_difference` as a dataclass

Implemented in 7d3104748263a8021fe044068a361165f1e44be4

> and use `dataclasses.asdict()` instead of manually composing the dict. But at this point in time there is no clear advantage.

I did this originally but it fails.  enitity_instances are not able to be pickled since they originate from the SWIG wrapper.  Therefore a custom `to_dict()` method is required.

PR #304, opened 2024-10-17
IVS-165 dont raise on earlier errors

aothms:


Ghesselink:
Nice. Using our new JSON formatter, I've added some modifications to also display the correct scenario, last step and instance id to our gherkin console output (there was a similar issue here). 

From (display last scenario rather than scenario for which the error occured)
image
To (correct scenario, last step and instance_id)
image

Ghesselink (approved):

PR #306, opened 2024-10-18
ALS017 alignment vertical segment geometric continuity (IVS-53)

civilx64:


civilx64:
A minor revision will be needed on the front end as well to display "gradient" instead of "tangent direction":
image

Ghesselink (approved):
Looks good. Nice to see that we can develop rules now with 12 steps with very minimal modifications in the step implementation.
 All the schema and syntax errors are passing.

PR #314, opened 2024-10-23
IVS-178/VER000 - Versioning and Revision Control

Ghesselink:


features/VER000_Versioning-and-revision-control.feature
```diff
@@ -0,0 +1,31 @@
...
+@E00020
+
+Feature: VER000 - Versioning and revision control
+    The rule verifies the capability to track changes to building data over time and maintain a comprehensive history of those changes.
+    https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Attributes/Revision_Control/content.html
+
+
+    Scenario Outline: Check for activation
+
+        Given an IfcOwnerHistory
```
aothms, line 13:
```suggestion
        Given an IfcRoot
        Given its attribute OwnerHistory
```

I would start from IfcRoot. It's functionally a bit better because otherwise the OwnerHistory could still be somewhere floating around without actually being used in the model.

features/VER000_Versioning-and-revision-control.feature
```diff
@@ -0,0 +1,31 @@
...
+        Given its attribute <attribute>
+
+        Then The IFC model contains information on the selected functional part
+
+        Examples: 
+            | attribute |
+            | LastModifiedDate |
+            | LastModifyingUser |
+            | LastModifyingApplication |
```
aothms, line 22: What do we mean with these scenario outlines? As far as I can tell, only one scenario is necessary to pass the check and have the box coloured green?

It seems to me you'd rather have these combined into one scenario so that only when these three attributes are set and ChangeAction is not "NOTDEFINED" then the box lights up.

aothms (changes requested):

features/VER000_Versioning-and-revision-control.feature
Ghesselink: I thought so, I was not sure. 

We didn't have this `and` operator when looking for attributes of entity instances, so I thought this would be a good time [to create it](https://github.com/buildingSMART/ifc-gherkin-rules/pull/314/commits/3b19ad4205b3d536c70f7b0658f343b49ca1ee3d) ..
The commit is a bit long, but the most important is the modified scenario; 
```
        Given an IfcRoot
        Given its attribute "OwnerHistory"
        Given its attribute "LastModifiedDate" and "LastModifyingUser" and "LastModifyingApplication" exists
        Given ChangeAction is not "NOTDEFINED"

```
Along with the [associated step implementation](https://github.com/buildingSMART/ifc-gherkin-rules/blob/758c94f14a1fef547e4af93d4e849854dc9d7f53/features/steps/givens/attributes.py#L157-L188)

---

I struggled a bit with the parsing of multiple attributes in a single line. What worked in other statements [(e.g.)](https://github.com/buildingSMART/ifc-gherkin-rules/blob/758c94f14a1fef547e4af93d4e849854dc9d7f53/features/steps/givens/attributes.py#L191) didn't work now. 
I thought it would be the most straighforward to wrap attributes within quotes. So 
```
Given its attribute Representation
will be 
Given its attribute "Representation"
```

Perhaps we should be more consistent with this, and also quote attributes in statements like; 
```
UserDefinedOperationType = not empty
A relationship IfcRelNests must exist from IfcAlignment to IfcReferent
The value of attribute RepresentationIdentifier must be "FootPrint" or Axis
```
To 
```
"UserDefinedOperationType" = not empty
A relationship "IfcRelNests" must exist from IfcAlignment to IfcReferent
The value of attribute "RepresentationIdentifier" must be "FootPrint" or Axis
```

Ghesselink:
Superceded by https://github.com/buildingSMART/ifc-gherkin-rules/pull/321

PR #319, opened 2024-10-31
IVS-197 - Fix / GEM004 - IfcTopologyRepresentation

Ghesselink:
- Consider valid types of [IfcTopologyRepresentation](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcTopologyRepresentation.htm) as opposed to IfcShapeRepresentation
- Add testfiles for topology representation, one per schema
- Organize valid types and identifiers per schema in resource folder
- Version bump GEM004 & GEM052 (?)


features/GEM052_Correct-geometric-subcontexts.feature
```diff
@@ -24,8 +24,8 @@ Reference: https://github.com/buildingSMART/Sample-Test-Files/issues/137.
 
         Examples: 
             | schema | source |
-            | IFC4.3 | valid_RepresentationIdentifier_IFC4.3.csv |
```
aothms, line 27: In this case we can just remove the outline. and start with the same given as above:

Given a model with Schema "IFC4.3" or "IFC4"

features/resources/IFC4/valid_ProjectDeclaration.csv
aothms, line 1: I don't understand why this is in the PR.

features/resources/IFC4/valid_TopologyRepresentationType.csv
aothms, line 1: I feel like there is still a bit of an open question here. For these items in topologyrepresentation, which ContextType/Identifier are supposed to be used in Ifc(Geometric)RepresentationContext?

https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/annex_e/structural-analysis-model/structural-curve-member.html doesn't use subcontexts, so also doesn't pass our rules, but I don't think there is a reasonable geometric item that fits these topologies.

I think we should try and get structural-curve-member.ifc to be meaningful and valid for all geometric contexts checks before we can make a final decision here.

aothms (changes requested):
Nice work, but as said in my last comment, I think this alludes to some bigger inconsistency issues regarding topology representation that we should get a better feel for first.

features/GEM052_Correct-geometric-subcontexts.feature
Ghesselink: That's true. I think we can even remove the schema selection from the Given statements; they are already selected in the python implementation. So then, we'll simply have three scenario's; 
- Shape representation identifiers
- Shape representation types
- Topology representation types

Something like [this](https://github.com/buildingSMART/ifc-gherkin-rules/pull/319/commits/ceee0cb293a88dd666d96a1fbad3ee4d5beeb37c)?

features/resources/IFC4/valid_ProjectDeclaration.csv
Ghesselink: It's from PJS002 and uses the same gherkin implementation as GEM004 (the current rule) and GEM052. 

The gherkin impl won't find `resources/valid_ProjectDeclaration.csv `anymore because it'll always look for `resourses/schema/csv_file.csv.`

I thought to duplicate the csv and move it to the folder of each schema would be the clearest.

features/resources/IFC4/valid_TopologyRepresentationType.csv
civilx64: > For these items in topologyrepresentation, which ContextType/Identifier are supposed to be used in Ifc(Geometric)RepresentationContext?

`ContextIdentifier` should be 'Reference' per [CT 4.1.7.5.1](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Topology_Representation/Reference_Topology/content.html) which also corresponds to the value for `RepresentationIdentifier` used for IfcTopologyRepresentation in the example model.

As you've noted, IfcTopologyRepresentation does not have to be (but may be) in a GeometricRepresentationContext.  So `IfcRepresentationContext` and `IfcGeometricRepresentationContext` are both valid.

[IfcGeometricRepresentationContext](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcGeometricRepresentationContext.htm) lists the recognized values of `ContextType` as 'Model', 'Plan', and 'NotDefined'.
Requiring 'Model' as the ContextType for topology representations seems to be the best fit.

features/GEM052_Correct-geometric-subcontexts.feature
civilx64, line 1: Perhaps this is misunderstanding on my end, but doesn't GEM052 require a different set of csv files, e.g. `valid_ContextType.csv` and `valid_ContextIdentifier.csv`?

Ghesselink: Not necessarily. I remember this was a topic of discussing, and the final decision was use the same list for the identifiers and types for both (shape) representations and contexts.

features/resources/IFC4/valid_RepresentationType.csv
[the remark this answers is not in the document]
Ghesselink: [Commits speak louder than words](https://github.com/buildingSMART/ifc-gherkin-rules/pull/319/commits/eb8403d36dbc849a46949af355b410120af058f8)
Thoughts?

features/resources/IFC4/valid_TopologyRepresentationType.csv
Ghesselink: I'm not completely sure I understand, but [I've added a modified ifc file to the unit tests ](https://github.com/buildingSMART/ifc-gherkin-rules/pull/319/commits/bbaed90e236a48b6ef84a4ff4f5d601f5984067e)that; 
- Passes the geometric contexts checks (GEM052)
- Uses 'Reference' as Identifier
- Uses 'Model' as Type
- We can hopefully use to derive more testfiles for afterwards

> should be 'Reference' per [CT 4.1.7.5.1](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Topology_Representation/Reference_Topology/content.html)

I read 'Structural activities `may` have a 'Reference' '.  'May' is rather vague.

PR #326, opened 2024-11-24
IVS-235 - Bulk insert ModelInstances

rw-bsi:
Requires DB constraint to be in place to prevent duplicates / FK constraint errors
(see [PR 27](https://github.com/buildingSMART/ifc-validation-data-model/pull/27))

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -110,7 +110,7 @@ def after_feature(context, feature):
                 task = ValidationTask.objects.get(id=context.validation_task_id)
                 model_id = task.request.model.id
 
-                stepfile_ids = sorted(set(o.instance_id for o in outcomes_to_save))
+                stepfile_ids = sorted(set(o.instance_id for o in outcomes_to_save if o.instance_id))
                 for stepfile_id in stepfile_ids:
                     if stepfile_id:
                         instance = ModelInstance(
--- a/features/environment.py
+++ b/features/environment.py
@@ -86,6 +86,7 @@ def after_feature(context, feature):
     execution_mode = context.config.userdata.get('execution_mode')
     if execution_mode and execution_mode == 'ExecutionMode.PRODUCTION': # DB interaction only needed during production run, not in testing
         from validation_results import OutcomeSeverity, ModelInstance, ValidationTask
+        from django.db import transaction
 
         def reduce_db_outcomes(feature_outcomes):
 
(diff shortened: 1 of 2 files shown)


features/environment.py
```diff
@@ -102,36 +103,31 @@ def reduce_db_outcomes(feature_outcomes):
...
-                # Previously we have the current model SPF id in the instance_id
-                # field. This needs to be updated to an actual foreign key into
-                # our ModelInstances table.
-                if outcome.instance_id:
-                    outcome.instance_id = inst_id_mapping[outcome.instance_id]
-
-            ValidationOutcome.objects.bulk_create(outcomes_to_save)
+        outcomes_instances_to_save = list()
+
+        if outcomes_to_save:
```
aothms, line 108:
```suggestion
        if outcomes_to_save and context.validation_task_id is not None:
```

Any reason why you removed the `... and context.validation_task_id is not None`? The code is a bit spaghetti with all the operating modes, I guess it doesn't hurt to keep it in?

features/environment.py
```diff
@@ -102,36 +103,31 @@ def reduce_db_outcomes(feature_outcomes):
...
+            with transaction.atomic():
+                task = ValidationTask.objects.get(id=context.validation_task_id)
+                model_id = task.request.model.id
+
+                stepfile_ids = sorted(set(o.instance_id for o in outcomes_to_save))
```
aothms, line 113: Because we later do: `if stepfile_ids:` on line 122, I'd rather filter out falsy items before so that stepfile_ids is really empty and not something like `[None]` which would evaluate to true, but have an outcomes_instances_to_save be empty.

aothms, line 113:
```suggestion
                stepfile_ids = sorted(set(o.instance_id for o in outcomes_to_save if o.instance_id))
```

features/environment.py
```diff
@@ -102,36 +103,31 @@ def reduce_db_outcomes(feature_outcomes):
...
+                for stepfile_id in stepfile_ids:
+                    if stepfile_id:
+                        instance = ModelInstance(
+                            stepfile_id=stepfile_id,
+                            model_id=model_id
+                        )
+                        outcomes_instances_to_save.append(instance)
```
aothms, line 120:
```suggestion
                  instance = ModelInstance(
                      stepfile_id=stepfile_id,
                      model_id=model_id
                  )
                  outcomes_instances_to_save.append(instance)
```

Then the if is no longer needed here

aothms (changes requested):
I think the combination of ModelInstance and ValidationOutcome bulk_creation better be wrapped in a transaction.atomic? Just so that the chance of orphaned instances is reduced even more.

Also, I still don't really know how this now actually works with transactions across multiple tasks in parallel and the isolation level of the transactions.  It would be good to at least conceptually understand when multiple tasks in parallel try to create the same instances simultaneously. Should we set the isolation level to `Serializable`?

features/environment.py
rw-bsi: it's at best redundant, worst it silently fails because of a failed assertion (has outcomes, but somehow task_id is lost?) and I prefer to bomb out loudly :-)

features/environment.py
rw-bsi: Will update

rw-bsi:
> I think the combination of ModelInstance and ValidationOutcome bulk_creation better be wrapped in a transaction.atomic? Just so that the chance of orphaned instances is reduced even more.
> 
> Also, I still don't really know how this now actually works with transactions across multiple tasks in parallel and the isolation level of the transactions. It would be good to at least conceptually understand when multiple tasks in parallel try to create the same instances simultaneously. Should we set the isolation level to `Serializable`?

There is a transaction.atomic block at line 109, no harm to push it down, but has same effect?

`serializable` is not required imo - each instance is linked to one or more outcomes, writing instances with `ignore_conflicts` **everywhere** prevents duplicates (by definition only from something running in parallel), and reading them all again makes sure we get current and previously created instances. Key is we do it everwhere the same: schema and gherkin_rules

```
ModelInstance.objects.bulk_create(outcomes_instances_to_save, ignore_conflicts=True) # ignore conflicts with existing
model_instances = dict(ModelInstance.objects.filter(model_id=model_id).values_list('stepfile_id', 'id')) # retrieve all
```

PR #328, opened 2024-12-09
Fixes ALS017

civilx64:
- newer version of `ifcopenshell` corrects issue with parabolic vertical curves in imperial units
- change in step implementation now uses the correct values for `u` in `evaluate(u)` for calculating placement transforms at the ends of alignment segments.


features/steps/utils/geometry.py
```diff
@@ -188,23 +188,35 @@ class AlignmentSegmentContinuityCalculation:
...
+                # Horizontal representation - use the SegmentLength directly
+                u = abs(self.previous_segment.SegmentLength.wrappedValue)
+            case "IFCGRADIENTCURVE" | "IFCSEGMENTEDREFERENCECURVE":
+                # for vertical and cant representations (IfcGradientCurve and IfcSegmentedReferenceCurve),
+                # u should be the total distance traversed by the segment along the base curve
+                # (length of the segment projected to the "Distance Along" x-axis).
+                # The actual distance along the parent curve (e.g. IfcLine, IfcCircle, IfcPolynomialCurve, etc.) will be longer.
+                # Therefore, we will get the projected distance based on the placements of the two segments.
+                u = abs(self.segment_to_analyze.Placement.Location.Coordinates[0] - \
+                    self.previous_segment.Placement.Location.Coordinates[0])
```
aothms, line 207: Isn't this a bit cyclical? We want to assess basically whether segment to analyse and previous segment align, but if we do that based on assessing the respective placements we would miss some cases where the length of previous is actually not enough to reach the placement of current?

In other words: this assesses continuity of vertical in the Y axis, but doesn't account for gaps in the domain of the vertical profile, Is that a problem?

civilx64: Potentially.  I had gotten wrapped around the axle with confusing length along the base curve versus segment length of the parent curve.  Using the placements was the quickest way to resolve this as well as confirm my understanding of the piecewise function implementation.

The cleanest way forward IMHO is to add a property `endPoint` with type `taxonomy::cast<taxonomy::matrix4d>` to [curve_segment_evaluator](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/ifcgeom/mapping/IfcCurveSegment.cpp#L99) given that we have both `length_` and `projected_length_` available.  Then (I think) we can just call `pwf.end_point()` from python to get the calculated point at the end of the segment - versus dealing with the complication of `length` versus `projected_length` on the python side.

I'm keen to take this on as it will confirm my burgeoning C++ capabilities :wink:.

.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install behave pytest tabulate pyparsing sqlalchemy numpy pydantic pydot sqlalchemy_utils django python-dotenv deprecated pandas pyspellchecker rtree
-        wget -O /tmp/ifcopenshell_python.zip https://s3.amazonaws.com/ifcopenshell-builds/ifcopenshell-python-`python3 -c 'import sys;print("".join(map(str, sys.version_info[0:2])))'`-v0.8.1-0267e2b-linux64.zip
+        wget -O /tmp/ifcopenshell_python.zip https://s3.amazonaws.com/ifcopenshell-builds/ifcopenshell-python-`python3 -c 'import sys;print("".join(map(str, sys.version_info[0:2])))'`-v0.8.1-eadf35f-linux64.zip
```
Ghesselink, line 36:
```suggestion
        wget -O /tmp/ifcopenshell_python.zip https://s3.amazonaws.com/ifcopenshell-builds/ifcopenshell-python-`python3 -c 'import sys;print("".join(map(str, sys.version_info[0:2])))'`-v0.8.1-1d27161-linux64.zip
```
Updated ifcopenshell to this version on the [DEV](https://github.com/buildingSMART/ifc-gherkin-rules/commit/7572e156ecff5b992e7406e83d77bc7039b19803) branch. We should probably also update it here to avoid overwriting.
This version seem also to reflect well your and Rick's work on ifcopenshell

civilx64: We actually need something newer that was just brewed today.  See 3f65633f8d7be02f272d103959227c7070c0fd5d.

features/steps/utils/geometry.py
civilx64: [IfcOpenShell PR#6006](https://github.com/IfcOpenShell/IfcOpenShell/pull/6006) now provides correct values of `function_item.length()` for vertical alignment (IfcGradientCurve) segments.  Our calling code was adjusted accordingly.

civilx64:
I also made some adjustments / corrections to the unit test files:

1. Normalized the vector components of IfcDirection.  (Technically they do not have to be normalized but in practice they usually are).
2. Corrected the values of SegmentLength for parabolic vertical curve segments.  SegmentLength is along the parabola, and not just the difference in x between PVT and PVC.

aothms (approved):
Nice work for taking it all the way to C++ to track down the issues here!

PR #334, opened 2024-12-28
GEM112 no duplicate points on indexed poly (IVS-302)

aothms:


features/GEM112_No-duplicated-points-within-an-indexed-poly-curve.feature
```diff
@@ -0,0 +1,11 @@
...
+@GEM
+@version1
+Feature: GEM112 - No duplicated points within an indexed poly curve
+The rule verifies that all the indexed poly curves will have no duplicate points.
+
+  @E00050
+  Scenario: Agreement on no duplicated points within an indexed poly curve
+
+    Given An IfcIndexedPolyCurve
+    Then It must have no consecutive points that are coincident after taking the Precision factor into account
```
civilx64, line 11:
```suggestion
    Then It must have no points that are coincident after taking the Precision factor into account
```
The step implementation compares all points in the poly curve,

features/steps/thens/geometry.py
```diff
@@ -234,6 +234,21 @@ def step_impl(context, inst: ifcopenshell.entity_instance, clause: str):
                 yield ValidationOutcome(inst=inst, observed=(points_coordinates[i], points_coordinates[j]),
                                         severity=OutcomeSeverity.ERROR)
 
+
+@gherkin_ifc.step("It must have no consecutive points that are coincident after taking the Precision factor into account")
```
civilx64, line 238: The comment and implementation seem to agree that all points are being compared, but the text of the step indicates that only consecutive points are being compared.

It seems that the word "consecutive" should be removed from this step and added to the one above, making it
`"It must have no consecutive duplicate points {clause} first and last point"`

civilx64 (commented):
~~Use of the word "consecutive" in the text of this step is confusing.  See specific in-line comments.~~

Edit: word should remain as-is; implementation should use `pairwise()` to be more pythonic.

features/GEM112_No-duplicated-points-within-an-indexed-poly-curve.feature
```diff
@@ -0,0 +1,11 @@
+@implementer-agreement
+@GEM
+@version1
+Feature: GEM112 - No duplicated points within an indexed poly curve
+The rule verifies that all the indexed poly curves will have no duplicate points.
```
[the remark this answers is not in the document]
aothms: Good idea. I copied the wording from GEM111 earlier though, so I've added the same clarification there. 0aeae1c006dfe613b40ccdee914b85b2bf164268

features/GEM112_No-duplicated-points-within-an-indexed-poly-curve.feature
aothms: Really only the consecutive ones it seems:

```
    for i, j in [(i-1, i) for i in range(1, len(points_coordinates))]:
        if math.dist(points_coordinates[i], points_coordinates[j]) < precision:
```

i.e distance between the ith and i-minus-1th point. Did I overlook sth?

features\steps\thens\geometry.py:257

aothms:
Thanks for the review @civilx64 Please have a look at my comment regarding consecutive, are we looking parts of the code?

features/GEM112_No-duplicated-points-within-an-indexed-poly-curve.feature
civilx64: Agree, it is checking consecutive points per informal proposition 1 in [IfcIndexedPolyCurve](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcIndexedPolyCurve.htm).  

Please rewrite with the python [pairwise](https://docs.python.org/3/library/itertools.html#itertools.pairwise) built-in.  This is what we use to compare two consecutive alignment segments in the ALS functional part.

Do we have something in the backlog to check informal proposition 2, that three points used as `ArcIndex` are not colinear?

aothms:
>  If you would kindly refactor the code to use pairwise() per my comment 

I liked the idea but I got stuck.

In short, things in IFC are never what it seems.

Indexed Polygonal Curve is:
- 1 not always polygonal (`Segments` attribute is either a line seg or a circular arc)
- 2 not always indexed (`Segments` attribute is optional in which case it just uses the points in the order provided without indexing)

This combination makes it so that what would seemingly make things more explicit and first-principled, actually only muddies the water with more exceptions. Or at least I didn't get it to work in a satisfying manner.

I've merged the dev branch into this though

PR #335, opened 2024-12-28
GEM113 no colinear arcs

aothms:
Very similar to the face planarity check (but opposite, faces need to be planar, arcs need to be not collinear). Expect conflicts with the two PRs, but just because both add novel stuff in the same place. No functional conflicts.


Ghesselink:
Installed [mpmath](https://github.com/buildingSMART/validate/commit/d9a0e48806d1af3b919c420535bab32ad92fdb9d) on the validate platform

Doing some UAT on the platform, the testfiles behave as expected (checking the header policy checks at the same time) 
image
image
image
image

Ghesselink (approved):

features/GEM113_Indexed-poly-curve-arcs-must-not-be-defined-using-colinear-points.feature
```diff
@@ -0,0 +1,11 @@
+@implementer-agreement
+@GEM
+@version1
+Feature: GEM113 - Indexed poly curve arcs must not be defined using colinear points
+The rule verifies, that all the three points of any IfcArcIndex segment of an IfcIndexedPolyCurve are not colinear after taking the Precision factor into account
```
evandroAlfieri, line 5:
```suggestion
The rule verifies that all the three points of any IfcArcIndex segment of an IfcIndexedPolyCurve are not colinear after taking the Precision factor into account
```

PR #337, opened 2025-01-15
IVS-315 Quadratic complexity in GEM003

aothms:
This must have been an old one, but this is the only case I could find where we still refer to `context.instances` in the step impl function. This causes quadratic complexity because we have then the same loop in the decorator and a loop in the impl function.

See JIRA for the test model.

This was also a nice opportunity to do some clean up.

I found out that all `values must be unique|identical` were already using the `... at depth 1` logic except for this one. So that was an opportunity to actually remove all other patterns and really require a depth specification. With the current decorator it also doesn't make sense to think about uniqueness/identical in the context of a singular element which would be how the impl function would be called without a depth specification. 

I think for the `within_model` hack we already concluded at some point that we can get rid of it now that we have the `... at depth 1` which is kind of a generalization of that. But I didn't fully remove it yet, only the reference from this impl function.

### DIFF
--- a/features/steps/validation_handling.py
+++ b/features/steps/validation_handling.py
@@ -1,5 +1,6 @@
 import functools
 import json
+import re
 from utils import misc
 from functools import wraps
 import ifcopenshell
@@ -148,9 +149,9 @@ def handle_given(context, fn, **kwargs):
             pass # (1) -> context.applicable is set within the function ; replace this with a simple True/False and set applicability here?
     else:
         context._push('attribute') # for attribute stacking
-        if 'at depth 1' in context.step.name:
-            #todo @gh develop a more standardize approach
-            context.instances = list(filter(None, map_given_state(context.instances, fn, context, depth=1, **kwargs)))
+        depth = next(map(int, re.findall('at depth (\d+)$', context.step.name)), 0)
+        if depth:
+            context.instances = list(filter(None, map_given_state(context.instances, fn, context, depth=depth, **kwargs)))
         else:
             context.instances = map_given_state(context.instances, fn, context, **kwargs)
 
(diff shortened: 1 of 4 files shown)


aothms:
Oh fixing GRF proved to be a bit tricky, requiring to set `... at depth 2`, because `HasCoordinateOperation` already returns an aggregate, so we have

2| All instances of ...
1| Attribute HasCoordinateOperation
0| Value dict

And we want to compare these across all instances in the model, hence depth 2.

So that's why we were really cheating earlier, kind of circumventing the depth 1 we had previously being bypassed by the within_model.

So it's a bit cleaned up now, but there is a somewhat interesting complication still in the recursive should_apply() function.

Previously we had an `all()` predicate there, i.e `is_nested(values) and all(not is_nested(v) for v in values)` and this worked probably, but at depth level 2 we can have inconsistent depths, because of None values. So I changed the predicate there from all to any. It's not necessarily wrong, but it's a change to be aware of at least.

evandroAlfieri:
Good ones! Version bump for both, due to behaviour change?

aothms:
> due to behaviour change?

I think we can get away with not doing that, because the behaviour should remain consistent, just more performant :)

aothms:
Edit: No sorry, still not correct..

The predicate should be all() but the Nones should be filtered?

features/steps/validation_handling.py
```diff
@@ -148,9 +149,9 @@ def handle_given(context, fn, **kwargs):
             pass # (1) -> context.applicable is set within the function ; replace this with a simple True/False and set applicability here?
     else:
         context._push('attribute') # for attribute stacking
-        if 'at depth 1' in context.step.name:
-            #todo @gh develop a more standardize approach
-            context.instances = list(filter(None, map_given_state(context.instances, fn, context, depth=1, **kwargs)))
+        depth = next(map(int, re.findall('at depth (\d+)$', context.step.name)), 0)
+        if depth:
```
Ghesselink, line 153: Thanks. :-)

Ghesselink (approved):
> So that's why we were really cheating earlier, kind of circumventing the depth 1 we had previously being bypassed by the within_model.

Where we cheating or was GRF001 simply developed after the decorator work? The rule was developed a while ago, so I suspect we were indeed cheating .. 
Anyway, this works really nice. Thanks. 

This is also good to take in mind with the implementation cleanup, there will probably also be some adaptations in the decorator. 
Like mentioned in Slack, the file I've tested it with runs almost instantly with these implementation (as opposed to 15min+) :) 
I took a note to do some further logging and quantify these statements (i.e. regarding the improvement of performance)

PR #338, opened 2025-01-15
IVS-322 Relationship from ... to ... performance

aothms:
This performance issue is a bit more subtle, but often there are `1 to many` relationships where many is really many such as spatial containment.

I already adapted this impl function once to use inverses, but ifcopenshell is also kind enough to give you the attribute index for which an inverse is realized.

`{inst} & to_entity` is simply not so fast if it's 10000 elements in to_entity (mostly because of dealing with large aggregates in python-ifcopenshell is slow per definition).

But if we use the attribute index (and convert that to a name which we cross check with the CSV) we already know that inst is part of to_entity.

### DIFF
--- a/features/steps/givens/relationships.py
+++ b/features/steps/givens/relationships.py
@@ -32,9 +32,9 @@ def step_impl(context, inst, relationship, dir1, entity, dir2, other_entity, tai
     related_attr_matrix = system.get_csv(filename_related_attr_matrix, return_type='dict')[0]
     relating_attr_matrix = system.get_csv(filename_relating_attr_matrix, return_type='dict')[0]
 
-    relationships = [i for i in context.model.get_inverse(inst) if i.is_a(relationship)]
+    relationships = [i for i in context.model.get_inverse(inst, with_attribute_indices=True, allow_duplicate=True) if i[0].is_a(relationship)]
 
-    for rel in relationships:
+    for rel, attribute_index in relationships:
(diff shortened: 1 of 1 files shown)


Ghesselink:
I've done an experiment with SPS007, ran a large file and logged the execution time of the step 
```
'a *required* relationship IfcRelContainedInSpatialStructure to IfcElement from IfcSpatialElement'`
```
This step was executed 16 times in total, with an average execution time of 
```
sum(context.execution_time) / len(context.execution_time)
0.0030741095542907715
```

When I replace the code with the old one, this execution time is significantly more
```
sum(context.execution_time) / len(context.execution_time)
0.5054181516170502
```

In other words, in this example it's 164 times faster. Nice! :-)

Ghesselink:
I did the same experiment for both statements (there is another one in another scenario)
`a *required* relationship IfcRelContainedInSpatialStructure to IfcElement from IfcSpatialElement'`

There were 52900 instances processed in the file. In total, it took 50 seconds. The average processing time per instances was less than 1/1000s ..
```
sum(exc_times) / len(exc_times)
0.0009486370140753532
sum(exc_times)
50.18289804458618
len(exc_times)
52900
```

I tried running the same file with the current implementation. After three hours it still didn't process locally and the batteries of my laptop were almost drained, that's probably enough to prove the point.

It would probably be helpful to add this executing time logging per statement to the validation platform, I've added a Jira ticket under the Performance epic for this.

Ghesselink (approved):

PR #340, opened 2025-01-23
IVS-168 Re-enable rules

aothms:


civilx64 (commented):
Previously we had disabled BRP001 and TAS001 for performance reasons.  Have we tested to confirm that those issues are resolved?

aothms:
> Previously we had disabled BRP001 and TAS001 for performance reasons. Have we tested to confirm that those issues are resolved?

My mind was a bit fuzzy on that. I thought it was because of an ifopsh bug, but maybe that was only the sweeps which are based on composite curves and have per segment orientation (I think it was related to that).

Ok, so what I will do is rerun these to get some definitive number of a random sample of models.

civilx64:
> My mind was a bit fuzzy on that. I thought it was because of an ifopsh bug, but maybe that was only the sweeps which are based on composite curves and have per segment orientation (I think it was related to that).

SWE001 had an issue with edge orientation (ifcopsh bug) per the notes captured in [IVS-142](https://buildingsmart.atlassian.net/browse/IVS-142).

BRP001 and TAS001 were disabled because of performance reasons in [hotfix 0.6.7.1](https://forums.buildingsmart.org/t/validation-service-updates/4992/13).
> 
> Ok, so what I will do is rerun these to get some definitive number of a random sample of models.

Sounds good.  What I would like you to do is revise this PR to enable SWE001 only, then create a new JIRA ticket under the performance epic for BRP001 and TAS001.

civilx64:
Schema errors that do need to be addressed since these are "passing" models.
image

image

civilx64:
Please address the few lingering schema issues and this should be good for UAT.  Thanks!

civilx64 (commented):
Please address the few lingering schema issues and this should be good for UAT. Thanks!

Issue #342, opened 2025-01-28
GEM003 - Unique Representation Identifier error in files without product representation


fg55b:
Greetings!

I am encountering the GEM003 - Unique Representation Identifier error in files that do not include any product representation. This issue first appeared in release v0.6.10, where both the GEM003 _rule_ and the _then_ step used on it were modified.

The error message is as follows. It does not specify the ID or entity where the issue occurs. The error can be reproduced by omitting the Representation attribute of an IfcProduct.

Thank you for your attention, and let me know if additional details are required!

![Image](https://github.com/user-attachments/assets/45cbb44e-756f-4588-a076-08c208a440e9)

aothms:
Many thanks for reporting, this is a serious issue, but luckily not very frequently occurring: only when all products lack a representation; in which case we consider the wrong depth to apply the uniqueness constraint to. We'll fix this asap.

PR #343, opened 2025-01-28
IVS-368 reverse depth

aothms:
#342 reverse the direction in which depth is applied

Previously, depth was applied 'upwards': normal rule execution is applied to the leafs of the tree and depth was a means to add several steps upwards, i.e to recurse less deep into the nested structure of values representing our tree.

The problem with this is that it is inconsistent. Because trees can be less deep than the scenario 'expects'. In #342 the problem was that when no product has a representation, the upwards-applied depth=1 causes the rule to be applied not (as expected) to the set of representation identifiers within a product but (instead) a bunch of None values (because all product.Representation=None, and hence subsequent depth potential was not realised)

The work-around in this PR is to calculate depth from the root downwards. This essentially means that every Given step that translates to an aggregate <sup>1</sup> increases the maximum depth.

<sup>1</sup> (which is consistent in the schema, would be good to double check things like monstrous constructs like IfcPropertySetDefinitionSet - which is used selects between a single value and an aggregate, but because the aggregate is still wrapped in a defined type, it's essentially still singular values on both sides of the select)


Ghesselink (approved):
> The work-around in this PR is to calculate depth from the root downwards. This essentially means that every Given step that translates to an aggregate 1 increases the maximum depth.

Makes sense :-) 
I've deployed locally and it seems to work well. The relevant rule (GEM003) is not activated as expected. Furthermore, [I updated the headers of all the GEM003 files](https://github.com/buildingSMART/ifc-gherkin-rules/pull/343/commits/129f309ddbf73a1fc534ccdd333b3efa63ecf2d8). 
image
image

PR #348, opened 2025-01-31
IVS-307 performance

aothms:
**ctx000**

Takes about a minute now on the reference model.

The biggest change here is that doing traversal on the relationship directly also enters the elements that the material is assigned to and elements have very deep instance trees due to their representations.

Also, HasAssociations is defined on IfcObjectDefinition

**pse001**

Main problem here is that for every instance we parse the full data structure again to see what is allowed (because previously the loop was inside the step fn). Nothing that some quick caching cannot solve, but it's a bit ugly because of the dictionary that has to be passed. Could be refactored more elegantly probably, but this is the least intrusive I could think of.

**alb023**

The reference file had 24197 IfcCurveSegments. Looping over them in the decorator and then finding their relevant IfcAlignmentSegment by iterating over the context in the impl function creates exponential behaviour. We have something for this called `full_stack_rull` which not only gives you the instance but a path through the values in all given steps. Implementation of that was broken though.. it assumed that the depth in context.instances increased with 1 every step, rather we need to look at the depth of context.instances (or the stackframe we're inspecting actually) that we have for the frame and then apply as many indexes to that from the numeric path based on that depth. Adding @civilx64 as a reviewer due to this rule.

### DIFF
--- a/features/CTX000_Presentation-colours-and-textures.feature
+++ b/features/CTX000_Presentation-colours-and-textures.feature
@@ -1,6 +1,6 @@
 @implementer-agreement
 @CTX
-@version1
+@version2
 @E00020
 
 Feature: CTX000 - Presentation colours and textures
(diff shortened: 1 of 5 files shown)


Ghesselink:
Followed up by https://github.com/buildingSMART/ifc-gherkin-rules/pull/348

Issue #356, opened 2025-02-14
IFC105 possible false negative


RickBrice:
I'm getting a violation of IFC105, but I don't understand why. The coding seems correct per [4.1.2.6.5 Material Single](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Association/Material_Association/Material_Single/content.html) and [4.1.6.2.2 Property Sets for Materials](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Definition/Property_Sets/Property_Sets_for_Materials/content.html)

![Image](https://github.com/user-attachments/assets/5878242c-ecaa-4a04-b2cb-5537400c0c74)

Here is the IFC snippet

~~~
#859=IFCPROPERTYSINGLEVALUE('CompressiveStrength',$,IFCPRESSUREMEASURE(5.9),#858);
#860=IFCDIMENSIONALEXPONENTS(1,0,0,0,0,0,0);
#861=IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.);
#862=IFCMEASUREWITHUNIT(IFCLENGTHMEASURE(0.0254),#861);
#863=IFCCONVERSIONBASEDUNIT(#860,.LENGTHUNIT.,'inch',#862);
#864=IFCPROPERTYSINGLEVALUE('MaxAggregateSize',$,IFCPOSITIVELENGTHMEASURE(0.75),#863);
#865=IFCMATERIAL('Precast Concrete, f''c = 5.900 KSI',$,'concrete');
#866=IFCMATERIALPROPERTIES('Pset_MaterialConcrete',$,(#859,#864),#865);
#867=IFCSHAPEREPRESENTATION(#102,'Body','AdvancedSweptSolid',(#845));
#868=IFCPRODUCTDEFINITIONSHAPE($,$,(#867));
#869=IFCBEAM('175Q29X0939esviFnGy5yJ',$,'Segment 1',$,$,#854,#868,$,$);
#870=IFCRELASSOCIATESMATERIAL('3e2K92RorF8xwVZ_4OQbaV',$,'Associates_Concrete_to_Precast_Segment',$,(#869),#865);
~~~

#866 references #865. #865 is referenced by #870 which is a rooted entity.

[...]

PR #357, opened 2025-02-14
IVS-397 IFC105 material properties

aothms:


Ghesselink (approved):
See https://github.com/buildingSMART/ifc-gherkin-rules/issues/356#issuecomment-2660954282

image

PR #359, opened 2025-02-25
IVS-221 -  PSE002

Ghesselink:
Includes testfiles from PSE001 with the following pset names 

- pset ActuatorTypeElectricActuator
- Pset WallCommon
- psET_WallCommon
- Pset&WallCommon
- Pset*WallCommon
- Pset_WallCommon -> Is a N/A testfile here because it doesn't activate PSE002 (it's a PSE001-testfile)
- Pset-ActuatorTypeElectricActuator
- PSET_WallCommon
- Pset WallCommon
- Pset&MyWall

E.g. for the last example (`Pset&MyWall`), there is a function to [normalize](https://github.com/buildingSMART/ifc-gherkin-rules/blob/889b41c3f8074244a09267753de7efc18350eb1d/features/steps/thens/relations.py#L209) the name back to `Pset_MyWal`l. Because `Pset_MyWall` is not one of the predefined values an IBP-error is raised. Without this normalization, the wrong/custom prefixes would often lead to false positives.

> Is a N/A testfile here because it doesn't activate PSE002 (it's a PSE001-testfile)

The reverse is also true, one of these examples (outside Pset_Wallcommon) is a NA for PSE001. See also 
[test/files/pse001/na-pse001-no_exact_pset_name_match.ifc](url)


features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
```diff
@@ -0,0 +1,38 @@
+@industry-practice
+@PSE
+@version1
+@E00020
+Feature: PSE002 - Custom properties and property sets validation
+The rule verifies that each custom property set is correctly defined.
```
civilx64, line 6:
```suggestion
The rule verifies that each custom property set is named in accordance with best practices.
```

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
```diff
@@ -0,0 +1,38 @@
...
+Standard property sets must begin with 'Pset_' (i.e. PSE001)
```
civilx64, line 7:
```suggestion
Standard property sets must begin with 'Pset_' (enforced by PSE001)
```

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
```diff
@@ -0,0 +1,38 @@
...
+while this rule specifically checks for generic or custom cases where 'Pset' is followed by any separator (e.g., _, -, ., or a blank space).
```
civilx64, line 8:
```suggestion
This rule specifically raises a warning for property set names that begin with 'Pset' (case insensitive) followed by any separator (e.g., _, -, ., or a blank space).
```

civilx64 (commented):
I've made some suggestions to improve the wording of the feature file.  Otherwise looks good!

Please double-check me on the submodule update.

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
```diff
@@ -0,0 +1,38 @@
...
+  
+  Background:
+   Given An .IfcPropertySet.
+   Given its Name attribute starts with r'^(?!Pset_)[Pp][Ss][Ee][Tt]\w*'
```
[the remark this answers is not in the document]
aothms: Lastly [sorry for the long chain of comments, I guess that's my brain slowly waking up to things]

I think the semantics of this rule are better rephrased differently:

```suggestion
   Given its attribute .Name.
   Given the value does not start with 'Pset_'
   
    Then  the value must not conform to the expression /^[Pp][Ss][Ee][Tt]/
    
[end of feature]
```

In other words, the rule is simply: people are not recommended to have their custom psets start with pset. It does not need to be checked with anything property definitions in the spec.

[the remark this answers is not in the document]
aothms: Final lastly.

I think you can remove the `\w*` because we only care about the partial match of the beginning.

aothms (changes requested):

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
aothms: That's a bit circular, I would say:

> The rule verifies that property set names do not start with any upper- and lowercase variation of 'pset' in order to prevent confusion with standardized property sets in the IFC specification.

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
aothms: I would not use 'must' because this is a fact, not a requirement

> The property sets defined in the IFC specification all start with 'Pset_'

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
aothms: I don't think a 3rd sentence is necessary if we stick to a more descriptive 1st sentence. 

Also note, theoretically nothing need to come after the [Pp][Ss][Ee][Tt] as * matches 0 or more characters. And it's also not necessarily a 'separator' character. `'Psettlement'` is also not allowed ;)

civilx64: Thanks for providing a second set of eyes. I am good with your recommendations and defer to @evandroAlfieri as the tie breaking vote if necessary.

Ghesselink:
> Did we agree on a plan to consistently agree on:or was that not part of the current refactoring effort.

We considered it, but for PSE001 it's not very handy as we'll have the value for the attribute as the `inst` rather than the entity itself in the subsequent steps. E.g. when we refactor it like this, instead of having the '`IfcPropertySet`' as inst in the next step, we' would have the string '`Pset_MyWall`'.
We have even documented this, to my surprise https://github.com/buildingSMART/ifc-gherkin-rules/blob/048edc36a86b0321522c281ae70bbfd1a2864275/features/steps/givens/attributes.py#L52-L74
Maybe there is a workaround for this. For example, by going back into the stack tree somehow when for a particular step implementation we encounter a wrong type (a `str` in this case rather than an `ifcopenshell.entity_instance`)

However, in this (PSE002) case it's perfectly possible. I've [refactored](https://github.com/buildingSMART/ifc-gherkin-rules/pull/359/commits/27bba8a30dde893a225e671613d529af64297113) some parts of the code so we have all the options available. 
```
    Scenario: Raise a warning for unstandardized property set names

      Given An .IfcPropertySet.
      Given Its attribute .Name.
      Given Its value does not start with Pset_

      Then Its value must not conform to the expression /^[Pp][Ss][Ee][Tt]/
```

The[ testfiles are also named clearer now](https://github.com/buildingSMART/ifc-gherkin-rules/pull/359/commits/048edc36a86b0321522c281ae70bbfd1a2864275) and there is a '`Psettlement`' example :-)

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
Ghesselink: Sounds good :-) 

Current description ; 
```
The rule verifies that property set names do not start with any upper- and lowercase variation of 'pset' in order to prevent confusion with standardized property sets in the IFC specification.
The property sets defined in the IFC specification all start with 'Pset_'
```

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
Ghesselink: implemented

features/rules/PSE/PSE002_Custom-properties-and-property-sets-validation.feature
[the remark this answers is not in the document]
Ghesselink: [Implemented](https://github.com/buildingSMART/ifc-gherkin-rules/pull/359#issuecomment-2707773396)

aothms:
> We considered it, but for PSE001 it's not very handy as we'll have the value for the attribute as the inst rather than the entity itself in the subsequent steps. E.g. when we refactor it like this, instead of having the 'IfcPropertySet' as inst in the next step, we' would have the string 'Pset_MyWall'.

Of course. Thanks for reminding me!

> For example, by going back into the stack tree somehow when for a particular step implementation we encounter a wrong type (a str in this case rather than an ifcopenshell.entity_instance)

That's a bit magical and only works if we go from instance -> str in the given. The same could also apply to attributes that give us instance -> instance.

The other option I guess is naming, that would also allow comparing values perhaps and fits in nicely with the refactoring because we could come up with a new syntax structure for previously stored variable names. But it will get very hard figuring out the nesting, what if these are at different depths? Let's leave it this for now and revisit this when we need it.

      Given An .IfcPropertySet.
      Given Its attribute .Name. =PropertySetName
      Given Its attribute .Description. =PropertySetDescription
      Then {PropertySetName} should not be equal to {PropertySetDescription}

aothms (approved):

PR #366, opened 2025-03-03
Improve blt002 feature description

janbrouwer:
BLT002 Feature description is missing the desciption for the second scenario.

Might even be a good idea to show the scenario titles to the validation service users.


evandroAlfieri:
@civilx64 @Ghesselink might be a good chance to update BLT003 in a similar fashion

Ghesselink:
Opened a new PR :) 
This one was trying to push to main, and got a bit in a puzzle trying to change the base to `development`.
Also changed BLT001 (the same rule, but with Doors) 

> Might even be a good idea to show the scenario titles to the validation service users.

Agree. This could probably be done by storing an integer (1 for the first scenario, 2 for the second, etc) along with the error messages. The complicated part is that the order and names of scenario change over time, so we must keep a good versioning strategy.

Ghesselink:
Closed due to https://github.com/buildingSMART/ifc-gherkin-rules/pull/367

PR #368, opened 2025-03-04
IVS-430 Fix ifc105 again and again

aothms:


Ghesselink:
Should we also add this exception to the feature description? 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/tfk-fix-ifc105-again-and-again/features/rules/IFC/IFC105_Resource-entities-need-to-be-referenced-by-rooted-entity.feature

`The inverse attributes that are followed are: StyledByItem HasCoordinateOperation LayerAssignments LayerAssignment HasSubContexts HasProperties (material and profile def) HasRepresentation (material)
`
to 

`The inverse attributes that are followed are: StyledByItem HasCoordinateOperation LayerAssignments LayerAssignment HasSubContexts HasProperties (material and profile def) HasRepresentation (material) HasColours (indexed colour map) HasTextures (indexed texture map)
`

Issue #363, opened 2025-03-01
Error in ALS005 for alignment with only horizontal


RickBrice:
Per concept template [4.1.7.1.1 Alignment Geometry - Horizontal](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Geometric_Representation/Alignment_Geometry/Alignment_Geometry_-_Horizontal/content.html) the alignment representation should be

IfcShapeRepresentation.
   RepresentationIdentifier=Axis
   RepresentationType=Curve2D

For the attached file, a normative rule error is occurring stating that Curve3D is expected.

![Image](https://github.com/user-attachments/assets/db5a80ab-edd0-4058-ab6a-7a47e541b8e4)

[Alignment.ifc.txt](https://github.com/user-attachments/files/19040674/Alignment.ifc.txt)

RickBrice:
Fixed with 0.7.1 release

PR #371, opened 2025-03-08
fix ALS005 for alignments with horizontal only (IVS-432)

civilx64:
addresses #363


civilx64:
Diff view is possibly a bit confusing.  I added a new scenario which meant that the scenario numbers for many of the test files needed to be incremented by 1.

Ghesselink:
I found some Schema errors for the following testfiles 
image
image

The second one has to do with the header (can easily be fixed by taking the header of the passing files and modifying the filename), but I'm not sure regarding the first error message
image

features/rules/ALS/ALS005_Alignment-shape-representation.feature
```diff
@@ -14,24 +14,36 @@ The rule verifies that each IfcAlignment uses correct representation.
...
+    Then .RepresentationIdentifier. ^is^ 'FootPrint' or 'Axis'
 
 
   @E00010
-  Scenario: Agreement on each IfcAlignment using correct representation - Type = 'Axis'
+  Scenario: Agreement on each IfcAlignment using correct representation - horizontal only
 
     Given .RepresentationIdentifier. ^is^ 'Axis'
-    Then The value of attribute RepresentationType must be Curve3D
+    Given .RepresentationType. ^is^ 'Curve2D'
```
Ghesselink, line 24: This would probably also be nice to do for other files (e.g. ALS006 & ALS007), so we can get rid of another step implementation.

PR #375, opened 2025-03-13
ALB010 regression (IVS-454)

civilx64:
Warnings no longer being raised

Issue #66, opened 2023-05-02
Remove "IFC4x1" from the Given


evandroAlfieri:
https://github.com/buildingSMART/ifc-gherkin-rules/blob/47b6a14abb41da91b97d5ca65dc3c1f1ab243429/features/GRF001_Identical-coordinate-operations.feature#L8

Ghesselink:
https://github.com/buildingSMART/ifc-gherkin-rules/pull/70

JTurbakiewicz:
@evandroAlfieri  @Ghesselink,

@Ghesselink has removed IFC4x1 from the rule with https://github.com/buildingSMART/ifc-gherkin-rules/commit/1b2f0ae45e9fcd1ab98f0e53d12217ad903bbf7b

I've checked and IFC4x1 is not present anywhere else in repo. I'm closing the issue.

PR #376, opened 2025-03-21
IVS-401 - Add SPS007 exception for hosted annotations

Ghesselink:
- Create new scenario (2) with two testfiles; one passing and one failing 
- Modify relationships step_impl to check for absence of relationships (in this case, we have to consider IfcAnnotation only when it is not a hosted element) 
- Register custom type `exists_or_not_exist` which affects the `required` variable
-  Revert https://github.com/buildingSMART/ifc-gherkin-rules/pull/375 ;not required anymore when tail is a boolean value)
- Simplify the yield of Outcome; the step_impl is effectively always yielding an outcome, the relevance is determined by the type of the step; Given or Then). 
- Got a deprecation warning for the escape character `'/d'`, fixed it by changing it to a raw string


civilx64 (commented):
Code looks good from the surface.  Sample file no longer triggers an error for SPS007.

You might consider cleaning up the unit test files to make certain they all pass the schema checks.

features/rules/SPS/SPS007_Spatial-containment.feature
```diff
@@ -1,20 +1,24 @@
...
 
-        Given an .<entity>.
-        Then a *required* relationship IfcRelContainedInSpatialStructure to IfcElement from IfcSpatialElement
+    Scenario: Instances of IfcGrid must be contained within a spatial structure
+
+        Given an .IfcGrid.
+        Then a relationship IfcRelContainedInSpatialStructure must exist to IfcGrid from IfcSpatialElement
+
+
+    Scenario: Instances of IfcAnnotation must be contained within a spatial structure, except when the annotation is already a hosted eleent
```
aothms, line 16: > eleent

element.

I would also prefer more precise language, e.g

> ... when the annotation nests another annotation

Which is still really hard to read for non-ifc people

> ... when the annotation is a nested one under another annotation

Maybe rule master annex native speaker has some ideas @civilx64

test/files/SPS/sps007/pass-sps007-scenario02-annotation_hosted_element.ifc
```diff
@@ -0,0 +1,144 @@
...
+#18=IFCDIRECTION((0.,1.));
+#19=IFCPROJECTEDCRS('n7rJmd','7yPIU9FJik','EPSG:7844','EPSG:5711','UTM','56',$);
+#20=IFCRIGIDOPERATION(#12,#19,0.,0.,0.);
+#21=IFCRELDECLARES('1wYK6WxCrEVuvIYx98krR$',#2,'vofFGYo','nDwErbk4bU',#1,(#22));
+#22=IFCACTOR('2mtFEuUsjAZvsYlSxLubX3',$,'XA5xdjh','m4g56gsA7B',$,#4);
+#23=IFCSITE('2G0ATHGhj5T8pYNNtAHW7K',#2,'qfpeGrpS','oF7Hjskd4',$,#13,$,$,.ELEMENT.,$,$,$,$,$);
+#24=IFCRELAGGREGATES('23F9xPv0L64BRir8Qemk5E',#2,'cq5acgJ','HRnXa7z',#1,(#23));
+#25=IFCFACILITY('1ImCI2aTD1DQ6ZcsL4wZU1',#2,'JlsLnx4','7oXx5GP7',$,#13,$,$,.ELEMENT.);
+#26=IFCRELAGGREGATES('2AnbcmI2r3HwhOaiFhHKMQ',#2,'JXZia1evi','VmSXvwm',#23,(#25));
+#27=IFCRELCONTAINEDINSPATIALSTRUCTURE('0y1pwwXIn9pwEP14uHse4P',#2,'Pa5r3Notr','vinelT',(#28),#25);
```
aothms, line 34: So this is correct, but I think it would be good to have the 'reverse' as another fail file. Where we contain the 'nested' rather than the 'nestee'.

#27=IFCRELCONTAINEDINSPATIALSTRUCTURE('0y1pwwXIn9pwEP14uHse4P',#2,'Pa5r3Notr','vinelT',(#30,#43,#56,#66),#25);

Because:

> Given A relationship IfcRelNests does not exist to IfcAnnotation from IfcAnnotation

is not very clear of  course in terms of the directionality

aothms:
Sorry for the two late drive by comments. Nice work.

features/rules/SPS/SPS007_Spatial-containment.feature
civilx64: No new ideas. The second option you proposed is better.

PR #377, opened 2025-03-21
IVS-401 - Further SPS007 improvements

Ghesselink:
Continuation of [IVS-401 ](https://github.com/buildingSMART/ifc-gherkin-rules/pull/376)

- Create testfile containing the 'nested' instead of 'nestee', resulting in a fail. 

>  Given A relationship IfcRelNests does not exist to IfcAnnotation from IfcAnnotation

- Improve the scenario description. See also https://github.com/buildingSMART/ifc-gherkin-rules/pull/376#discussion_r2007052361 do you have also idea's @civilx64 

Furthermore, I was reading the mentioned text in the CT 
> The hosted component should not be contained in the spatial hierarchy, i.e. the concept Spatial Containment shall not be used at the level of hosted components. The hosted component is contained in the spatial structure by the spatial containment of its host.

We're now only considering annotations. Are there more exceptions that we should explicitly handle, or can we consider a more general rule (e.g. exclude all elements that are hosted in a nesting relationship from the scenario check)? And should we also check the reverse explicitly (i.e. the hosted component should not be contained in the spatial hierarchy) rather than only exclude non-hosted components from the spatial hierarchy check? @aothms 
Then we could write something like 
```
Scenario: 
Given a relationship IfcRelNests to IfcProduct from IfcProduct # only IfcRelNests or also other relationships?
Then a relationship IfcRelContainedInSpatialStructure must not exist to IfcProduct from IfcSpatialElement
```
& 
```
Scenario:
Given a relationship IfcRelNests does not exist to IfcProduct from IfcProduct # only IfcRelNests or also other relationships?
Then a relationship IfcRelContainedInSpatialStructure must exist to IfcProduct from IfcSpatialElement
```


aothms:
> We're now only considering annotations

I don't know. You would think so, but nesting is only used in very specific cases: alignment (because the segments are ordered), ports (because of ... reasons?), sequencing tasks (tasks are ordered), and apparently annotations (because, I mean why not).

I'd be tempted to say we don't need to make a general rule out of it. But I get the uneasiness I think you feel. It all seems really arbitrary and it is.

test/files/GEM/gem002/fail-gem002-spaces_4.ifc
aothms, line 1: Is this a mistake?

aothms (approved):

civilx64:
> I'd be tempted to say we don't need to make a general rule out of it. But I get the uneasiness I think you feel. It all seems really arbitrary and it is.

I agree; let's keep it scoped just to IfcAnnotation for now and move on.

civilx64 (commented):
LGTM; just need to resolve the comment re: GEM002.

test/files/GEM/gem002/fail-gem002-spaces_4.ifc
Ghesselink: Yes, sorry. Think it was a local file for testing Jan Brouwer's files for GEM002. Removed it 
https://github.com/buildingSMART/ifc-gherkin-rules/pull/377/commits/51df9f365b11acffd98f21739a25fd7bacaea3be

PR #379, opened 2025-03-24
IVS-301 - OJP000 - Object Placement

Ghesselink:
Very simple rule, verifying the existence of at least one product with a local placement. 

> We need to add this activation rule because we’re getting strange cases where geometry exists and object placement “doesn’t”, which is against the standard (PlacementForShapeRepresentation WR). It’s not that it doesn’t have placement but simply we have only one rule on OBJ that when not triggered results in a grey card.


features/rules/OJP/OJP000_Object-placement.feature
```diff
@@ -0,0 +1,15 @@
...
+Feature: OJP000 - Object placement
+The rule verifies the presence of an object placement to define the overall context and directory of objects within the model. 
+Among others, the context definition includes default units and geometric representation context for shape representations.
+
+  @E00010
+  Scenario: Check for activation . 
+
+      Given An .IfcObject.
+      Given Its attribute .ObjectPlacement.
+      Given [Its entity type] ^is^ 'IfcLocalPlacement'
```
aothms, line 13:
```suggestion
```

I think this should be removed. OJP000 - Object placement should only check for the existence of a value for the ObjectPlacement attribute, not the type.

Otherwise it should be LOP000 - Local placement or something.

features/rules/OJP/OJP000_Object-placement.feature
```diff
@@ -0,0 +1,15 @@
+@implementer-agreement
+@OJP
+@version1
+Feature: OJP000 - Object placement
+The rule verifies the presence of an object placement to define the overall context and directory of objects within the model. 
+Among others, the context definition includes default units and geometric representation context for shape representations.
```
aothms, line 6:
```suggestion
The rule verifies the presence of an object placement to define the overall local coordinate system and placement of the product in space
```

aothms (changes requested):

features/rules/OJP/OJP000_Object-placement.feature
aothms: Cool, question is though if we still need OJP000 later on if we have rules from Grid, Linear and Local placement. But for now it's good to keep it I guess.

aothms (approved):

features/rules/OJP/OJP000_Object-placement.feature
```diff
@@ -0,0 +1,13 @@
+@implementer-agreement
+@OJP
+@version1
+Feature: OJP000 - Object placement
+The rule verifies the presence of an object placement to define the overall local coordinate system and placement of the product in space
```
evandroAlfieri, line 5: @Ghesselink please review the descriptions of all 000 rules. They must always, consistently start with "The rule verifies the presence of IFC entities used to..."
cc: @civilx64

evandroAlfieri (changes requested):

features/rules/OJP/OJP000_Object-placement.feature
Ghesselink: No, it's maybe a bit redundant .. But let's stick with having an activation rule for every functional part, to not add more exceptions to IFC, hehe :)

features/rules/OJP/OJP000_Object-placement.feature
civilx64, line 5:
```suggestion
The rule verifies the presence of IFC entities used to define an object placement, which in turns establishes the local coordinate system and spatial position of the product.
```

Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/pull/379/commits/23ea49b3e9c7aa7c281e9017e2422fbc5b392299

civilx64 (approved):

PR #380, opened 2025-03-24
IVS-331 - BLT000 - Built Elements

Ghesselink:
Many built elements already activate the BLT functional part, BLT000 activates BLT explicitly when any built element is present (or IfcBuldingElement for IFC4 and IFC2X3). 
This will prevent false negatives. For instance, a file with `IfcBeam` won't currently activate BLT but will when BLT000 is merged.


aothms (approved):
Nice!

PR #381, opened 2025-03-24
IVS-233 - BBX000 - Bounding Box

Ghesselink:


features/rules/BBX/BBX000_Bounding-box.feature
```diff
@@ -0,0 +1,35 @@
+@implementer-agreement
+@BLT
```
aothms, line 2:
```suggestion
@GEM
```

I suppose?

features/rules/BBX/BBX000_Bounding-box.feature
```diff
@@ -0,0 +1,35 @@
...
+@version1
+@E00020
+
+Feature: BBX000 - Bounding Box
+The rule verifies the presence of an orthogonal box, 
+oriented parallel to the axes of the object coordinate system in which it is defined and containing a geometry object, 
+which defines the spatial extent of the latter.
```
aothms, line 9:
```suggestion
oriented parallel to the axes of the object coordinate system which defines the spatial extent of the latter.
```

I find this grammatically clearer

features/rules/BBX/BBX000_Bounding-box.feature
```diff
@@ -0,0 +1,35 @@
...
+
+
+Scenario: Check for activation - Shape Representation Type
+
+    Given an .IfcProduct. 
+    Given Its attribute .Representation.
+    Given Its attribute .Representations.
+    Given .RepresentationType. ^is^ 'BoundingBox'
+
+    Then The IFC model contains information on the selected functional part
```
aothms, line 35:
```suggestion

Scenario: Check for activation - Shape Representation Identifier

    Given an .IfcProduct. 
    Given Its attribute .Representation.
    Given Its attribute .Representations.
    Given .RepresentationIdentifier. ^is^ 'Box'
    
    Then .RepresentationType. ^is^ 'BoundingBox'

Scenario: Check for activation - Shape Representation Type

    Given an .IfcProduct. 
    Given Its attribute .Representation.
    Given Its attribute .Representations.
    Given .RepresentationType. ^is^ 'BoundingBox'
    
    Then .RepresentationIdentifier. ^is^ 'Box'
```

1. Thinking out loud here, but we could also upgrade this to a BBX001 rule. Box as the identifier should in my view always imply BoundingBox and vice versa.

2. I would remove the first scenario. The 2nd and 3rd describe the only valid way to use an IfcBoundingBox. RepresentationType='BoundingBox' implies an IfcBoundingBox in the items: https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcShapeRepresentationTypes.htm

aothms (changes requested):

features/rules/BBX/BBX000_Bounding-box.feature
evandroAlfieri: BBX, we need that for scorecards

Ghesselink: Doesn't it have to be both BBX and GEM?
This is the case in BRP, e.g. https://github.com/buildingSMART/ifc-gherkin-rules/blob/development/features/rules/BRP/BRP003_Planar-faces-are-planar.feature

Furthermore, I noticed that BRP002 only has a `@GEM` tag, so we should probably change this too. 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/development/features/rules/BRP/BRP002_Single-component-in-connected-faceset.feature

features/rules/BBX/BBX000_Bounding-box.feature
Ghesselink: That makes sense. 
What if we split it up two BBX000 and BBX001. BBX001 being the scenario you outlined, BBX000 would be simply
```

Scenario: Check for activation - Shape Representation Type
    Given an .IfcProduct. 
    Given Its attribute .Representation.
    Given Its attribute .Representations.
    Given .RepresentationType. ^is^ 'BoundingBox'

Scenario: Check for activation - Shape Representation Type
    Given an .IfcProduct. 
    Given Its attribute .Representation.
    Given Its attribute .Representations.
    Given .RepresentationIdentifier. ^is^ 'Box'
```
I agree it's silly because BBX000 will never be triggered independently of BBX001. However, we took the decision to have a activation rule for every FP. This way, it's explicit, otherwise the absence of an 000-rule will be an exception that should be documented somewhere (at a place more likely overlooked than a feature file). The same will then apply for other FPs too. 

What do you think @evandroAlfieri ?

features/rules/BBX/BBX000_Bounding-box.feature
Ghesselink: IIRC we decided to only go for one tag per rule. 
So the current tag should be `BBX` and BRP002 should be `@BRP` only

features/rules/BBX/BBX000_Bounding-box.feature
aothms: Sounds good to me, but then I would do:

BBX000: 
- Given .RepresentationType. ^is^ 'BoundingBox'
- Then ...

BBX001:
- Scen1
    - Given .RepresentationType. ^is^ 'BoundingBox'
    - Then RepId Box
- Scen2
    - Given RepId Box
    - Then .RepresentationType. ^is^ 'BoundingBox'

And maybe:

BBX002: "Bounding boxes can only be used as top-level shape representation items to outline the geometric extent of the product", but that would be low prio and I think more or less covered by the ShapeRepresentationTypes function in the schema. It's just that sometimes people think they can use a BoundingBox as part of the physical shape of the product, but that doesn't really happen in exported models from software I think.

PR #382, opened 2025-03-24
IVS-463 Fix mpmath precision

aothms:
Precision was not correctly set which meant rounding errors were accumulated beyond the typical tolerance values on models with large coordinates


features/steps/thens/geometry.py
```diff
@@ -294,7 +294,7 @@ def step_impl(context, inst: ifcopenshell.entity_instance):
 @gherkin_ifc.step("the boundaries of the face must conform to the implicit plane fitted through the boundary points")
 def step_impl(context, inst: ifcopenshell.entity_instance):
     import mpmath as mp
```
Ghesselink, line 296: Should we import packages at the top of the file? Or is there a reason for not doing this? 

I remember a couple of months ago the DEV environment was not running correctly due to an import error. Because it was inside this step implementation, it was quite difficult to debug. 

---

Edit; just looked it up. Is it because when we import mpmath inside the step_impl, we set a new instance of `MPContext`?

Ghesselink (approved):
I had a local deployment running already, the mentioned file indeed passes the validation 
image

In contrast to the current DEV 
image

features/steps/thens/geometry.py
aothms: > Or is there a reason for not doing this?

I guess not.

It indeed certainly makes import errors less discoverable.

But somehow I have this habit to import  things where they are used so that (a) the purpose of the module is easily understood (b) other functionality works without having to install the module

It doesn't make the MPcontext somehow more local, because imports are actually not local. When you import in the function there is indeed a local name assigned to that imported module, but that's only the name. The actual module is cached in sys.modules so the second time you import it will give you that module with the updated context.

I think it's still a bad habit.

civilx64 (approved):
Please clean up the mpmath import location and we'll be good to merge.

features/steps/thens/geometry.py
Ghesselink: Well, to be honest, the reasons you mention make sense to me. It's easy to see where is `mp` is used for. 
Just applied Scott's suggestion to be consistent with PEP8
https://github.com/buildingSMART/ifc-gherkin-rules/pull/382/commits/fc6656a9283d1adafa03b575caf499c506187e96

> It doesn't make the MPcontext somehow more local, because imports are actually not local. When you import in the function there is indeed a local name assigned to that imported module, but that's only the name. The actual module is cached in sys.modules so the second time you import it will give you that module with the updated context.

These internal mechanics were not completely clear to me, thanks for the explanation :-)

Ghesselink:
> Just applied Scott's suggestion to be consistent with PEP8
https://github.com/buildingSMART/ifc-gherkin-rules/commit/fc6656a9283d1adafa03b575caf499c506187e96

Ready to merge

PR #387, opened 2025-03-27
Add shapely

aothms:
just something I observed if you use latest ifopsh, doesn't hurt to add it now.. pretty sure soon scott needs to update ifopsh again...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
         echo $CONDA/bin >> $GITHUB_PATH
     - name: Install dependencies
       run: |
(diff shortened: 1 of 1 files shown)


civilx64 (approved):

civilx64:
@aothms I checked and it appears we have this included already in `requirements.txt` for the platform in the validate repo. I'd appreciate a double-check from you to confirm.

Ghesselink:
Just read this, shapely is indeed in `validate`
https://github.com/buildingSMART/validate/blob/647ba07058cdc34e703778bd35eea2f851eceada/backend/requirements.txt#L43
@civilx64

PR #388, opened 2025-03-27
IVS-469 Fix swe001 apply unit factor

### LINKED ISSUE #385: SWE001 false negative again
I think I found another case where SWE001 falsely reports a self intersection

Image

[self_intersection.ifc.txt](https://github.com/user-attachments/files/19469539/self_intersection.ifc.txt)

If I just slightly round two values below E-11, the test passes. 

Image

aothms:
Fixes #385

Most of the comparisons between precision values happens between data read directly from the model (such as point.Coordinates for an IfcCartesianPoint instance) but if we use ifcopenshell.geom.create_shape() we either need to apply 'convert-back-units', True to the settings, or convert the precision value also to meters.

### DIFF
--- a/features/steps/utils/ifc.py
+++ b/features/steps/utils/ifc.py
@@ -1,3 +1,4 @@
+import typing
 from .misc import do_try
 import ifcopenshell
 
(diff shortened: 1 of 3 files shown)


evandroAlfieri:
version bump?

aothms:
> version bump?

D'oh, of course

Ghesselink:
The attached false negative is positive now on a local deployment 
image

The new precision is indeed a lot smaller (in m); 
```
entity_contexts
>>> {#56=IfcGeometricRepr...001,#52,$)}
list(entity_contexts)[0].Precision ** ifcopenshell.util.unit.calculate_unit_scale(next(iter(entity_contexts)).file)
>>> 0.9931160484209338

# vs (old)
list(entity_contexts)[0].Precision
>>> 0.001
```

Ghesselink (approved):

features/steps/utils/ifc.py
```diff
@@ -11,15 +12,30 @@ def is_valid(inst, representation_id, representation_type):
...
+    Returns:
+        float: Precision value
+    """
     precisions = []
     if not entity_contexts:
         return default_precision
+    if return_in_m:
+        unit_scale = ifcopenshell.util.unit.calculate_unit_scale(next(iter(entity_contexts)).file)
+    else:
+        unit_scale = 1.
```
civilx64, line 33: @aothms so that I am clear, unit scale in the context of these changes is purely related to order of magnitude within SI system (m vs mm vs cm etc) because `ifcopenshell.geom` is always meters - correct?

Based on other work with alignments `unit_scale` makes me think of imperial vs. metric conversions.  So perhaps we can make this portion of the code base explicit by using something like  `unit_magnitude_factor` versus `unit_scale`.  Also maybe `return_in_m` could be refactored to `return_in_default_magnitude`.

aothms: If I understand your question correctly, it's both:

In IFC you have:

- SI Unit -> factor by means of prefix + SI unit enum
- Conversion based (for imperial) -> factor by means explicit measure + any base unit

The factor we get here (recursively) folds any factor into the SI equivalent, so for conversion based there is a while loop (because conversion based can refer to another conv.based or si with prefix).

https://github.com/IfcOpenShell/IfcOpenShell/blob/42c49af4e32660e13cfe44db594584c95b13024e/src/ifcopenshell-python/ifcopenshell/util/unit.py#L705-L707

---

The reason is indeed that ifcopenshell.geom is always meters by default (based on settings). So a Precision of 1 inch (where 1 inch is defined as `Conversion(2.54, Si(centi, meter))` needs to be converted to 0.0254.

civilx64: Got it. In that case we can leave the code as-is; no refactoring needed.

Issue #115, opened 2023-11-20
SPS005 - Don't allow elements simultaneously participating as nesting and aggregating children roles


aothms:
https://github.com/buildingSMART/IFC4.3.x-development/issues/734

Only constrain this from child to parent. Because from parent to child we actually do have cases of simultaneous nesting and aggregation. https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Composition/Nesting/Alignment_Layouts/Alignment_Layout_-_Reusing_Horizontal_Layout/content.html

PR #395, opened 2025-04-06
IVS-339 - PJS000

Ghesselink:


aothms:
Are you sure we're not being too dogmatic about this? Are we growth hacking the change log https://en.wikipedia.org/wiki/Goodhart%27s_law ?

If this is the level of 000-rules we're at now I think we can pause them for a while and go back to rules that can fail.

Ghesselink:
>  I think we can pause them for a while and go back to rules that can fail.

I think so, the next one in line would be https://github.com/buildingSMART/ifc-gherkin-rules/issues/115

civilx64:
This one is a bit unusual in that we expect to see a project in nearly every model that is exchanged. But that's not the case for all of the activation rules, e.g. GDP000. 

As for priorities, we're merely executing against the plan and the catalog backlog. Anything categorized as 0 or 1 is in scope for the current sprint.

features/rules/PJS/PJS000_Project.feature
```diff
@@ -0,0 +1,12 @@
+@implementer-agreement
+@PJS
+@version1
+Feature: PJS000 - Project
+The rule verifies the presence of IFC entities used to define the overall context of a model.
```
civilx64, line 5:
```suggestion
The rule verifies the presence of IFC entities used to define the overall context and a directory of objects contained within.
https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Project_Context/content.html
```

features/rules/PJS/PJS000_Project.feature
```diff
@@ -0,0 +1,12 @@
...
+
+  @E00010
+  Scenario: Check for activation .
```
civilx64, line 8:
```suggestion
  Scenario: Check for activation. 
```

civilx64 (commented):
`pass` and `na` behave as expected.  Please revise the headers to be in compliance with the header policy.

Ghesselink:
Headers are now passing 
image

civilx64 (approved):

PR #402, opened 2025-04-29
ALB012 precision (IVS-389)

civilx64:
Relaxes the tolerance for comparing the result of a calculated expression to the value provided in an IFC model for radius of curvature on vertical alignment segments.


features/steps/thens/attributes.py
```diff
@@ -117,9 +118,13 @@ def step_impl(context, inst, value_or_comparison_op:str, attribute:str=None, exp
         except Exception as e:
             raise ValueError(f"Error evaluating expression: {e}")
 
-        # Compare the attribute with the expression value, considering the precision
-        entity_contexts = geometry.recurrently_get_entity_attr(context, inst, 'IfcRepresentation', 'ContextOfItems')
-        precision = geometry.get_precision_from_contexts(entity_contexts)
+        # Compare the attribute with the expression value, considering the specified tolerance or
+        # precision of the applicable geometric context
+        if comparison_tolerance:
+            precision = float(comparison_tolerance)
```
aothms, line 124:
```suggestion
            precision = comparison_tolerance
```

^ with the above the cast is not necessary on our side

features/rules/ALB/ALB012_Alignment-vertical-segment-radius-of-curvature.feature
```diff
@@ -1,6 +1,6 @@
 @implementer-agreement
 @ALB
-@version1
+@version2
 @E00020
 Feature: ALB012 - Alignment vertical segment radius of curvature
   The rule verifies the 'RadiusOfCurvature' design parameter for vertical alignment segments.
```
aothms, line 6:
```suggestion
  The rule verifies the 'RadiusOfCurvature' design parameter for vertical alignment segments. Where specified, an absolute tolerance value in the length unit of the model is used for the comparison.
```

If I understand correctly, the tolerance is file units, so this means that an alignment in millimeters has to be more precise then an alignment in meters. I don't think this is wrong per se. The reverse is also odd to have a tolerance in meters that would be converted to feet for example. But it wouldn't hurt to be explicit about it.

aothms (changes requested):

features/steps/thens/attributes.py
```diff
@@ -55,11 +55,12 @@ def accumulate_errors(i):
...
 
-@gherkin_ifc.step("The value of attribute {attribute} must be {value_or_comparison_op}")
-@gherkin_ifc.step("The value of attribute {attribute} must be {value_or_comparison_op} {display_entity:display_entity}")
-@gherkin_ifc.step("The value of attribute {attribute} must be {value_or_comparison_op} the expression: {expression}")
-@gherkin_ifc.step("The resulting value must be {value_or_comparison_op}")
-def step_impl(context, inst, value_or_comparison_op:str, attribute:str=None, expression:str=None, display_entity=" "):
+@gherkin_ifc.step("The value of attribute .{attribute}. must be '{value_or_comparison_op}'")
+@gherkin_ifc.step("The value of attribute .{attribute}. must be ^{value_or_comparison_op}^ [{display_entity:display_entity}]")
+@gherkin_ifc.step("The value of attribute .{attribute}. must be ^{value_or_comparison_op}^ the expression: [{expression}]")
+@gherkin_ifc.step("The value of attribute .{attribute}. must be ^{value_or_comparison_op}^ the expression: [{expression}] [within a tolerance of] {comparison_tolerance}")
```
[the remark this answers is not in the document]
civilx64: Agree.  Incorporated in 77aebdfbf682daa656a3ab1a1bea1520247cd9b1

aothms (approved):

PR #404, opened 2025-05-05
QTY001 (IVS-87)

civilx64:
Pretty simple rule, but adds more substance to the `QTY` functional part.


aothms:
I don't think this is the intention behind the docs. Rather, I think the intention of the docs there is to establish the equivalent of our PSE001 rule. I.e:

- Given Name equals one of the predefined names, e.g Qto_DoorBaseQuantities [0]
- Then in that case it needs to (a) have the BaseQuantities MoM (pse001/psets do not have that). Names need to be one of the specified names in the spec [0] and type.

[0] https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/Qto_DoorBaseQuantities.htm

For equivalence between PSE001 and QTO001 it would be nice establish exactly the same structure. Although I'm not sure whether we can enforce that the `Qto_` name prefix is reserved for definitions from the spec.

civilx64:
Point taken - but the rule as-is captures and enforces a documented agreement, correct?  So we can rename this to QTY010 or something that doesn't conflict with existing PSE numbers, then come back to build QTY001 similarly to PSE001 as you have suggested.

civilx64:
> Although I'm not sure whether we can enforce that the Qto_ name prefix is reserved for definitions from the spec.

Why not?

aothms:
> Why not?

Not that I personally would be against it, just because I don't recall it's ever written down like that and people might complain.

> Point taken - but the rule as-is captures and enforces a documented agreement

No I don't think so, because the BaseQuantities MoM should *only* be set under the condition that the name matches any of the predefined names in the spec. (bottom half of https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/annex-b3.html)

civilx64:
Per discussion today, the `Qto_` prefix will not be enforced as part of this rule. _Note: this would have been implemented as QTY002, not QTY001, for consistency with the PSE functional part._

Also as discussed, `.MethodOfMeasurement. == 'BaseQuantities'` is only applicable to the standard `Qto` property sets defined in the specifications.

aothms (approved):
Beautiful. Nothing to add. Also many thanks for splitting up the:

```
@gherkin_ifc.step("The IfcPropertySet Name attribute value must use predefined values according to the '{table}' table")
@gherkin_ifc.step("The IfcPropertySet must be assigned according to the property set definitions table '{table}'")
@gherkin_ifc.step("Each associated IfcProperty must be named according to the property set definitions table '{table}'")
@gherkin_ifc.step("Each associated IfcProperty must be of type according to the property set definitions table '{table}'")
@gherkin_ifc.step("Each associated IfcProperty value must be of data type according to the property set definitions table '{table}'")
```

There's still some undocumented fragments like normalize_pset() as to why they are there that I find a bit hard to place. But they are here now and I don't think they affect anything considering the given statements we have in place. The fact that they only manipulate pset_ prefixes and not qto_ prefixes seems to indicate it can/should go away though.

Ghesselink:
> There's still some undocumented fragments like normalize_pset() 

@aothms 
`Normalize_pset` is used for to cover 'cheats' like `Pset Wallcommon` or `pset_wallcommon` instead of the correct `Pset_Wallcommon`. It'll still pass for PSE001 now, but raises a warning for PSE002. 

The tests also passing now :-) One `Given` step from OJT001 was not adjusted yet.

There are still some merge conflicts in the datamodel submodule, trying to figure that out.

PR #405, opened 2025-05-20
IVS-327 - SPS005

Ghesselink:
See also https://github.com/buildingSMART/ifc-gherkin-rules/issues/115


features/rules/SPS/SPS005_Simultaneous-spatial-relationships.feature
```diff
@@ -0,0 +1,13 @@
...
+@E00040
+Feature: SPS005 - Simultaneous spatial relationships
+The rule verifies that an IfcElement does not simultaneously act as a child in two or more spatial relationships. For example, an element 
+cannot participate in both nesting and aggregating children roles.
+
+  Scenario: Constraints on spatial relationships
+
+    Given An .IfcElement.
+    
+    Then ^Exactly^ one of the following relationships must be non-empty: 'Nests', 'Decomposes', 'ContainedInStructure', 'AdheresToElement'
```
aothms, line 13:
```suggestion
    Then ^Exactly^ '1' of the following relationships must be non-empty: 'Nests', 'Decomposes', 'ContainedInStructure', 'AdheresToElement'
```

^ maybe make it a bit easy on ourselves and consistency with other rules

features/steps/thens/relations.py
```diff
@@ -431,3 +431,20 @@ def step_impl(context, inst):
 
     if inst not in context.visited_instances:
         yield ValidationOutcome(inst=inst, severity=OutcomeSeverity.ERROR)
+
+
+@gherkin_ifc.step("^{stmt}^ {num} of the following relationships must be non-empty: '{inverse_attrs}'")
```
aothms, line 436:
```suggestion
@gherkin_ifc.step("^{stmt}^ {num:d} of the following relationships must be non-empty: '{inverse_attrs}'")
```

features/steps/thens/relations.py
```diff
@@ -431,3 +431,20 @@ def step_impl(context, inst):
...
+def step_impl(context, inst, num, stmt, inverse_attrs):
+    text_to_num = {
+    'zero': 0, 'one': 1, 'two': 2, 'three': 3,
+    }
+    attrs = [item.strip().strip("'") for item in inverse_attrs.split(',')]
+    count = sum(bool(getattr(inst, attr, None)) for attr in attrs)
```
aothms, line 442: Be sure to double check on IFC2X3 which does not have AdheresToElement

aothms (approved):
Looks good, but please do some checks on real world models.

features/steps/thens/relations.py
Ghesselink: It also doesn't have Nests. When I upload a model with IFC2X3, this will then return None. 
```
context.model.schema
'IFC2X3'
attrs
['Nests', 'Decomposes', 'ContainedInStructure', 'AdheresToElement']
[getattr(inst, attr, None) for attr in attrs]
[None, (#939=IfcRelAggregate...92,(#869)),), (), None]
```

And in case of multiple spatial relationships;
```
[getattr(inst, attr, None) for attr in attrs]
[None, (#939=IfcRelAggregate...92,(#869)),), (#554=IfcRelContained...2,#869),$),), None]
# count is still 2
sum(bool(getattr(inst, attr, None)) for attr in attrs)
2

```

Ghesselink:
Did some checks on models from the validation service, but could not find any models that were failing for this rule. 
When I modified them manually, it still works. It's also quite straightforward; the only thing we're checking is the number of non-empty attributes of a list of 4. There's not much potential for unexpected behavior, famous last words ..
image

aothms:
Great work! Thanks for double checking on some models

PR #408, opened 2025-06-01
IVS-161 - GRF003

Ghesselink:
> When there is a Building or an Alignment in the model, either IfcProjectedCRS or IfcGeorephicCRS is provided. Relying on IfcSite placement or attributes for georeferencing doesn't qualify as proper georeferencing
> 
> EDIT
> 
> Another option for this rule is:
> 
> When there is an IfcSite with attributes RefLatitude, RefLongitude or RefElevation (since these are bad georeferencing) 
> Then there must be either IfcProjectedCRS or IfcGeographicCRS (since these are the best way to convey such info)
> 

Current rule implementation tests both scenarios, or should we check only one?

--- 
Also, in IFC4X3_ADD2 models, the presence of an IfcAlignment or IfcBuilding alone already activates the georeferencing function part; is that something we want?

From https://github.com/buildingSMART/ifc-gherkin-rules/pull/409
> Lastly, GRF003 also implicates that a model contains georeferencing when an IfcBuilding or IfcAlignment is present. Is this something we want? If not, I see two options;
> 
> 1) Add something to the step to exclude something from activation within the decorator (e.g. Given an .IfcAlignment. not activating the functional part).
>  2)   Removing the scenario entirely.

--- 
+ Includes schema error fixes (invalid guid) on a bunch of testfiles. 

```
    #2001	IfcAlignmentSegment	Error	On instance:
        #2001=IfcAlignmentSegment('39fS3hA8n5YOBvYphrBOIP',#1,$,$,$,#2501,#2631,#2002)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^                            
    IfcGloballyUniqueId base64 validation:
        The attribute GlobalId should be valid base64 encoded 128-bit number.
    Violated by:
        Guid first character must be either a 0, 1, 2, or 3.
```


civilx64:
> Current rule implementation tests both scenarios, or should we check only one?

Just the first one.

 > Includes schema error fixes (invalid guid) on a bunch of testfiles.

The screenshot says that the GlobalId is invalid because it must start with 0, 1, 2, or 3 but the flagged entity has GlobalId starting with `3`.  Why is this invalid?

features/rules/GRF/GRF003_CRS-presence-with-spatial-entities.feature
```diff
@@ -0,0 +1,37 @@
+@industry-practice
+@GRF
+@version1
+@E00010
+Feature: GRF003 - CRS presence with spatial entities
+The rule verifies that when spatial elements like buildings or alignments are present in a model, proper georeferencing is established using an coordinate reference system.
```
civilx64, line 6:
```suggestion
The rule verifies that when buildings or alignments are present in a model, proper georeferencing is established using a coordinate reference system. 
```

features/rules/GRF/GRF003_CRS-presence-with-spatial-entities.feature
```diff
@@ -0,0 +1,37 @@
+@industry-practice
+@GRF
+@version1
+@E00010
+Feature: GRF003 - CRS presence with spatial entities
+The rule verifies that when spatial elements like buildings or alignments are present in a model, proper georeferencing is established using an coordinate reference system. 
+Use of RefLatitude, RefLongitude, or RefElevation on IfcSite does not constitute valid georeferencing on its own.
```
civilx64, line 7:
```suggestion
Use of RefLatitude, RefLongitude, or RefElevation on IfcSite does not constitute valid georeferencing due to the large horizontal distances typically encountered in the infrastructure domain.
```

features/rules/GRF/GRF003_CRS-presence-with-spatial-entities.feature
```diff
@@ -0,0 +1,37 @@
...
+    Given .<attribute>. ^is not^ empty
+
+    Then There must be at least 1 instance(s) of IfcCoordinateReferenceSystem
+
+    Examples: 
+        | attribute    | 
+        | RefLatitude  |
+        | RefLongitude | 
+        | RefElevation |
+
```
civilx64, line 24:
```suggestion
```

test/files/GRF/grf003/fail-grf003-site_georeferencing_attrs_no_crs.ifc
civilx64, line 1: delete - not needed

test/files/GRF/grf003/fail-grf003-building_georeferencing_defined_on_site.ifc
civilx64, line 1: delete - not needed

test/files/GRF/grf003/pass-grf003-correct_site_and_building_georeferencing.ifc
civilx64, line 1: rename this to remove the reference to attribution on `IfcSite`.

features/rules/GRF/GRF003_CRS-presence-with-spatial-entities.feature
Ghesselink: Do we delete this altogether or is there still a space for it in another IBP-rule (not necessarily for GRF)?

features/rules/GRF/GRF003_CRS-presence-with-spatial-entities.feature
Ghesselink: Will delete the whole sentence since the scenario is also removed.

features/rules/GRF/GRF003_CRS-presence-with-spatial-entities.feature
Ghesselink: [Deleted](https://github.com/buildingSMART/ifc-gherkin-rules/pull/408/commits/36467ee7f0f9fa4e640a1b397085f723f495947a)

test/files/GRF/grf003/pass-grf003-correct_site_and_building_georeferencing.ifc
Ghesselink: [Done](https://github.com/buildingSMART/ifc-gherkin-rules/pull/408/commits/36467ee7f0f9fa4e640a1b397085f723f495947a)

test/files/GRF/grf003/fail-grf003-site_georeferencing_attrs_no_crs.ifc
Ghesselink: [Deleted](https://github.com/buildingSMART/ifc-gherkin-rules/pull/408/commits/36467ee7f0f9fa4e640a1b397085f723f495947a)

features/rules/GRF/GRF003_CRS-presence-with-spatial-entities.feature
Ghesselink: [Deleted](https://github.com/buildingSMART/ifc-gherkin-rules/pull/408/commits/36467ee7f0f9fa4e640a1b397085f723f495947a)

features/steps/validation_handling.py
```diff
@@ -175,8 +175,9 @@ def handle_then(context, fn, **kwargs):
     # ensure the rule is not activated when there are no instances
     # in case there are no instances but the rule is applicable (e.g. SPS001),
     # then the rule is still activated and will return either a pass or an error
+    # an exception is when the feature tags contain '@not-activation'
     is_activated = any(misc.recursive_flatten(instances)) if instances else context.applicable
-    if is_activated:
+    if is_activated and not ('no-activation' in context.tags and context.config.userdata.get('execution_mode') == 'ExecutionMode.PRODUCTION'):
```
aothms, line 180: Why only execution_mode=production?

aothms:
> but I don't see a need to add IfcSite

That's the confusing thing about site. It has so many aspects used to have spatial reference (RefLatitude), geography representation, plot id properties.

But agreed, let's stick to Facility, conceptually clear and in 99.9% of the cases we have one.

features/steps/validation_handling.py
Ghesselink: My idea was that this way, in CI/CD and locally, we can still determine `pass` testfiles are actually validated by one of the `Then `statements.

aothms: Ok, it doesn't really hurt, but I think the idea was that we would keep the behaviour as much the same as possible so this is a bit weird.

civilx64:
@Ghesselink docs should be updated here: https://github.com/buildingSMART/validate/blob/development/docs/dev/gherkin_deep_dive.md.

Add a new heading for Feature Tags and include `@no-activation`.

Ghesselink:
Done in https://github.com/buildingSMART/validate/pull/193

PR #411, opened 2025-06-08
IVS-236 - GRF005 - CRS unit type differences

Ghesselink:
This rule is a bit more complex and less straightforward, some points of (potential) discussion; 

- When there are no units linked to the CRS, we consider the units based on the EPSG code. 
- Should we check other attributes as well? E.g. VerticalDatum might refer to a ESPG with different units compared to the project units?
- In testfiles containing `IfcGeographicCRS` there is often no `IfcMapConversion`; do I miss something?
- How do we handle project units in foot? They are often not associated directly through units but through `IfcConversionBasedUnit` (not through IfcProject.UnitsInContext.Unit).
- What about millimetre/metre? This is done with a prefix in the project units and is considered in the step implementation. 

Is this potential for a rule? the `VerticalDatum` of a CRS is described as:

> Name by which the vertical datum is identified. The vertical datum is associated with the height axis of the coordinate reference system and indicates the reference plane and fundamental point defining the origin of a height system. It needs to be provided, if the Name identifier does not unambiguously define the vertical datum as well and if the coordinate reference system is a 3D reference system.


test/files/GRF/grf005/pass-grf005-scaled_foot_to_metre.ifc
```diff
@@ -0,0 +1,34 @@
...
+#13=IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.);
+#14=IFCSIUNIT(*,.AREAUNIT.,$,.SQUARE_FOOT.);
+#15=IFCSIUNIT(*,.VOLUMEUNIT.,$,.CUBIC_FOOT.);
+#16=IFCSIUNIT(*,.PLANEANGLEUNIT.,$,.RADIAN.);
+#17=IFCMEASUREWITHUNIT(IFCPLANEANGLEMEASURE(0.017453292519943295),#16);
+#18=IFCCONVERSIONBASEDUNIT(#12,.PLANEANGLEUNIT.,'DEGREE',#17);
+#19=IFCUNITASSIGNMENT((#13,#14,#15,#18));
+#20=IFCPROJECT('2X9wjf5oPB4PQjenCYrhHZ',#5,'',$,$,$,$,(#11),#19)
+#21=IFCPROJECTEDCRS('EPSG:2277',$,'NAD83',$,'','3',$);
+#22=IFCMAPCONVERSION(#11,#21,316131.64,5690966.11,1.,1.,0.,0.3046);
```
civilx64, line 30: EPSG:2277 identifies the length unit as [US Survey Foot](https://en.wikipedia.org/wiki/Foot_(unit)#US_survey_foot) which is not the same as international foot (exactly 0.3048 m).  In either case, 0.3046 is incorrect. In this case it should be:

```shell
Python 3.12.6 (v3.12.6:a4a2d2b0d85, Sep  6 2024, 16:08:03) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 1200 / 3937
0.3048006096012192
```

Let the record state that I am merely providing information and *not* advocating for perpetuation of this obsolete length measure.

test/files/GRF/grf005/pass-grf005-scaled_foot_to_metre.ifc
```diff
@@ -0,0 +1,34 @@
...
+#23=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.E-05,#9,#10);
```
civilx64, line 31: Appears to be a duplicate of # 11.

civilx64:
> Should we check other attributes as well? E.g. VerticalDatum might refer to a ESPG with different units compared to the project units?

I am leaning towards checking all three attributes that potentially refer to an EPSG code, but I can't say how often this actually happens.  If there are different units for HorizonalDatum and VerticalDatum, that would indicate the need to use IfcMapConversionScaled.

civilx64:
> In testfiles containing IfcGeographicCRS there is often no IfcMapConversion; do I miss something?

Read the docs for [IfcCoordinateOperation](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcCoordinateOperation.htm) if you have not already.

IfcGeographicCRS can be used with IfcRigidOperation. 
See `test/files/GRF/grf005/na-grf005-geographic_referencing_no_map_conversion.ifc` for example.

features/rules/GRF/GRF005_CRS-unit-type-differences.feature
```diff
@@ -0,0 +1,26 @@
...
+
+  Scenario Outline: When the length unit of the Local CRS (from IfcProject) is not equal to the length unit of the Projected CRS, then the IfcMapCOnversion.Scale must be provided and cannot be 1.0
+
+    Given A model with Schema 'IFC4.3'
+    Given An .IfcMapConversion.
+    Given There must be at least 1 instance(s) of .<IfcCoordinateReferenceSystem>.
+    Given The <unit> unit(s) of the project ^is not^ equal to the <unit> unit(s) of the .<IfcCoordinateReferenceSystem>.
+    
+    Then .Scale. ^is not^ empty
+    Then .Scale. ^is not^ 1.0
```
aothms, line 20: Rather than requiring not empty and not 1.0, can we require the actual value that is needed? I.e the one conversion factor divide by the other.

features/rules/GRF/GRF005_CRS-unit-type-differences.feature
```diff
@@ -0,0 +1,24 @@
+@industry-practice
+@GRF
+@version1
+@E00100
+Feature: GRF005 - CRS unit type differences
+The rule verifies that a map conversion scale is explicitly defined and correctly set when the length and/or angle units used in the local engineering coordinate system differ from those in the referenced CRS (Projected or Geographic). 
+If these units differ, assuming a scale of 1.0 would result in incorrect spatial positioning, and the scale must therefore be explicitly provided with the correct value.
```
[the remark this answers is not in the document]
Ghesselink: https://github.com/buildingSMART/ifc-gherkin-rules/pull/411/commits/c706d07985d1c005a21967874690dd8658f709c3

test/files/GRF/grf005/fail-grf005-scaled_foot_to_metre.ifc
```diff
@@ -0,0 +1,34 @@
+ISO-10303-21;
+/* IFC units are in metres, ESPG:2277 uses survey foot and the map conversion reflects this */
```
civilx64:
```suggestion
/* IFC units are in metres, ESPG:2277 uses survey foot and the map conversion uses an incorrect unit conversion factor */
```

PR #417, opened 2025-06-16
SPS008  - spatial container representations (IVS-537)

civilx64:


features/rules/SPS/SPS008_Spatial-container-representations.feature
```diff
@@ -0,0 +1,40 @@
+@industry-practice
+@SPS
+@version1
+
+Feature: SPS008 - Spatial Container Representations
+  The rule verifies that certain spatial containers do not have a geometric representation,
+  but are rather represented via their contained elements.
```
aothms, line 7:
```suggestion
  but are rather solely a container for other elements or aggregated into other facility parts.
```

I would rather not use the word represented (but this is subjective) I would mention both containment as well as aggregation e.g Building -> Storey; Bridge -> BridgePart is aggregation.

aothms (approved):
Nice

PR #418, opened 2025-06-21
LIP002 empty instance info (IVS-540)

civilx64:
Addresses issue noted in buildingSMART/IFC4.3.x-sample-models/pull/27

### DIFF
--- a/features/rules/LIP/LIP002_Linear-placement-fallback-coordinates.feature
+++ b/features/rules/LIP/LIP002_Linear-placement-fallback-coordinates.feature
@@ -11,8 +11,7 @@ Feature: LIP002 - Linear placement fallback coordinates
 
     Scenario: Confirm the presence of CartesianPosition
 
-      Given Its attribute .CartesianPosition.
-      Then Assert existence
+      Then the value of attribute .CartesianPosition. must be ^not empty^
 
     Scenario: Confirm the values of CartesianPosition
 
--- a/features/steps/thens/existence.py
+++ b/features/steps/thens/existence.py
@@ -93,7 +93,7 @@ def step_impl(context, inst):
     if not any(misc.recursive_flatten(inst)):
         expected = get_previous_step_before_assertion(context)
         yield ValidationOutcome(
-            instance_id=inst,
+            inst=inst,
             expected=expected,
             observed="Nonexistent",
             severity=OutcomeSeverity.ERROR,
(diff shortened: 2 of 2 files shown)


Ghesselink (commented):
This should not block the current PR, but it seems that LIP002 is not the only rule that is affected by this inconsistency. Indeed, we use `instance_id=X` for passed outcomes / [given](https://github.com/buildingSMART/ifc-gherkin-rules/blob/f3ae68c722a2f9b14f01d702f6c172a359d6b848/features/steps/validation_handling.py#L147) statements and `inst=Y` for failing outcomes / then statements. 
I went over all our gherkin step implementations and found a couple of other inconsistencies: 
- In [attributes.py](https://github.com/buildingSMART/ifc-gherkin-rules/blob/f3ae68c722a2f9b14f01d702f6c172a359d6b848/features/steps/thens/attributes.py#L136-L143)
- In [requirements.py](https://github.com/buildingSMART/ifc-gherkin-rules/blob/f3ae68c722a2f9b14f01d702f6c172a359d6b848/features/steps/givens/relationships.py#L65) (passing instance_Id for both given and then statements)
- In [entities.py](https://github.com/buildingSMART/ifc-gherkin-rules/blob/f3ae68c722a2f9b14f01d702f6c172a359d6b848/features/steps/givens/entities.py#L48) (appears unused)
- In [attribute_value.py](https://github.com/buildingSMART/ifc-gherkin-rules/blob/f3ae68c722a2f9b14f01d702f6c172a359d6b848/features/steps/steps/attribute_value.py#L63)

I still have to figure out what the consequences are in terms of outcomes not correctly being transferred over to the next gherkin step and/or database. We should probably make it easier for ourselves and use either 'instance_id' or 'inst' for all our outcomes.

civilx64:
I can revert the `Assert existence` step implementation as I ended up revising LIP002 to use a different approach.

aothms:
I've also been bitten by this in the past. I have the suspicion this is actually a left-over inconsistency from when we merged in our decorator work a bit hastily.

> use instance_id=X for passed outcomes / [given](https://github.com/buildingSMART/ifc-gherkin-rules/blob/f3ae68c722a2f9b14f01d702f6c172a359d6b848/features/steps/validation_handling.py#L147) statements and inst=Y for failing outcomes / then statements

Can't we simply unify this instead. Because
1. It's rather ugly, because it is not an id we're passing to it, so really confusing
2. Doesn't it prevent using given steps as then criteria and vice versa, this was the aim in the past right?

Maybe rename it to value because it is not always an (entity) instance

PR #420, opened 2025-06-21
GRP000 Groups (IVS-337)

civilx64:
(IVS-337)


civilx64:
We have GRP001 already active and explicitly includes systems (and zones as a subtype of systems) in the feature description.  SYS000 is indeed in the backlog but SYS001 is disabled.  My preference would be to leave this rule as-is, especially since the feature description references a concept template that includes systems.  We could still move forward with SYS000 and scope it to IfcSystem and subtypes.  A model correctly utilizing those entities would then activate the GRP and SYS functional parts, which IMHO makes sense.  It seems logical that a zone is a group of spaces and a system is a group of things like electrical circuits or other distribution systems like stormwater drainage pipes, inlet, and manholes.

PR #422, opened 2025-06-26
IVS-550 - Catch infinite json values

Ghesselink:


civilx64 (approved):

features/steps/validation_handling.py
```diff
@@ -328,6 +328,15 @@ def expected_behave_output(context: Context, data: Any, is_observed : bool = Fal
             data = ast.literal_eval(data)
         except (ValueError, SyntaxError):
             pass
+    
+    def ensure_finite_numbers(obj):
+        if isinstance(obj, float) and not math.isfinite(obj):
+            return "infinity"
```
aothms, line 334: 1. Just one more comment that not-a-number values will also be cast to infinity in this approach

```
>>> math.isfinite(float('nan'))
False
```

2. And also note that the behaviour of handling infinity is inconsistent to whether using numpy (warning) or built-in python (exception)

Maybe we should still adapt the rule to never divide by zero?

```
>>> 1/0
ZeroDivisionError: division by zero
>>> import numpy as np
>>> np.array([1.])[0]
np.float64(1.0)
>>> np.array([1.])[0] / 0.
<stdin>:1: RuntimeWarning: divide by zero encountered in scalar divide
np.float64(inf)
```

PR #432, opened 2025-07-31
Fixed typing of the IfcOffsetCurveByDistances entity in ALS005

fg55b:
Fixed the typing of the _IfcOffsetCurveByDistances_ entity on **ALS005 - Alignment shape representation**.

### DIFF
--- a/features/rules/ALS/ALS005_Alignment-shape-representation.feature
+++ b/features/rules/ALS/ALS005_Alignment-shape-representation.feature
@@ -2,7 +2,7 @@
 @ALS
 @version2
 Feature: ALS005 - Alignment shape representation
-The rule verifies that each IfcAlignment uses correct representation.
+  The rule verifies that each IfcAlignment uses correct representation.
 
   Background:
     Given A model with Schema 'IFC4.3'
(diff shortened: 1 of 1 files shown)


features/rules/ALS/ALS005_Alignment-shape-representation.feature
```diff
@@ -2,7 +2,7 @@
 @ALS
 @version2
 Feature: ALS005 - Alignment shape representation
-The rule verifies that each IfcAlignment uses correct representation.
+  The rule verifies that each IfcAlignment uses correct representation.
```
Ghesselink, line 5: Sorry for the nitpicking, the space is not needed
```suggestion
The rule verifies that each IfcAlignment uses correct representation.
```

Ghesselink (approved):
Thanks :)

features/rules/ALS/ALS005_Alignment-shape-representation.feature
fg55b: Thanks for the comment, Geert. The 'format on save' feature was enabled.

Already fixed!

PR #446, opened 2025-09-08
IVS-635 - Improve caching of models in between features

Ghesselink:
| Size (MB) | T1 | T2 | Impr         |  
| :-------- | :--------------- | :--------------- | :------------ | 
| 31.46   | 305.02 s         | 80.75 s          | x3.87 | 
| 3.07    | 37.97 s          | 13.74 s          | 2.76 | 
| 0.93    | 16.05 s          | 7.96 s           | 2.01 |  
| 0.22    | 6.05 s           | 3.53 s           | 1.71 |  
| 0.01    | 2.25 s           | 0.90 s           | 2.5 |

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -15,12 +15,9 @@ from validation_results import ValidationOutcome, ValidationOutcomeCode, Outcome
 from main import ExecutionMode
 
 @functools.cache
-def _open_model(fn, pure):
+def read_model(fn, pure):
     return (ifcopenshell.simple_spf.open(fn)
             if pure else ifcopenshell.open(fn))
-
-def read_model(fn, pure):
-    return _open_model(fn, pure)
 
 def print_directory_tree(start_path, level=0):
     """


features/environment.py
```diff
@@ -13,16 +14,13 @@
...
-    else:
-        model_cache[(fn, pure)] = ifcopenshell.open(fn)
-    return model_cache[(fn, pure)]
+@functools.cache
+def _open_model(fn, pure):
+    return (ifcopenshell.simple_spf.open(fn)
+            if pure else ifcopenshell.open(fn))
 
+def read_model(fn, pure):
+    return _open_model(fn, pure)
```
aothms, line 23:
```suggestion
@functools.cache
def read_model(fn, pure):
    return (ifcopenshell.simple_spf.open(fn)
            if pure else ifcopenshell.open(fn))
```

I don't think any extra indirection is needed. As long as the functools.cache is a top-level item in the namespace so that the decorator is applied only once.

aothms (changes requested):

PR #451, opened 2025-09-15
PPP (pure python parser) header only

aothms:
Before

mem_plot_20250915_121451

After

mem_plot_20250915_114929

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -16,7 +16,8 @@ from main import ExecutionMode
 
 @functools.cache
 def read_model(fn, pure):
-    return (ifcopenshell.simple_spf.open(fn)
+    # @nb --purepythonparser is only used for @critical rules which is only IFC101 which only looks at the header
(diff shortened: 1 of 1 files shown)


Ghesselink (approved):
Sorry, I already had something done yesterday evening but didn't create a PR yet. Nice to see the 1400MB (!) spike gone, that will be helpful. Do you know which process the purple line represents (1455)?

This solution looks a lot simpler, just one line difference. 
IFC101 is indeed the only gherkin rule that uses the step file parser, do we expect other rules to use the parser as well? For instance, IFC102 is just a simple entity lookup and `ifcopenshell.simple_spf.open(fn).data_` would be sufficient too.

aothms:
`4155` (first plot) `1897` (second plot) is the implementer agreement gherkin check. Hopefully with the reduction of pythons-side memory usage we'll have a reduction there as well.

PR #454, opened 2025-10-09
Validation for proper re-use of horizontal layout (IVS-620)

### LINKED ISSUE #442: Improper reusing horizontal - False Positive
The attached example illustrates an improper use of [CT 4.1.4.4.1.1](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Composition/Nesting/Alignment_Layouts/Alignment_Layout_-_Horizontal,_Vertical_and_Cant/content.html). The alignment contains two vertical layouts and the validation service does not flag it as an error.

The structure is:
IfcAlignment-IfcRelNests-IfcReferent,IfcAlignmentHorizontal,IfcAlignmentVertical,IfcAlignmentVertical.

Reusing a horizontal with two vertical is governed by [CT 4.1.4.4.1.2](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Composition/Nesting/Alignment_Layouts/Alignment_Layout_-_Reusing_Horizontal_Layout/content.html)

[Improper_Reusing_Horizontal.ifc.txt](https://github.com/user-attachments/files/21953530/Improper_Reusing_Horizontal.ifc.txt)

civilx64:
Closes #442.


features/rules/ALB/ALB031_Alignment-layouts-default-case.feature
```diff
@@ -0,0 +1,18 @@
...
+which represents the default case for nesting alignment layouts.
+Alignment layouts are abbreviated as follows: horiz = IfcAlignmentHorizontal, vert = IfcAlignmentVertical,
+and cant = IfcAlignmentCant
+
+  Scenario: Agreement on IfcAlignment layout nesting per 4.1.4.4.1.1.
+
+      Given A model with Schema 'IFC4.3'
+      Given an .IfcAlignment.
+      Given .IsDecomposedBy. ^is^ empty
+      Given its attribute .IsNestedBy.
```
aothms: I think it reads a bit clearer if you use the relationship predicate here, either:

(a) check for existance

```suggestion
    Given a relationship .IfcRelNests. from .IfcAlignment. to .IfcObject.
```

(b) check for existence and move focus

```suggestion
    Given a relationship .IfcRelNests. from .IfcAlignment. to .IfcObject. and following that
```

features/rules/ALB/ALB031_Alignment-layouts-default-case.feature
```diff
@@ -0,0 +1,18 @@
...
+
+      Then the alignment layouts must include [1 horiz, or 1 horiz and 1 vert, or 1 horiz and 1 vert and 1 cant]
```
aothms: I think in other rules we don't use `{comma} or` but just `or`.

We typically also have "token separators" part of the splitting, so you would do `str.split('] or [')`

```suggestion
      Then the alignment layouts must include [1 horiz] or [1 horiz and 1 vert] or [1 horiz and 1 vert and 1 cant]
```

features/rules/ALB/ALB032_Alignment-layouts-reusing-horizontal.feature
```diff
@@ -0,0 +1,19 @@
...
+  Scenario: Agreement on IfcAlignment layout nesting per 4.1.4.4.1.1.
+
+      Given A model with Schema 'IFC4.3'
+      Given an .IfcAlignment.
+      Given its attribute .IsDecomposedBy.
+      Given its attribute .RelatedObjects.
+      Given its attribute .IsNestedBy.
+
+      Then the alignment layouts must include [1 vert, or 1 vert and 1 cant]
+
```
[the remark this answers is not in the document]
civilx64: > allowing to introduce more specific profiles,

`profiles` here is confusing because

- a domain expert thinks of it as "vertical layout"
- an IFC expert thinks of it as "open or closed 2D shape to be extruded or swept"
- your usage suggests "ontological construct for information theory" (I think ..?)

civilx64:
> Let me rephrase my question, in my way of schematising earlier (where additional indentation is aggregation and layouts (thanks) are within parens) is the following scenario now correctly flagged by the rule and how?

The first of these examples is valid.  The second is not (because you need a vertical layout before you can introduce cant).  I will add unit test files for both.

In the root-to-leaf examples, none of those are valid because

1. You can only have parent and child (no grandchildren or great-grandchildren)
2. The parent alignment A does not have any layouts.  It must have at least 1 and no more than 1 ([1:1] in EXPRESS) horizontal layouts

These requirements may not be explicit in the spec but they are obvious to industry practitioners.  I highly doubt that any authoring tool would export data with multiple depths of aggregation, so I don't see value in creating additional rules to check and enforce them.  At minimum we could capture in the backlog, but we needn't worry about putting additional logic into ALB031 and 032.

It may be helpful to remember that this rule is in the ALB (alignment business logic) functional part.  The representations for alignment geometry corresponding to business logic are

| Entity | Corresponds to layouts of |
| - | - |
| IfcCompositeCurve | horizontal only |
| IfcGradientCurve | horizontal plus vertical |
| IfcSegmentedReferenceCurve | horizontal plus vertical plus cant |

In practice it is common to have one horizontal layout that corresponds to at least 2 and often 5 or more vertical layouts.  That is why we have the concept template for aggregating multiple vertical layouts to a horizontal layout.

civilx64:
I re-read it again today (on an actual monitor, not on phone) and agree that the intent of the language in the concept template is that a "parent" only has 1 and exactly 1 horizontal layouts - no vertical or cant (as they are provided on the "child" alignments).

So the VS needs to identify and enforce three scenarios:

1. Alignment is "independent" (no re-use of layouts via aggregation) - in ALB031
2. Alignment is "parent" - IfcRelAggregates.RelatingObject - (horizontal layout only) - in ALB032
3. Alignment is "child" - IfcRelAggregates.RelatedObjects - (vertical layout only or vertical layout + cant layout) - in ALB032

features/rules/ALB/ALB031_Alignment-layouts-default-case.feature
```diff
@@ -0,0 +1,19 @@
...
+The rule verifies that alignments are nested in accordance with Concept Template 4.1.4.4.1.1
+which represents the default case for nesting alignment layouts.
+Alignment layouts are abbreviated as follows: horiz = IfcAlignmentHorizontal, vert = IfcAlignmentVertical,
+and cant = IfcAlignmentCant
+
+  Scenario: Agreement on IfcAlignment layout nesting per 4.1.4.4.1.1.
+
+      Given A model with Schema 'IFC4.3'
+      Given an .IfcAlignment.
+      Given a relationship .IfcRelAggregates. ^does not exist^ from .IfcAlignment. to .IfcAlignment. and following that
```
aothms: I'm a bit confused by the combination of `^does not exist^` and `and following that`.  If it doesn't exist you can also not follow (traverse) along that relationship edge.

```suggestion
      Given a relationship .IfcRelAggregates. ^does not exist^ from .IfcAlignment. to .IfcAlignment.
```

features/rules/ALB/ALB032_Alignment-layouts-reusing-horizontal.feature
```diff
@@ -0,0 +1,25 @@
...
+    Given a *required* relationship .IfcRelAggregates. from .IfcAlignment. to .IfcAlignment.
+    Given a relationship .IfcRelNests. from .IfcAlignment. to .IfcObject.
+    Then the alignment layouts must include [1 horiz]
+
+Scenario: Agreement on IfcAlignment layout nesting per 4.1.4.4.1.1 for "child" alignment
+
+    Given a *required* relationship .IfcRelAggregates. from .IfcAlignment. to .IfcAlignment. and following that
+    Given a relationship .IfcRelNests. from .IfcAlignment. to .IfcObject.
+    Then the alignment layouts must include [1 vert] or [1 vert and 1 cant]
+
```
aothms: Since we don't follow the root-to-leaf approach there is at least one more scenario necessary to constrain that aggregation stops at one level below the root.

Something like:

    Given a relationship .IfcRelAggregates. from .IfcAlignment. to .IfcAlignment. and following that
    Then a relationship .IfcRelAggregates. ^does not exist^ from .IfcAlignment. to .IfcAlignment.

features/rules/ALB/ALB032_Alignment-layouts-reusing-horizontal.feature
```diff
@@ -0,0 +1,25 @@
...
+Alignment layouts are abbreviated as follows: horiz = IfcAlignmentHorizontal, vert = IfcAlignmentVertical,
+and cant = IfcAlignmentCant
+
+Background:
+    Given A model with Schema 'IFC4.3'
+    Given an .IfcAlignment.
+
+Scenario: Agreement on IfcAlignment layout nesting per 4.1.4.4.1.1 for "parent" alignment
+
+    Given a *required* relationship .IfcRelAggregates. from .IfcAlignment. to .IfcAlignment.
```
aothms:
```suggestion
    Given a relationship .IfcRelAggregates. from .IfcAlignment. to .IfcAlignment.
```

features/rules/ALB/ALB032_Alignment-layouts-reusing-horizontal.feature
aothms: In the JSON

    "exist_or_not_exist": ["exists", "must exist", "does not exist"],

Maybe we should add a "must not exist" and use it here

civilx64: > Since we don't follow the root-to-leaf approach there is at least one more scenario necessary to constrain that aggregation stops at one level below the root.

Do we really expect implementers to aggregate layouts deeper than one 'child' level?  I'm of the opinion that the scenario implementations as-is are "good enough" to enforce the concept templates - and most importantly address the false positive raised in the original issue #442.

civilx64: > I wouldn't see why not

As I understand it, proprietary tools map their internal data structures to IFC when exporting.  No authoring tool would have vertical or cant layouts that are children of another vertical or cant layout.  It doesn't make any sense in the context of alignment design for road/rail.

That said, I will take a shot at adding a third scenario per the `must not exist` suggestion upthread.

PR #458, opened 2025-10-22
IVS-324 Rule catalog

aothms:
See https://buildingsmart.github.io/ifc-gherkin-rules/branches/development

#### Focussed on hyperlinks

Steps are clickable

afbeelding

as well as usages

afbeelding

#### Flat listing is a bit messy

We should probably create intermediate pages for the func parts and have a higher depth TOC

Not sure what categorisation we can use for the steps, probably file name.

#### TODO

- [x] Disabled rules still need to be marked as such
- [ ] Version info would be nice to list also on the feature page
- [ ] Lots of docstrings in step implementation functions
- [ ] Intermediate levels as func parts and step impl file name
- [ ] Mark `@proposed` rules (see below)
- [x] Undo some temporary commits ~and only deploy from main branch~ (we now have side-by-side output for branches so no longer restricted to main branch)
- [x] Reference from validate docs and from UI
- [ ] Unique function names for step impl, to have persistent ids and readable names in the toc. Now they're all called `step_impl()`

#### Future rules

My proposal is to add every entry in the rule catalog sheet as a placeholder file in the repo with a `@proposed` tag along with a basic description (no scenarios) That way:

- They will show up here in the catalog
- More easy to guarantee there are no duplicate rule ids

### DIFF
--- a/docs/git.py
+++ b/docs/git.py
@@ -14,42 +14,6 @@ def git(*args, cwd=REPO_ROOT):
 def introduced_commit(path, repo):
     return (git("log", "--follow", "--diff-filter=A", "--format=%H", "-1", "--", path, cwd=repo) or '').strip()
 
-def parents_of(commit, repo):
-    line = git("rev-list", "--parents", "-n", "1", commit, cwd=repo)
-    parts = line.split()
-    return parts[1:] if len(parts) > 1 else []
-
-def diffs_touching_pattern(pathspec, pattern, repo):
-    out = git("log", "-G", pattern, "--format=%H", "--follow", "--", pathspec, cwd=repo)
-    return [l for l in out.splitlines() if l]
-
-def first_parent(commit, repo):
-    line = git("rev-list", "--parents", "-n", "1", commit, cwd=repo)
-    parts = line.split()
-    return parts[1] if len(parts) > 1 else None
-
-def commits_following_renames(path, repo):
-    out = git("log", "--follow", "-M", "--format=%H", "--", path, cwd=repo)
-    return [l for l in out.splitlines() if l]
-
-def parent_path_for_commit(commit, child_path, repo):
-    ns = git("diff-tree", "-M", "--name-status", "-r", commit, cwd=repo)
-    for line in ns.splitlines():
-        if not line or line[0] != "R":
-            continue
-        parts = line.split("\t")
-        if len(parts) == 3:
-            _, oldp, newp = parts
-            if newp == child_path:
-                return oldp
-    return child_path
-
-def blob_diff(parent, parent_path, commit, child_path, repo):
-    if parent:
-        return git("diff", "-U0", f"{parent}:{parent_path}", f"{commit}:{child_path}", cwd=repo)
-    else:
-        return git("show", "-U0", commit, "--", child_path, cwd=repo)
-
 
 def tags_for_commit(commit, repo):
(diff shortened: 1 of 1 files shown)


civilx64 (commented):
A couple of high-level comments:

1. This appears to stomp over the docs that we previously had.  I assume that the intent is for the rule catalog to supplement (not replace) the existing docs - correct?
2. What is the user story (problem statement) behind adding this rule catalog to our docs?

The user story as I understand it:

* as a user of the Validation Service
* I need easily accessible documentation
* so that I can understand what a particular rule is checking, and what (if anything) I can do as an end user to make the model pass

Some (most ?) failing rules need to be addressed by the software vendor and are out of the user's control.  So perhaps the user in this story is the developer of an authoring tool, not the user of said tool that uploads a model.

Ghesselink:
Looks really nice :-) 

--- 
 It seems there is also a fourth user story?  

- 4
  - As a validation service maintainer / rule developer
  - I want to have an insight to the usage of each step implementation (i.e. where and how often each impl is used)
 
There's some overlap with 'bystander' here.. We sometimes have some issues with unused step implementations or steps without an implementation (recently fixed by @civilx64). The catalog also helps us in this regard. 


---- 

What do you mean with 
> Intermediate levels as func parts and step impl file name

?

---

> Mark @proposed rules (see below)

Agree with this. A problem with documentation is that it can become scattered, the rule-administration is probably the best example. Can/should we also implement something that makes it easy to propose new rules?

aothms:
Right. But on the validate repo right, not ifc-gherkin-rules https://github.com/buildingSMART/validate/tree/docs/gh-pages

Anyway, I specifically opted for sphinx so that these can be integrated.

docs/git.py
```diff
@@ -0,0 +1,136 @@
...
+import sys
+from natsort import natsorted
+
+from paths import REPO_ROOT
+
+def git(*args, cwd=REPO_ROOT):
+    out = sp.run(["git", *args], cwd=cwd, check=True, capture_output=True, text=True)
+    return out.stdout.strip()
+
+def introduced_commit(path, repo):
```
Ghesselink: Some functions in this file appear unused .. Are they for the todo list?

.github/workflows/catalog.yml
```diff
@@ -0,0 +1,65 @@
...
+          pip install sphinx sphinx_rtd_theme natsort
+          wget -O /tmp/ifcopenshell_python.zip https://s3.amazonaws.com/ifcopenshell-builds/ifcopenshell-python-`python3 -c 'import sys;print("".join(map(str, sys.version_info[0:2])))'`-v0.8.3-260bc80-linux64.zip
+          mkdir -p `python3 -c 'import site; print(site.getusersitepackages())'`
+          unzip -d `python3 -c 'import site; print(site.getusersitepackages())'` /tmp/ifcopenshell_python.zip
+
+      - name: Build docs
+        run: |
+          set -eux
+          cd docs
+          python build_docs.py
```
Ghesselink: The commit I made should appear in https://buildingsmart.github.io/ifc-gherkin-rules/branches/development/, right? I'm still seeing the old version, but maybe I'm missing a step .

docs/paths.py
Ghesselink: I had a couple of (minor) local setup issues. The behavior in the vscode debugger versus the terminal was slightly different, so figured it would be a good idea to make this always consistent.

PR #462, opened 2025-11-04
IVS-633 Python-side memory optimizations - rules

aothms:
1. Implement MAX_ERROR_COUNT in handle_then() to terminate processing when an amount of errors has already been generated for a feature (we don't show them in the UI, no consequences on score cards)
2. Implement a light weight DTO / dataclass equivalent of ValidationOutcome not tied to django orm that is more light weight in memory (i.e `__slots__`, less intelligence, string interning).

#### before

mem_plot_20251104_123325_dev_branch

#### after

mem_plot_20251104_131117_perf_branch_w_max10

### Next steps

Running rules individually. Interestingly gives the same pronounced patterns on individual rules. Looking into these now.

find features/ -name '*.feature' -exec python -m behave --no-capture -v {} --define input=/mnt/c/Users/tkrij/Documents/AECgeeks/projects/buildingsmart/validate/backend/.dev/files_storage/Duplex_MEP_20110907.ifc --define task_id=4334 --define execution_mode=ExecutionMode.PRODUCTION \;

mem_plot_20251104_135022

rule | max mem (mb) | runtime (s)
---|---|---
ifc105 [30178] | 504.26953125 | 18.737269928999012
ctx000 [29484] | 483.46875  | 16.87474232800014
brp001 [29362] | 472.46484375  | 47.58386559099017
pse002 [30665] | 427.79296875 |  8.861968547003926
gem111 [29786] | 415.2421875  | 7.811844519004808


aothms:
Focussing on BRP001 I'm approaching a state where there is some progress. In this graph you see the before (`development` branch) and after. The `misc.PackedSequence` doesn't have the simplistic elegance I was hoping for though..

mem_plot_20251125_102934

aothms:
The spikes are more related to actual rule execution, for example IFC105 has such as spike, also when running behave directly.

mem_plot_20251125_141157

aothms:
Tests are passing 🎉  Curious for your review @Ghesselink

aothms:
> The max_outcomes_per_rule is currently not applied in validate, right? Think this will be a nice time-save. 

Ah crap, the branch was there, but somehow must have forgotten to create a PR. There is one now: https://github.com/buildingSMART/validate/pull/259

> The risk probably is that we'll exclude unique and relevant outcomes.

Yes but currently there is also no way to get to them in the report anyway. And for the scorecards 1 outcome is enough. So there are no implications.

> Address " todo his better be handled more efficiently in ifcopenshell, but idea here is to prevent allocating large lists" to get a grasp of the C++ internal mechanics.

PR #471, opened 2025-11-26
PJS001 conversion based units (IVS-680)

### LINKED ISSUE #107: Rule proposal: conversion based unit labels
IfcConversionBasedUnit has a string label for its name, but the docs have a recommended table of units. I would propose the following two rules:

- Case: use lowercase (always? or when part of the table?)
- Value: when using a name from this table make sure the value conforms to the table.

https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcConversionBasedUnit.htm

civilx64:
closes #107

### DIFF
--- a/features/steps/steps/propertysets_qtys_units.py
+++ b/features/steps/steps/propertysets_qtys_units.py
@@ -433,7 +433,7 @@ def step_impl(context, inst, attr_name):
             if attr_value not in accepted_names:
                 yield ValidationOutcome(inst=inst, expected=accepted_names, observed=attr_value,
                                         severity=OutcomeSeverity.ERROR)
-        case "CONVERSIONFACTOR":
+        case "ConversionFactor":
             unit_name = inst.Name
             if unit_name in accepted_names:
                 inst_factor = inst.ConversionFactor.ValueComponent.wrappedValue


features/resources/IFC2X3/valid_ConversionBasedUnits.csv
```diff
@@ -0,0 +1,13 @@
+Name,UnitType,ConversionFactor,SIUnitPrefix,SIUnitName,Description
+inch,LENGTHUNIT,25.4,MILLI,METRE
+foot,LENGTHUNIT,304.8,MILLI,METRE
+yard,LENGTHUNIT,914.,MILLI,METRE
+mile,LENGTHUNIT,1609.,None,METRE
+acre,AREAUNIT,4046.86,None,METRE
+litre,VOLUMEUNIT,0.001,None,METRE
```
aothms: This irks me a bit. `1l = 0.001 m^3` but you write METRE (combined with the volumeunit so it's not that bad) but still. I wonder if it's better to have an `Exponent` field. Just so that I can sleep better at night.

```suggestion
litre,VOLUMEUNIT,0.001,None,METRE,3
```


Alternatively, you could of course use dimensional exponents to summarize a lot of fields. Although you cannot encode the prefixes.

```
(M, L, T, I, Θ, N, J) := (0,  3,  0,  0,  0,  0,  0)
```

Then it would become:

```
Name,ConversionFactor,MassExponent,LengthExponent,TimeExponent,ElectricCurrentExponent,ThermodynamicTemperatureExponent,AmountOfSubstanceExponent,LuminousIntensityExponent
litre,0.001,0, 3, 0, 0, 0, 0, 0
```

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -19,6 +22,35 @@
...
+class ConversionBasedUnitDefinition:
+    """
+    used to hold data from table of conversion-based units defined in the IFC spec
+    Ref: IFC2X3 - https://standards.buildingsmart.org/IFC/RELEASE/IFC4/FINAL/HTML/schema/ifcmeasureresource/lexical/ifcconversionbasedunit.htm
+    Ref: IFC4 - https://standards.buildingsmart.org/IFC/RELEASE/IFC2x3/FINAL/HTML/ifcmeasureresource/lexical/ifcconversionbasedunit.htm
+    Ref: IFC4X3 - https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcConversionBasedUnit.htm
+    """
+    Name: str
+    UnitType: str
+    ConversionFactor: str
```
aothms: I don't know. Is this is the annotation for the constructor or the attribute accessor post construction. For the latter it's wrong. This is of course the nature of python type annotations that it doesn't matter, but I wouldn't do it in this way. I prefer that the type annotations map to the reality in the conceptual model.

And since anyway you have a post_init that does a lot of stuff. I would just create a static from_csv_row() function or something that does the type coercion and then call a clean constructor.

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -19,6 +22,35 @@
     'QTO_OCCURRENCEDRIVEN': 'IfcObject',
 }
 
+@dataclass
```
aothms:
```suggestion
@dataclass(frozen=True,slots=True,kw_only=True,repr=True,eq=True)
```

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -375,3 +409,27 @@ def step_impl(context, inst, table, inst_type=None):
             # not a universal error. This is more IDS territory.
             # if not values:
             #     yield ValidationOutcome(inst=inst, expected= {"oneOf": accepted_data_type['instance']}, observed = {'value':None}, severity=OutcomeSeverity.ERROR)
+
+@gherkin_ifc.step(
+    "Its attribute .{attr_name}. must be defined [according to the table] 'valid_ConversionBasedUnits'"
+)
+def step_impl(context, inst, attr_name):
+    table = "valid_ConversionBasedUnits"
+    unit_definitions = get_table_definition(context.model.schema, table)
```
aothms: Part of the performance work I did is move all CSV reads outside of the step_impl functions. Because you don't want to open a file for a every instance you check.

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -375,3 +409,27 @@ def step_impl(context, inst, table, inst_type=None):
...
+            if attr_value not in accepted_names:
+                yield ValidationOutcome(inst=inst, expected=accepted_names, observed=attr_value, severity=OutcomeSeverity.ERROR)
+        case "CONVERSIONFACTOR":
+            unit_name = inst.Name
+            if unit_name in accepted_names:
+                inst_factor = inst.ConversionFactor.ValueComponent.wrappedValue
+                inst_si_unit = inst.ConversionFactor.UnitComponent
+                conv_unit_def = ConversionBasedUnitDefinition(**unit_definitions[unit_name])
+                expected_factor = ifcopenshell.util.unit.convert(value=conv_unit_def.ConversionFactor,from_unit=conv_unit_def.SIUnitName,from_prefix=conv_unit_def.SIUnitPrefix,to_unit=inst_si_unit.Name,to_prefix=inst_si_unit.Prefix)
+                if not math.isclose(a=inst_factor, b=expected_factor, rel_tol=1e-06, abs_tol=1e-06):
```
aothms: I quickly checked just now:

> For the values to be considered close, the difference between them
> must be smaller than at least one of the tolerances.

Some values are awfully close to the abs_tol, so this is risky.

```
>>> y = 0.00001639
>>> x = 0.00001639 + (1.e-6-1.e-9)
>>> print(f"{x:.10f}")
0.0000173890
>>> print(f"{y:.10f}")
0.0000163900
>>> math.isclose(x, y, rel_tol=1e-06, abs_tol=1e-06)
True
```

Since these values can never be 0, I would say we only use a rel_tol.

```suggestion
                if not math.isclose(a=inst_factor, b=expected_factor, rel_tol=1e-06, abs_tol=0.):
```

(I didn't test this, but even if this is maybe not the correct solution, I think the current approach doesn't work well for small values and does need to change)

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -375,3 +409,27 @@ def step_impl(context, inst, table, inst_type=None):
...
+            attr_value = getattr(inst, attr_name)
+            if attr_value not in accepted_names:
+                yield ValidationOutcome(inst=inst, expected=accepted_names, observed=attr_value, severity=OutcomeSeverity.ERROR)
+        case "CONVERSIONFACTOR":
+            unit_name = inst.Name
+            if unit_name in accepted_names:
+                inst_factor = inst.ConversionFactor.ValueComponent.wrappedValue
+                inst_si_unit = inst.ConversionFactor.UnitComponent
+                conv_unit_def = ConversionBasedUnitDefinition(**unit_definitions[unit_name])
+                expected_factor = ifcopenshell.util.unit.convert(value=conv_unit_def.ConversionFactor,from_unit=conv_unit_def.SIUnitName,from_prefix=conv_unit_def.SIUnitPrefix,to_unit=inst_si_unit.Name,to_prefix=inst_si_unit.Prefix)
```
aothms: There is still one aspect missing.

unit.convert() does completely disregard `to_unit` when it's a not registered non-si unit. In that case it only looks at `to_prefix`. (see the if-elif branch there in the code).

So we never actually really check if the ConvUnit.ConversionFactor.UnitComponent.Name is specified in the CSV

Edit: see 2nd scenario suggestion.

features/rules/PJS/PJS001_Correct-conversion-based-units.feature
```diff
@@ -0,0 +1,22 @@
...
+  IFC 4: https://standards.buildingsmart.org/IFC/RELEASE/IFC4/FINAL/HTML/schema/ifcmeasureresource/lexical/ifcconversionbasedunit.htm
+  IFC 2X3: https://standards.buildingsmart.org/IFC/RELEASE/IFC2x3/FINAL/HTML/ifcmeasureresource/lexical/ifcconversionbasedunit.htm
+
+  Background: Selection of conversion-based units in default unit assignment
+    Given an .IfcConversionBasedUnit.
+
+  Scenario: Validating correct names for area, length, and volume units
+    Given .UnitType. ^is^ 'AREAUNIT' or 'LENGTHUNIT' or 'VOLUMEUNIT' or 'PLANEANGLEUNIT'
+    Then its attribute .Name. must be defined [according to the table] 'valid_ConversionBasedUnits'
+
```
aothms:
```suggestion
  Scenario: Validating conversion to the appropriate si unit
    Given its attribute .ConversionFactor.
    Given its attribute .UnitComponent.
    Then its attribute .Name. must be defined [according to the table] 'valid_ConversionBasedUnits' column UnitType
```

(I'm making up the `... column UnitType`,  but you get the idea).

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -73,8 +104,11 @@ def get_pset_definitions(schema, table):
         tbl_path = system.get_abs_path(f"resources/{table}.csv")
 
     tbl = system.get_csv(tbl_path, return_type='dict')
-    return {d['property_set_name']: d for d in tbl}
-
+    if table == "valid_ConversionBasedUnits":
```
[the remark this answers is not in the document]
civilx64: I updated the other tables for consistency.

features/steps/steps/propertysets_qtys_units.py
civilx64: Doesn't the `@functools.cache` decorator address this?  The file is only opened the first time; for remaining instances the function parameters are the same so the same dict is returned from the cache without another read from the filesystem.

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -19,6 +22,35 @@
...
+    Name: str
+    UnitType: str
+    ConversionFactor: str
+    SIUnitName: str
+    SIUnitPrefix: str = None
+    Description: str = None
+
+    def __post_init__(self):
+        # capture π in table definitions
+        if "π" in self.ConversionFactor:
```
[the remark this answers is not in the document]
civilx64: I ended up replacing this with the decimal equivalent.

features/resources/IFC2X3/valid_ConversionBasedUnits.csv
civilx64: This field is not actually used for anything.  I revised it to align with the values in the [IfcDimensionsForSIUnit](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcDimensionsForSIUnit.htm) function.

features/steps/steps/propertysets_qtys_units.py
aothms: Yes you're 100% right. I must have been confused with another rule that uses uncached csv reading or I added the caching as part of the performance work.

features/steps/steps/propertysets_qtys_units.py
civilx64: resolved in https://github.com/buildingSMART/ifc-gherkin-rules/pull/471/commits/24ec01823476ae8b30dddb4f5243afa147dde9bd

features/steps/steps/propertysets_qtys_units.py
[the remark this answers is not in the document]
civilx64: resolved in https://github.com/buildingSMART/ifc-gherkin-rules/pull/471/commits/d0f5ef181290c10217c8684282c19fdf81743a9e

PR #482, opened 2025-12-17
PJS001 degrees in IFC2x3 (IVS-701)

civilx64:
Revised per user feedback: eliminate strict schema-specific checking and rather use the table from the 4.3 spec (which includes all conversion-based units from prior versions).

### DIFF
--- a/features/rules/PJS/PJS001_Correct-conversion-based-units.feature
+++ b/features/rules/PJS/PJS001_Correct-conversion-based-units.feature
@@ -1,6 +1,6 @@
 @implementer-agreement
 @PJS
-@version1
+@version2
 Feature: PJS001 - Correct conversion based units
 
   The rule verifies that conversion-based units used per Concept Template 4.1.9.9


features/steps/steps/propertysets_qtys_units.py
```diff
@@ -112,7 +112,7 @@ def normalize(name, case_sensitive=True):
         return name if case_sensitive else name.lower()
 
     schema_specific_path = system.get_abs_path(f"resources/{schema.upper()}/{table}.csv")
-    if os.path.exists(schema_specific_path):
+    if schema_specific_path is not None and os.path.exists(schema_specific_path):
```
aothms: get_abs_path() uses glob, then index 0th element and catches index error then defaults to None.

So path.exists call is probably unnecessary, but the is not None check is not.

Some type hints and better descriptive function names would really help here, but not for now..

aothms (approved):
Looks great. A bit magical that this works without changes to the feature file.

**NB: We should probably really bump the version though!!**

PR #485, opened 2025-12-27
IVS-686 - GRF008

Ghesselink:
Relatively simple rule. <del>However, tests failing due to difficulties with unpacking IfcMeasureValue in the new PackedSequence</del>

Edit: got it working :)

### DIFF
--- a/features/rules/GRF/GRF008_Rigid-operation-units.feature
+++ b/features/rules/GRF/GRF008_Rigid-operation-units.feature
@@ -7,7 +7,7 @@ The rule verifies that correct unit types are assigned to rigid operation coordi
 
   Scenario Outline: Correct rigid operation coordinate units
 
-    Given A model with Schema 'IFC4' or 'IFC4.3'
+    Given A model with Schema 'IFC4.3'
     Given an .IfcRigidOperation.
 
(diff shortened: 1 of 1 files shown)


features/rules/ALS/ALS006_Alignment-horizontal-shape-representation.feature
civilx64: Nota bene: edit to this file was done to agree with previous overall refactorings.  New rule GRF008 uses this step implementation as well.

features/rules/GRF/GRF008_Rigid-operation-units.feature
```diff
@@ -0,0 +1,18 @@
+@implementer-agreement
+@GRF
+@version1
+Feature: GRF008 - Rigid operation units
+The rule verifies that correct unit types are assigned to rigid operation coordinates
+
+
+  Scenario Outline: Correct rigid operation coordinate units
+
+    Given A model with Schema 'IFC4' or 'IFC4.3'
```
civilx64:
```suggestion
    Given A model with Schema 'IFC4.3'
```

`IfcRigidOperation` is a new entity as of IFC 4.3.  Also all of the unit test files are 4.3 as well.

civilx64 (commented):
Nice work! Unit test are passing / failing as expected.  Please correct the schema specification per my suggestion, then we can merge.

PR #494, opened 2026-02-21
IVS-759 consider ifc shape aspect in ifc105

aothms:


Ghesselink (approved):
First, should we bump the version number? 

Normally I'd suggest to add a test file, but it's a bit difficult to isolate the case for this rule. Hence, I ran a manual check and it runs well. As expected, IfcShapeAspect instances are not evaluated. 

```
from validation_results import OutcomeSeverity, ModelInstance, ValidationTask
any(
    context.model.by_id(c.inst).is_a('IfcShapeAspect')
    for c in context.gherkin_outcomes
    if c.severity == OutcomeSeverity.ERROR
)
False # == True in DEV
```

aothms:
> First, should we bump the version number?

Done. Thanks geert!

PR #503, opened 2026-03-03
fix GRF005 for rounding and precision 

civilx64:
logic for map conversion scale factor comparison now matches the logic in conversion-based unit (PSJ001) factor comparisons.

(IVS-753)


Ghesselink (approved):
Fixed a small trailing whitespace issue, for the rest it looks good.

PR #508, opened 2026-03-18
Improve reporting in UI for PJS001 (IVS-786)

civilx64:
- filter the list of expected unit names based on unit type
- add "breadcrumbs" to the entity instance with incorrect conversion factor.

### DIFF
--- a/features/steps/steps/propertysets_qtys_units.py
+++ b/features/steps/steps/propertysets_qtys_units.py
@@ -430,10 +430,10 @@ def step_impl(context, inst, table, inst_type=None):
 def step_impl(context, inst, attr_name):
     unit_definitions = get_table_definition(schema="Not schema specific", table="valid_ConversionBasedUnits", case_sensitive=False)
     unit_type = getattr(inst, 'UnitType', None)
-if unit_type:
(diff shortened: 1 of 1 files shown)


features/steps/steps/propertysets_qtys_units.py
```diff
@@ -429,7 +429,7 @@ def step_impl(context, inst, table, inst_type=None):
 )
 def step_impl(context, inst, attr_name):
     unit_definitions = get_table_definition(schema="Not schema specific", table="valid_ConversionBasedUnits", case_sensitive=False)
-    accepted_names = list(unit_definitions.keys())
+    accepted_names = [k for k, v in unit_definitions.items() if v['UnitType'] == inst.UnitType]
```
Ghesselink: Should we be defensive here? In case inst doesn't have a UnitType attr (schema is invalid in that case, but PJS001 would still run). 
```suggestion
    unit_type = getattr(inst, 'UnitType', None)
if unit_type:
    accepted_names = [k for k, v in unit_definitions.items() if v.get('UnitType') == unit_type]
else:
    accepted_names = list(unit_definitions.keys()) 
```

Ghesselink (changes requested):

PR #513, opened 2026-04-07
Revise IFC105 to Industry Best Practice (IVS-817)

civilx64:
To keep a clean history in the code and database:

* IFC105 is now disabled
* IFC106 is now added as "new" Industry Best Practice (same logic as IFC105 but raises warnings, not errors)


Ghesselink (approved):

Ghesselink:
Fixed a couple of small errors (one file in the IFC106 folder still had the 'ifc105' prefix and there was a small mismatch for the file header)

PR #514, opened 2026-04-12
IVS-842 Update BRP003 description with approach

aothms:


civilx64:
This updated description was deployed as documentation, not as part of the feature file.  See buildingSMART/validate/pull/296.

PR #521, opened 2026-06-08
IVS-879 - Various performance improvements (WIP)

rw-bsi:
- caching (file open, inspect.signature etc...)
- review of `include_subtypes`
- expensive attribute lookups
- expensive `numpy` actions

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -3,6 +3,7 @@ import ifcopenshell.simple_spf
 from behave.model import Scenario
 from collections import Counter
 import functools
+from functools import lru_cache
 import os
 from rule_creation_protocol import protocol
 from features.exception_logger import ExceptionSummary
@@ -11,10 +12,27 @@ import time
 import logging
 import socket
 
+import cProfile
+
 from validation_results import ValidationOutcome, ValidationOutcomeDjango, ValidationOutcomeCode, OutcomeSeverity
 from main import ExecutionMode
 
-@functools.cache
+
+_GLOBAL_IFC_OPEN_CACHE = {}
+_native_open = ifcopenshell.open
+
+def _globally_cached_open(file_info, *args, **kwargs):
+    """Caches the model parsing process across all features and scenarios."""
+    
+    if isinstance(file_info, str): 
+        file_info = os.path.abspath(file_info)
+    if file_info not in _GLOBAL_IFC_OPEN_CACHE:
+        _GLOBAL_IFC_OPEN_CACHE[file_info] = _native_open(file_info, *args, **kwargs)
+    return _GLOBAL_IFC_OPEN_CACHE[file_info]
+
+ifcopenshell.open = _globally_cached_open
+
+
 def read_model(fn, pure):
     # @nb --purepythonparser is only used for @critical rules which is only IFC101 which only looks at the header
     return (ifcopenshell.simple_spf.open(fn, only_header=True)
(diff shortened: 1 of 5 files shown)


features/steps/steps/entity_selection.py
```diff
@@ -31,9 +31,7 @@ def step_impl(context, entity_opt_stmt, subtype_handling=None):
             case _:
                 include_subtypes = True
 
-        # @todo this better be handled more efficiently in ifcopenshell, but idea here is to prevent
-        # allocating large lists
-        instances = (inst for inst in context.model if ((inst.is_a(entity)) if include_subtypes else (inst.is_a().lower() == entity)))
+        instances = context.model.by_type(entity, include_subtypes=include_subtypes) or []
```
aothms, line 34: This was here on purpose. 

> allocating large lists

ifcopenshell.entity_instance is huge in memory. So if there are large lists of these moving around this creates significant spikes.

What we do instead is a generator so that instances do not stay alive long and materialized back into a numeric id as soon as it is processed.

features/steps/utils/geometry.py
```diff
@@ -112,8 +112,10 @@ def inner():
             # @nb to decide: should we return index pairs, or coordinate pairs here?
             coords = inst.Coordinates.CoordList
```
aothms, line 113: Maybe use more numpy?

features/steps/validation_handling.py
```diff
@@ -157,7 +160,17 @@ def handle_given(context, fn, **kwargs):
 def is_nested(val):
     return isinstance(val, (tuple, list, misc.PackedSequence))
 
+
 def apply_operation(fn, inst, context, current_path, kwargs):
+
+    def get_cached_signature(func):
+        
+        cache_key = getattr(func, "__wrapped__", func)
```
aothms, line 168: I'm not sure this would handle nested decorators well. Maybe needs to be a while loop.

I guess `misc.get_stack_tree` is fairly expensive so just passing it to every function is also overkill.

Maybe we can just set the signature onto the function object?

Or maybe access data directly `fn.__code__.co_varnames`?

features/steps/validation_handling.py
```diff
@@ -195,8 +209,15 @@ def iter_given_state(values, fn, context, current_path=[], depth=None, current_d
         # we have arrived at the specified depth, or there is no depth specified and we're at the leaf
         yield apply_operation(fn, values, context, current_path, kwargs)
     else:
-        for i, v in enumerate(values):
-            yield map_given_state(v, fn, context, current_path + [i], depth, current_depth + 1, **kwargs)
+        result_list = [
+            map_given_state(v, fn, context, current_path + [i], depth, current_depth + 1, **kwargs)
+            for i, v in enumerate(values)
+        ]
+        return iter(result_list)
```
aothms, line 216: I don't understand why this would make a significant difference, but there's surely a lot of overhead here in the nesting of nesting of values and general slowness of python in control structures and function calls around it.

Some of these things are weird vibe coded constructs that pack nested lists of instances into contiguous storage. Maybe has something to do with that.

features/environment.py
```diff
@@ -11,10 +12,27 @@
...
+
 from validation_results import ValidationOutcome, ValidationOutcomeDjango, ValidationOutcomeCode, OutcomeSeverity
 from main import ExecutionMode
 
-@functools.cache
+
+_GLOBAL_IFC_OPEN_CACHE = {}
+_native_open = ifcopenshell.open
+
+def _globally_cached_open(file_info, *args, **kwargs):
```
aothms, line 24: Maybe we can use before_all instead of before_feature for opening the model?

PR #523, opened 2026-06-12
New rule SWE003 (IVS-14)

### LINKED ISSUE #159: Possible false positive for IfcSectionedSolidHorizontal
@aothms @civilx64 

The informal propositions for [IfcSectionedSolidHorizontal](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/lexical/IfcSectionedSolidHorizontal.htm) says:
 > 2. If the type of sections is not [IfcParameterizedProfileDef](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/lexical/IfcParameterizedProfileDef.htm) (i.e. an arbitrary profile), **then the number of points and edges should be the same for two consecutive profiles**

The attached file has bridge beams with tapered end blocks where a different number of points are used to define an IfcPolyline for consecutive cross section profiles. This file passes the validation service without error.

Since the informal proposition says "should" instead of "shall", maybe this is ok. Just thought I'd bring it up.

[26122.ifc.txt](https://github.com/buildingSMART/validate/files/14841418/26122.ifc.txt)

civilx64:
Closes #159


features/steps/thens/values.py
```diff
@@ -166,3 +165,78 @@ def get_value(varname):
...
+
+                case _:
+                    pass
+
+    for pair_of_profiles in inst:
+        curr_profile, next_profile = pair_of_profiles
+        if (curr_profile is not None) & (next_profile is not None):
+            curr_profile, next_profile = misc.iflatten(curr_profile), misc.iflatten(next_profile)
+
+            for cp, np in zip(curr_profile, next_profile):
```
aothms, line 264: Be careful. `zip` truncates to shortest length. So if I'm reading correctly, in case of composite profile def, a different number of profiles would be silently accepted.

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
```diff
@@ -0,0 +1,37 @@
...
+
+    Given a model with Schema 'IFC4.3'
+
+    Given an .IfcSectionedSolidHorizontal. ^without subtypes^
+    Given its attribute .CrossSections.
+
+  Scenario: IfcArbitraryClosedProfileDef
+
+    Given [its entity type] ^is^ 'IfcArbitraryClosedProfileDef'
+    Given the values grouped pairwise at depth 1
```
aothms, line 20: I'm thinking out loud here, another way to phrase this would be

```
...
Given the edges of geometry
Given their length # length is array length __len__() here not vector length / norm2
Then the values must be identical at depth ...
```

The equal at depth thing would mean in case of arbitrary nesting you would compare tuples with integer length like `(4,(4,4,)) == (4,(4,5)) == (4,(4,5))` so that generalizes quite nicely. And maybe it would be a way to use more general constructs rather than the `the profiles must have the same number of points and edges` with a large match statement. Because we don't care about pairwise here, we care about all edge counts in the cross sections and with "identical at depth" we already have a building block for that.

There is a `steps.utils.geometry.get_edges` helper function already, maybe that can be uplifted to an actual step implementation function as well.

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
```diff
@@ -0,0 +1,37 @@
+@informal-proposition
+@SWE
+@version1
+Feature: SWE003 - Number of points and edges for IfcSectionedSolidHorizontal profiles
+
+The rule verifies that if the type of sections is not IfcParameterizedProfileDef,
+then the number of points and edges must be the same for two consecutive profiles.
+
+  Background: Sections for IfcSectionsSolidHorizontal
```
aothms:
```suggestion
  Background: Sections for IfcSectionedSolidHorizontal
```

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
```diff
@@ -0,0 +1,37 @@
...
+
+    Given a model with Schema 'IFC4.3'
+
+    Given an .IfcSectionedSolidHorizontal. ^without subtypes^
```
aothms: Any reason for `^without subtypes^`?

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
```diff
@@ -0,0 +1,37 @@
+@informal-proposition
+@SWE
+@version1
+Feature: SWE003 - Number of points and edges for IfcSectionedSolidHorizontal profiles
+
+The rule verifies that if the type of sections is not IfcParameterizedProfileDef,
+then the number of points and edges must be the same for two consecutive profiles.
```
aothms, line 7: Should there be a rule about profiles being of the same type (handling derived,composite appropriately)

Because now mixing IfcRectangleProfileDef with IfcArbitraryClosed will be accepted by this rule?

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
```diff
@@ -0,0 +1,37 @@
+@informal-proposition
+@SWE
+@version1
+Feature: SWE003 - Number of points and edges for IfcSectionedSolidHorizontal profiles
```
aothms, line 4:
```suggestion
Feature: SWE003 - Number of points and edges for IfcSectionedSolidHorizontal profiles with arbitrary profile definitions.
```

features/steps/thens/values.py
```diff
@@ -166,3 +165,78 @@ def get_value(varname):
...
+                        curr_crv_seg_count = len(curr_crv.Segments)
+                        next_crv_seg_count = len(next_crv.Segments)
+                        if curr_crv_seg_count != next_crv_seg_count:
+                            expected_msg = f"{curr_crv_seg_count} segments in profile definition curve"
+                            observed_msg = f"{next_crv_seg_count} segments in profile definition curve"
+                            yield ValidationOutcome(inst=next_crv, expected=expected_msg, observed=observed_msg,
+                                                    severity=OutcomeSeverity.ERROR)
+
+                    # iterate segments and confirm same number of points in each
+                    for seg_in_curr, seg_in_next in zip(curr_crv.Segments, next_crv.Segments):
```
aothms, line 218: Being pedantic here

But Segments could be `[LineIndex((0,1,2,3,0)]` or `[LineIndex((0,1), LineIndex(1,2), LineIndex((2,3), LineIndex(3,0)]` and these semantically are the same thing but we flag them as different.

features/steps/thens/values.py
```diff
@@ -166,3 +165,78 @@ def get_value(varname):
...
+                        # segment will be either IfcLineIndex or IfcArcIndex
+                        # the two segment types don't necessarily have to be the same - e.g. could sweep between polyline of three points and an arc
```
aothms, line 220: Here I disagree though. I don't see why we should support sweeping between a polyline and an arc.

More importantly the indices have a different meaning in case of

- IfcArcIndex: `LIST[3:3]` - exactly one edge by means of a 3-point arc
- IfcLineIndex: `LIST[2:?]` - one or more linear edges (N-1)

So we can't say seg.wappedValue equals pt_count or compare index counts across segment types.

aothms: Re:

> I don't see why we should support sweeping between a polyline and an arc.

I'm fine leaving this out (and I think we should) for this rule. But the number of edges from IfcArcIndex, IfcLineIndex needs to be properly counted. I would just sum the number (len(wV) - 1 for LineIndex; 1 for ArcIndex) of edges for the profile.

features/steps/thens/values.py
```diff
@@ -166,3 +165,78 @@ def get_value(varname):
...
+        if (curr_profile is not None) & (next_profile is not None):
+            curr_profile, next_profile = misc.iflatten(curr_profile), misc.iflatten(next_profile)
+
+            for cp, np in zip(curr_profile, next_profile):
+                curr_profile_type = cp.is_a().upper()
+                next_profile_type = np.is_a().upper()
+
+                if curr_profile_type == next_profile_type:
+                    match curr_profile_type:
+                        case "IFCARBITRARYCLOSEDPROFILEDEF":
```
aothms: If you have IfcDerived/Composite profile def. I see no reason why not to handle `IfcArbitraryProfileDefWithVoids`

features/steps/thens/values.py
```diff
@@ -166,3 +165,78 @@ def get_value(varname):
...
+                            curr_curve, next_curve = cp.OuterCurve, np.OuterCurve
+                            yield from handle_curves(curr_curve, next_curve)
+
+                        # This stub waa left in place for future similar rules that might be developed for surfaces -
+                        # IfcArbitraryClosedProfileDef is used for generating solids and
+                        # IfcArbitraryOpenProfile used for generating surfaces.
+                        case _:
+                            pass
```
aothms: I don't like these silent `pass` statements.

features/steps/thens/values.py
```diff
@@ -166,3 +165,78 @@ def get_value(varname):
...
     passed = binary_operators[op](v1, v2)
     yield ValidationOutcome(inst=inst, expected=v2, observed=v1, severity=OutcomeSeverity.PASSED if passed else OutcomeSeverity.ERROR)
+
+@full_stack_rule
+@gherkin_ifc.step('the profiles must have the same number of points and edges')
+def step_impl(context, inst):
+    def handle_curves(curr_crv, next_crv):
+        curr_crv_type = curr_crv.is_a().upper()
+        next_crv_type = next_crv.is_a().upper()
+        if curr_crv_type == next_crv_type:
```
aothms, line 184: I don't like this if without else. Even if we do not handle it as part of this rule I'd like to see something written down.

features/steps/thens/values.py
```diff
@@ -166,3 +165,78 @@ def get_value(varname):
...
+            match curr_crv_type:
```
aothms, line 185: I noticed you like your `match` statements. No need to change. It's also ifcopenshell's fault for being so untyped in python, but there's some downsides:

- Can't support inheritance
- Easy to make typos
- Need to be case specific
- etc.

features/steps/thens/values.py
civilx64: What should it be instead?  Yield an outcome with OutcomeSeverity.PASSED?

aothms: If I read it correctly, based on the pairing with the current scenario I would more expect something like

```py
# To be potentially expanded in case of other scenarios. Currently only used in the context of SWE003 which explicitly selects only IfcArbitraryClosedProfileDef faces.
assert curr_profile_type.is_a('IfcArbitraryClosedProfileDef') and next_profile_type.is_a('IfcArbitraryClosedProfileDef')
```

this match statement just kind of put me on the wrong foot.

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
civilx64: Agree, no need for this.  In 4.3 there are no subtypes anyway.  Updated in 5560b6ae6183cd7061cb75abb7f61e27483ebbb8

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
civilx64: This rule is designed to enforce informal proposition 2.:

> If the type of sections is not [IfcParameterizedProfileDef](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcParameterizedProfileDef.htm) (i.e. an arbitrary profile), then the number of points and edges should be the same for two consecutive profiles

Yes, mixing IfcRectangleProfileDef with IfcArbitraryClosed results in a passing result for SWE003.  The former is IfcParameterizedProfileDef and so in my mind this informal prop does not apply.  I'm not aware of anything else in the spec that requires Profiles to be of the same type.

That's my interpretation from a strict, formal sense.  From a practical sense I agree that Profiles should be of the same type, but that would have to be a best practice rule and should have some discussion among a wider audience before it lands in the validation service.

features/steps/thens/values.py
civilx64: Agree, we need a proper count of edges also.  Added in 212f3c4557f0e11291619117ca87d2a3029373e4

features/steps/thens/values.py
civilx64: Added in 08fed0535dcdf3bd42414ccfa1ab61a488665775

features/steps/thens/values.py
civilx64: clarified in 3553f57259765d285abb06913d2372182d19559a

features/steps/thens/values.py
civilx64: `IfcArbitraryProfileDefWithVoids` is now accepted due to changes from `match` to `is_a()` in 3553f57259765d285abb06913d2372182d19559a - since the later includes subtypes.

features/steps/thens/values.py
civilx64: Correct, that case is not handled correctly.  Which leads to a rather fundamental re-architecture of the feature file.

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
aothms: I get what you're saying. I think these perspectives are both valid

1. Require evidence from the spec in order to be able to constrain things
2. Be cautious of flagging things are valid when practically it's hard to justify

I think a nice middle ground would be to then be explicit in the spec and description (open suggestions)

> Feature: SWE003 - Number of points and edges for IfcSectionedSolidHorizontal profiles with arbitrary profile definitions. 

> The rule verifies that if the type of sections is not IfcParameterizedProfileDef,
> then the number of points and edges must be the same for two consecutive profiles.
> The rule only verifies profiles of type IfcArbitraryClosedProfileDef.

aothms, line 7:
```suggestion
The rule verifies that if the type of sections is not IfcParameterizedProfileDef,
then the number of points and edges must be the same for two consecutive profiles.
The rule only verifies profiles of type IfcArbitraryClosedProfileDef.
```

features/steps/thens/values.py
```diff
@@ -166,3 +166,108 @@ def get_value(varname):
...
+                    if curr_crv_pt_count != next_crv_pt_count:
+                        expected_msg = f"{curr_crv_pt_count} points in profile definition curve"
+                        observed_msg = f"{next_crv_pt_count} points in profile definition curve"
+                        yield ValidationOutcome(inst=next_crv, expected=expected_msg, observed=observed_msg,
+                                                severity=OutcomeSeverity.ERROR)
+
+                case "IFCINDEXEDPOLYCURVE":
+                    # check overall number of points
+                    curr_crv_pt_count = len(geometry.get_points(curr_crv))
+                    next_crv_pt_count = len(geometry.get_points(curr_crv))
```
Ghesselink, line 200: This is a typo?
```suggestion
                    next_crv_pt_count = len(geometry.get_points(next_crv))
```

features/steps/thens/values.py
Ghesselink: Can't this also cause errors? `Segments i`s an optional attribute for` IfcIndexedPolyCurve`. When I simulate this, I get an error 

```
>>> import ifcopenshell
>>> f = ifcopenshell.file(schema="IFC4X3_ADD2")
>>> pts = f.create_entity("IfcCartesianPointList2D", CoordList=[(0.,0.),(1.,0.),(1.,1.),(0.,0.)])
>>> crv = f.create_entity("IfcIndexedPolyCurve", Points=pts)
>>> list(zip(crv.Segments, crv.Segments)) 
    list(zip(crv.Segments, crv.Segments))
         ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

Image

It doesn't show up in CI because no fixture leaves Segments unset. Every polycurve test file populates it so the None path is never used ..

Fix is easy (tested it locally, I can apply it if you like); just fold the loop inside the existing if curr_crv.Segments and next_crv.Segments guard.

Ghesselink (changes requested):

features/rules/SWE/SWE003_Number-of-points-and-edges-for-IfcSectionedSolidHorizontal-profiles.feature
civilx64: Thanks, this helps with the issues I was having with composite profiles.  I started working in this direction but ran into a dead end in that the output of the second `Given` step is not pushed to the context stack.  `misc.encode_nested_tuples()` does not support tuples of `int`, meaning that the tuples with integer length are not available for the final `the values must be identical at depth n` step.

I see two possible paths forward:
1. adjust `encode_nested_tuples()` to handle tuples of `int`
2. store the tuples of integer length in something like `context.edges_or_points_count`

Which would you prefer? (my hunch is 1.) Ping @Ghesselink for input as well.

civilx64: Nevermind, I found an easier solution via returning a string e.g. "4 edges" instead of just the integer 4 which gets encoded as `context.model.by_id(4)` on the context stack.

Ghesselink: Edit: I see the PR is already merged, but thought I'd leave it here.

@civilx64 your hunch (option 1) was right. It's actually already implemented in the [PR ](https://github.com/civilx64/ifc-gherkin-rules/pull/2)I opened on your branch: encode_nested_tuples then tags entity ids and plain ints differently, so tuples of ints survive the round trip and the values must be identical at depth n works on real counts. 

So I think it's possible drop the "4 edges" string workaround. It works because strings happen to trigger the fallback path, and it leaves the underlying `by_id(4)` bug (retrieving an ifcopenshell entity instance instead of an int) in place for any future step that yields ints. 
With the PR merged the original Given the number of [edges] + Then identical at depth 1 flow works as intended.

PR #526, opened 2026-06-24
Best practice: detect entities outside the scope of IFC 4.3 MVDs (IVS-746)

civilx64:
Implementation here varies slightly from the plan in JIRA so that we could re-use the step implementation of IFC102.  The list of entities **NOT** in scope was generated in the ifc-4.x code base by introspecting the schema definition with `ifcopenshell.schema_by_name` and subtracting entities in `mvd_entity_usage.json`.


features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
```diff
@@ -0,0 +1,326 @@
+@IFC
+@version1
+@industry-practice
+Feature: IFC430 - Entities in scope for IFC 4X3 ReferenceView MVD
```
aothms, line 4: Love the rule code ;)

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
```diff
@@ -0,0 +1,326 @@
...
+
+The rule verifies that IFC models using schema version 4.3 and declaring the Reference View Model View Definition (MVD)
+do not export any entities that are outside the scope of this MVD.
+
+  Scenario Outline: Check for in-scope entities - IFC4.3 ReferenceView
+
+    Given A model with Schema 'IFC4.3'
+    Given A model with Model View Definition 'ReferenceView'
+    Given An IFC model
```
aothms, line 13: Is this needed?

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
```diff
@@ -0,0 +1,326 @@
...
+
+    Then There must be less than 1 instance(s) of .<Entity>. ^excluding subtypes^
```
aothms, line 15: Sorry, I find this a bit ugly


```suggestion
    Then There must be 0 instance(s) of .<Entity>. ^excluding subtypes^
```

or


```suggestion
    Then There must be no instance(s) of .<Entity>. ^excluding subtypes^
```

aothms: Shouldn't be that hard to add another overload to existance.py

While at it, it seems that these can be unified:

    "There must be {constraint} {num:d} instance(s) of .{entity}. ^{subtype_handling:include_or_exclude_subtypes}^"
    "There must be {constraint} {num:d} instance(s) of .{entity}."

And then I would just 

default `constraint` to 'equals' and `num` to 0.

and add an overload which picks up these new defaults

    "There must be no instance(s) of .{entity}."

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
```diff
@@ -0,0 +1,326 @@
+@IFC
+@version1
+@industry-practice
+Feature: IFC430 - Entities in scope for IFC 4X3 ReferenceView MVD
+
+The rule verifies that IFC models using schema version 4.3 and declaring the Reference View Model View Definition (MVD)
+do not export any entities that are outside the scope of this MVD.
+
+  Scenario Outline: Check for in-scope entities - IFC4.3 ReferenceView
```
aothms, line 9: A performance caveat here. Implementing these as a Scenario Outline means a top-down evaluation for all examples. I don't know if this has a measurable impact.

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
```diff
@@ -0,0 +1,326 @@
...
+
+    Given A model with Schema 'IFC4.3'
+    Given A model with Model View Definition 'ReferenceView'
+    Given An IFC model
+
+    Then There must be less than 1 instance(s) of .<Entity>. ^excluding subtypes^
+
+    Examples:
+      | Entity                                       |
+      | IfcAdvancedBrep                              |
```
aothms, line 19: Please document somewhere in this repo how this list was generated

features/steps/utils/ifc.py
```diff
@@ -47,10 +48,13 @@ def get_relation(instance, attrs : list):
     return next((rel for rel in relations if rel is not None), None) # always len == 1
 
 def get_mvd(ifc_file):
-    try:
-        detected_mvd = ifc_file.header.file_description.description[0].split(" ", 1)[1]
-        detected_mvd = detected_mvd[1:-1]
-    except:
+    pattern = r"\[(.*?)\]"
+    header = ifc_file.header
+    desc = header.file_description.description[0]
```
Ghesselink: description [0] will raise an error if the file description has an empty list (e.g. `FILE_DESCRIPTION((),'2;1')`). The old try/except block was a guardrail for this and returned 'None' here. The regex version is actually a nice improvement, but we should probably put the guardrail back so prevent messy headers to crash the rule. 

```suggestion
    descriptions = header.file_description.description
    if not descriptions:
        return None
    desc = descriptions[0]
```

```
>>> f = ifcopenshell.file(schema="IFC4X3_ADD2")
>>> f.header.file_description.description = () 
>>> f.header.file_description.description[0] 
    f.header.file_description.description[0]
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: tuple index out of range
>>> 
```

Ghesselink (changes requested):

features/steps/utils/ifc.py
aothms: Good catch, or

    next(iter(header.file_description.description), None)

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
civilx64: I created a small script in the ifc-4-x repository that generates these lists of excluded entities and just opened [PR#1144](https://github.com/buildingSMART/IFC4.x-development/pull/1144).

Does it make sense to keep it there in that repo or save elsewhere?

I also put some explanatory text in the PR description, but not sure how easy it is to follow.

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
civilx64: Doesn't it have to be `Scenario Outline` in order to use `Examples`?

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
civilx64: After re-running a local test to confirm, it appears not.  But these feature files were copied directly from IFC102 - which was the original intent, to reuse existing steps and implementations as-is.

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
civilx64: This is literally "six to one, half a dozen to another" :grin: - and what we are already using in IFC102.

Ghesselink (approved):

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
Ghesselink: Yep .. `Example `only works inside a Scenario Outline (behave doesn't run it otherwise). 
I got curious and ran it against a real model from the DB: 74 MB and around 1M entities. Opening the file took~5.5s. Checking all 308 entity types then takes ~0.2s total, and that's with 22 of them actually present (220k instances),. The Scenario Outline overhead on top is a tenth of a second and doesn't grow with model size. 

(fyi @aothms )

aothms: > Example only works inside a Scenario Outline 

Indeed, the alternative I had in my brain was a normal scenario with the much uglier:

```
    Scenario Check for in-scope entities - IFC4.3 ReferenceView
    Given A model with Schema 'IFC4.3'
    Given A model with Model View Definition 'ReferenceView'
    Given An IFC model
    Then There must be less than 1 instance(s) of .IfcAdvancedBrepWithVoids. and .IfcAdvancedFace. .IfcAlignmentCant. and .IfcAlignmentCantSegment. and ...
```

or maybe we would have put them in a .CSV file like we do in a lot of other cases.

Thanks for measuring @Ghesselink 💪 

> The Scenario Outline overhead on top is a tenth of a second and doesn't grow with model size.

That's indeed a very relevant observation, the `Given`s that are executed for every Example all concern the header and don't really push instances on the stack.

I'll resolve this.

features/rules/IFC/IFC430_Entities-in-scope-for-IFC-4X3-ReferenceView-MVD.feature
aothms: I think you have too much confidence in the literacy of people

PR #534, opened 2026-07-30
add TAS000 activation rule for tessellated geometry (IVS-342)

civilx64:
Note: this is targeting the development branch because by design, activation rules can never generate a failing (ERROR) ValidationOutcome.  In other words, scorecard for TAS000 can only be either gray or green.


Ghesselink (approved):
Nice!

PR #535, opened 2026-07-31
add out-of-scope entity IfcMapConversionScaled (IVS-889)

civilx64:


aothms:
Also see discussion in https://github.com/buildingSMART/IFC4.x-development/pull/1144

PR #540, opened 2026-08-20
New rule PJS004 (IVS-857)

civilx64:
to confirm property measure values have units defined directly or in the project context


features/steps/steps/propertysets_qtys_units.py
```diff
@@ -459,3 +470,77 @@ def step_impl(context, inst, attr_name):
...
+                                        observed=None,
+                                        severity=OutcomeSeverity.ERROR)
+        except KeyError as e:
+            raise e
+
+    properties = inst.HasProperties
+    for prop in properties:
+        if prop.is_a("IfcPropertyBoundedValue"):
+            attrs_to_check = ["UpperBoundValue", "LowerBoundValue", "SetPointValue"]
+            for attr in attrs_to_check:
```
aothms, line 496:
```suggestion
            if any(getattr(prop, attr) for attr in attrs_to_check):
```

You're yielding multiple times. I would find any() clearer.

aothms: You can also just continue to jump to the next prop. Or is it intentional to yield multiple times? The context presented to the user is the same: prop

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -459,3 +470,77 @@ def step_impl(context, inst, attr_name):
                     obs_message = f"Incorrect conversion factor {inst_factor} in #{inst.ConversionFactor.id()}={inst.ConversionFactor.is_a()}"
                     yield ValidationOutcome(inst=inst, expected=exp_message, observed=obs_message,
                                             severity=OutcomeSeverity.ERROR)
+
+@gherkin_ifc.step(
+    "Each associated .{inst_type:property_or_physical_quantity}. value must have units defined directly or in the project context [according to the table] '{table}'")
+def step_impl(context, inst, table, inst_type=None):
+    measure_unit_table = get_table_definition(context.model.schema, table)
+    def confirm_unit_in_project_context():
```
aothms, line 478:
```suggestion
    def confirm_unit_in_project_context(prop, measure_attr):
```

Please pass the measure to the function otherwise it's really unclear what the function depends on.

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -459,3 +470,77 @@ def step_impl(context, inst, attr_name):
...
+                                        expected=f"Units for {attr_measure_type} defined in property or in global project context",
+                                        observed=None,
+                                        severity=OutcomeSeverity.ERROR)
+        except KeyError as e:
+            raise e
+
+    properties = inst.HasProperties
+    for prop in properties:
+        if prop.is_a("IfcPropertyBoundedValue"):
+            attrs_to_check = ["UpperBoundValue", "LowerBoundValue", "SetPointValue"]
```
aothms, line 495: Could have been a mapping:

```
{"IfcPropertyBoundedValue": ["UpperBoundValue", "LowerBoundValue", "SetPointValue"], ...}
```

But as some props require different handling I find it ok either way

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -459,3 +470,77 @@ def step_impl(context, inst, attr_name):
...
+            for attr in attrs_to_check:
+                measure_attr = getattr(prop, attr)
+                if measure_attr:
+                    unit_attr = getattr(prop, "Unit")
+                    if not unit_attr:
+                        yield from confirm_unit_in_project_context()
+
+        elif prop.is_a("IfcPropertyEnumeratedValue"):
+            measure_attrs = getattr(prop, "EnumerationValues")
```
aothms, line 504: You do not check `prop.EnumerationReference.EnumerationValues`

features/rules/PJS/PJS004_Units-for-numeric-measures.feature
```diff
@@ -0,0 +1,12 @@
+@implementer-agreement
+@PJS
+@version1
+Feature: PJS004 - Units for numeric measures
```
aothms, line 4:
```suggestion
Feature: PJS004 - Units for numeric measures in properties
```

I'm not against this exact implementation, but the scope suggested is too broad, what remains:

- IfcElementQuantity: a lot simpler
- all other measures: a bit more involved, for example IfcTrimmedCurve with IfcParameterValue has always been a source of bugs. We do have existing logic for enumerating the defined type for each value in the model: 

https://github.com/IfcOpenShell/IfcOpenShell/blob/998060b686e22297a04eabb8154f50fdf5868641/src/ifcopenshell-python/ifcopenshell/express/rule_executor.py#L249

^ loop over attribute values and schema type in lock step for every entity instance in the model

features/steps/steps/propertysets_qtys_units.py
```diff
@@ -459,3 +470,77 @@ def step_impl(context, inst, attr_name):
...
+            raise e
+
+    properties = inst.HasProperties
+    for prop in properties:
+        if prop.is_a("IfcPropertyBoundedValue"):
+            attrs_to_check = ["UpperBoundValue", "LowerBoundValue", "SetPointValue"]
+            for attr in attrs_to_check:
+                measure_attr = getattr(prop, attr)
+                if measure_attr:
+                    unit_attr = getattr(prop, "Unit")
```
aothms, line 499:
```suggestion
                    unit_attr = prop.Unit
```

Why all the getattrs with literal strings?

aothms (changes requested):

PR #543, opened 2026-08-22
new rule RFT001 - referent nesting (IVS-929)

civilx64:


aothms (approved):
As a non-domain practitioner I would have expected it to make sense to nest referents also under the horizontal/vertical/cant parts directly (e.g where the point of max curvature sits on the horizontal only). But I cross referenced with the guide from Rick and I didn't see any wording that backs my intuition so I assume your conclusion is safe [nor in the 4.3 spec templates https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Object_Composition/Nesting/Object_Nesting/content.html]

---

In terms of implementation an alternative would have been to use a structure analogous to our spatial composition rule.

https://github.com/buildingSMART/ifc-gherkin-rules/blob/893f8275049b8960c3ce0c7379b4b57aefeaa8f9/features/rules/SPS/SPS002_Correct-spatial-breakdown.feature#L13

All other elements in the nesting table are about scheduling though so maybe its wise not to make an authoritative judgement about these at this point.

But the inconsistency in implementation and therefore also rule structuring (is this a concern of the relationship (as we did in the case of spatial composition); or is this a concern of the element) is worth flagging.

civilx64:
Thanks for the feedback. Referent should be nested only to IfcAlignment. The docs incorporate ISO 19148 by reference. Per its description it covers:

> locations relative to a one-dimensional object described by measurement along (and optionally offset from) that object

Accordingly, it is possible to exchange IfcAlignment as this one-dimensional object with no layouts and no shape representation. 

SPS002 is in the "spatial structure" functional part, so it's logical that the relationship is the focus. RFT001 is in the referent functional part, so the object, not the relationship, is the focus.

RickBrice:
I updated the guide document a little bit on this point. Not all referents will fall between segment end/start point. Examples are IfcReferent.INTERSECTION or IfcReferent.POSITION that represents the high/low point of a vertical curve segment. These referents may fall within the domain of individual segments, so what is the sorting order? If they are in an exclusive "referents-only" nest, then sorting isn't an issue.

aothms:
> so what is the sorting order? If they are in an exclusive "referents-only" nest, then sorting isn't an issue.

I know my opinion deviates from most, but I don't understand the desire to give weight to order. In the spec we don't talk about the significance of order, only in case of the order of segments and continuity. Referents have their station by which we can order them after deserialisation. Hor,Ver,Cant also use ordered nesting within an Alignment, also here we don't attach significance to the order. In the building domain we have a similar situation with IfcBuildingStorey aggregated (so un-ordered) relationship within an IfcBuilding. These Storeys have an Elevation attribute (or more detailed pset values) and most tools sort based on this in their UI. If we don't explicitly write about it in the spec, I would be much more relaxed if we didn't infer meaning from it. Same for multiple relationships vs one. These things may also make it harder to round trip in adjacent serialisations: semantic web, databases, or other near-compatible or more conceptual representation forms where ordered aggregates or objectified relationships are less natural to represent.

aothms:
> SPS002 is in the "spatial structure" functional part, so it's logical that the relationship is the focus. RFT001 is in the referent functional part, so the object, not the relationship, is the focus.

That was exactly my question. If we follow the concept templates of the IFC4.3 spec, why do we in some case make the relationship the focus and in other cases the object - what is the guidance here for future rule authoring? But it's "mustard after the meal" as we say in Dutch.

RickBrice:
> I don't understand the desire to give weight to order

I have only given weight to order because IfcRelNests.RelatedObjects are "parts in the ordered while/part relationship". IfcRelNests was specifically selected for alignment layouts, segments, and referents so it seems that order is a primary concern. Otherwise, wouldn't IfcRelAggregates be sufficient?

Thinking about what I wrote yesterday, "These referents may fall within the domain of individual segments, so what is the sorting order?" - that's a bit of a silly statement. I was thinking about the shape of the RelatedObject in the nest (a segment versus a point), not the position of it. The individual alignment segments and the referents are all positioned by a single point, making ordering by distance along simple. But this would require that ObjectPlacement is an IfcLinearPlacement. I'm beginning to rethink my assertion that it is better to keep alignment layouts and referents in separate nests.

aothms:
> IfcRelNests was specifically selected for alignment layouts, segments, and referents **so it seems that order is a primary concern**. Otherwise, wouldn't IfcRelAggregates be sufficient? [emphasis mine]

Not necessarily, I think things are not always so purposeful, also the semantics between Nesting vs Aggregation have changed quite a bit over the course of 2X3 to 4X3.

But I think most importantly, yes, the segments are nested with purposely ordered sequence, and I think mixing Aggregation and Nesting was maybe another novelty they didn't want embark on, so everything related to alignments became a nesting relationship. That's how I read things at least.

